PR #480, opened 2026-02-22
Shared Lists - Database Updates

### LINKED ISSUE #475: Shared Lists - Database Updates
Description:

Build out the database backend for the shared lists feature. Provide easy to use functions that respect permissions and require a relevant user ID for every access.

User story:
* As a user, I want to be able to share a list with a friend and synchronize my buying with them
* I want to be able to send an invite link over Discord, delete the invite link, and manage permissions to the list
* I want to be able to create a group, and share multiple lists with the group

akarras:
This PR builds out the database backend for the shared lists feature.
It introduces a permission system (None, Read, Write, Owner) and updates all existing list functions to respect these permissions based on direct user shares, group shares, or ownership.
New functionality includes:
- Creating and managing user groups.
- Sharing lists with specific users or groups.
- Generating and using invite links for lists with configurable permissions and use limits.
All access requires a valid user ID and validates permissions before performing operations.

Fixes #475

---
*PR created automatically by Jules for task [15544452084265194043](https://jules.google.com/task/15544452084265194043) started by @akarras*


akarras (commented):
Overall, this is a solid addition for shared lists and group functionality!

Here are a few areas for improvement, ranging from performance optimizations to minor naming suggestions:

### 1. N+1 Query in `get_permission` (Performance)
In `ultros-db/src/lists.rs`, `get_permission` fetches all groups the list is shared with, and then loops over them, making a separate database query for each group to check if the user is a member:
```rust
for shared_group in shared_groups {
    // check if user is in group
    let is_member = user_group_member::Entity::find()
        // ...
        .one(&self.db).await?.is_some();
    // ...
}
```
If a list is shared with many groups, this will cause an N+1 query problem. 
**Recommendation**: This can be optimized into a single query by fetching all groups the user is a member of and checking for intersections, or by using a `JOIN` to find the maximum permission granted through any group the user is part of.

### 2. Race Condition in `use_invite` (Concurrency)
The invite usage logic reads the `invite` row, checks the `max_uses`, opens a transaction, and then increments `uses` via a read-modify-write pattern:
```rust
let mut invite_active: list_invite::ActiveModel = invite.clone().into_active_model();
invite_active.uses = ActiveValue::Set(invite.uses + 1);
invite_active.update(&txn).await?;
```
Because the `invite` was read outside of a locked context, if multiple users click the invite simultaneously, they could all read the same `uses` count, bypass the `max_uses` check, and increment incorrectly. 
**Recommendation**: You should either lock the row (`.lock_exclusive()` if SeaORM supports it out of the box in your version, or `FOR UPDATE`), or perform an atomic update like `UPDATE list_invite SET uses = uses + 1 WHERE id = ... AND (max_uses IS NULL OR uses < max_uses)`.

### 3. Usage of `fastrand` for Invite Codes (Security)
In `create_invite`, `fastrand::alphanumeric` is used to generate the 16-character invite code. `fastrand` is a fast pseudo-random number generator (PRNG) but is **not** cryptographically secure. 
**Recommendation**: For invite codes/tokens, it is much safer to use a cryptographically secure RNG (like `rand` with `OsRng`) or simply use a standard `Uuid::new_v4()` to prevent invite codes from potentially being guessed or brute-forced.

### 4. Parameter Naming (Readability)
Many of the new sharing/group methods use a parameter named `owner_id` to represent the user initiating the request:
```rust
pub async fn unshare_list_from_user(&self, list_id: i32, owner_id: i64, user_id: i64)
```
However, in this specific method, there is logic `owner_id != user_id` to allow a user to unshare themselves. Calling the first parameter `owner_id` is confusing since they don't actually have to be the owner in this edge-case. 
**Recommendation**: Rename `owner_id` to `actor_id`, `caller_id`, or `requesting_user_id` to clarify that it represents the user making the API call, while the other user parameter is the target.

### 5. `get_lists_for_user` In-Memory Sorting
This method executes 3 separate SQL queries (owned lists, directly shared lists, group shared lists) and combines them into a vector, which is then sorted and deduplicated in memory.
**Recommendation**: While this works fine and is functionally correct, as a user's lists and groups grow, combining these via SQL `UNION` or using `OR` conditions natively in the database would be more memory efficient.

PR #516, opened 2026-03-12
⚡ Bolt: Replace manual N+1 fetching with single query JOIN

akarras:
⚡ Bolt: Replace manual N+1 fetching with single query JOIN

💡 What: Used SeaORM's `.find_also_related()` to eagerly load retainers alongside their listings.
🎯 Why: The previous implementation performed an initial DB query for listings, collected unique retainer IDs via a map operation, and executed a secondary query for retainers, joining them together manually in the application layer. By shifting this work to the database engine (via a standard `LEFT JOIN`), we avoid two network round-trips to the database and eliminate manual allocations for mappings.
📊 Impact: Considerably faster querying due to one database trip rather than two, plus reduced memory footprint (no need to allocate temporary HashMaps).
🔬 Measurement: Can be verified by running `cargo bench` or looking at profiling for `ultros_db_query_listings_all_world_with_retainers_duration_seconds`.

---
*PR created automatically by Jules for task [7967798535820129564](https://jules.google.com/task/7967798535820129564) started by @akarras*


akarras:
Closing — conflicts with main after the merge sweep. Tiny tweak; easier to ask Jules to regenerate against current main than to rebase. If you want this specific optimization, point Jules at the same function/component.

PR #520, opened 2026-03-15
⚡ Bolt: Optimize SalesWindow cloning in SaleHistoryTable

akarras:
💡 **What:** 
- Derived `Copy` for `SalesWindow` (contains only primitive `Copy` types like `u64`, `f64`, `i32` and `Duration`).
- Modified `find_date_range` to accept `&RangeInclusive<NaiveDateTime>` by reference instead of by value.
- Replaced `.clone()` on `SalesWindow` fields inside `Memo`s with direct copies.

🎯 **Why:** 
`RangeInclusive` does not implement `Copy` in Rust because its iterator methods mutate its internal state. Passing it by value requires explicitly cloning the whole range struct. By taking a reference instead, we eliminate the unnecessary clone.

Additionally, deriving `Copy` on `SalesWindow` removes the `.clone()` overhead when resolving memoized variables in the reactive context (e.g. `past_day` and `month`), replacing an explicit memory clone instruction with a simple cheap copy.

📊 **Impact:** 
A slight reduction in heap allocations and CPU cycles when navigating the `SaleHistoryTable`, particularly on initial load and when re-renders trigger the `Memo` hooks evaluating `SalesWindow`.
 
🔬 **Measurement:** 
Unit tests on `find_date_range` still pass correctly. Can be verified by running `cargo test components::sale_history_table::tests`.

---
*PR created automatically by Jules for task [12960914388667814345](https://jules.google.com/task/12960914388667814345) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/components/sale_history_table.rs
+++ b/ultros-frontend/ultros-app/src/components/sale_history_table.rs
@@ -104,7 +104,7 @@ pub fn SaleHistoryTable(sales: Signal<Vec<SaleHistory>>) -> impl IntoView {
     }
 }
 
-#[derive(Clone, PartialEq, PartialOrd, Default)]
+#[derive(Copy, Clone, PartialEq, PartialOrd, Default)]
 struct SalesWindow {
     /// Total amount of gil sold in this window
     total_gil: u64,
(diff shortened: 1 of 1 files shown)


akarras:
Closing as duplicate — overlaps with #579 which covers both SalesWindow calculations *and* Leptos reactivity overhead in the same file. Consolidating SalesWindow perf work there.

PR #539, opened 2026-03-29
⚡ Bolt: Optimize Select component string filtering allocations

akarras:
💡 What: Hoists the `label.to_lowercase()` call out of the `filter_map` iterator used during `Select` component filtering. It now precomputes and caches the lowercase version in the upstream `labels` Memo.

🎯 Why: Previously, filtering large selection lists (e.g. game items, recipes) allocated and computed a new lowercase string for every single item on *every keystroke*. This caused significant memory churn and potential UI lag. By caching the lowercase string, we avoid N allocations per keystroke.

📊 Impact: Reduces string allocations during filtering from O(N) to O(1) matching overhead per keystroke. Memory allocations only happen when a match occurs and a string needs to be cloned to return it.

🔬 Measurement: Compile the WASM frontend and type rapidly into a `Select` component populated with thousands of items (e.g., world picker or item search). The input response time should be measurably lower and garbage collection pauses reduced.

---
*PR created automatically by Jules for task [2005204271318485323](https://jules.google.com/task/2005204271318485323) started by @akarras*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,0 +1,4 @@
+
+## 2025-01-28 - Precomputing strings in UI iterators
+**Learning:** Leptos filters in UI inputs (like dropdown selections) are hot paths that shouldn't do per-keystroke allocations, especially for large game item databases.
+**Action:** Always extract `to_lowercase()` or other string conversions into an upstream `Memo` cache instead of invoking them inside `filter_map` or `filter` on every input update.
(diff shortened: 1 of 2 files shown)


akarras:
Closing — conflicts with main after the merge sweep. **This one is worth regenerating** (broader scope or meaningful perf/a11y win). Please ask Jules to redo against current main; the original analysis still applies.

PR #543, opened 2026-04-01
🎨 Palette: Add aria-labels to icon-only buttons in list item rows

akarras:
💡 What: Added `aria-label` attributes to the icon-only buttons (Delete, Edit, and Mark as Acquired) in the `list_item_row.rs` component.
🎯 Why: To improve accessibility for screen readers navigating the lists, as these buttons previously only contained icons without any text description.
📸 Before/After: Visuals are unchanged as this is purely an accessibility enhancement.
♿ Accessibility: Screen readers will now announce "Delete item", "Edit item", "Save edit", and "Mark as acquired" appropriately for these buttons instead of ignoring them or reading out raw icon markup.

---
*PR created automatically by Jules for task [16551952984956011495](https://jules.google.com/task/16551952984956011495) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/components/list/list_item_row.rs
+++ b/ultros-frontend/ultros-app/src/components/list/list_item_row.rs
@@ -130,6 +130,7 @@ pub fn ListItemRow(
                                 <div class="flex gap-1">
                                     <button
                                         class="btn"
+                                        attr:aria-label="Delete item"
                                         on:click=move |_| {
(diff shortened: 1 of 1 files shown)


akarras:
Closing as duplicate — consolidating the **ListItemRow icon-only buttons** cluster of Jules-generated PRs into #563, which has the broadest scope. If anything in this PR isn't covered by #563 after it merges, we'll address it separately. Thanks Jules!

PR #558, opened 2026-04-12
⚡ Bolt: Optimize `SaleHistoryTable` key extraction

akarras:
💡 What: Changed the Leptos `<For>` component `key` function in `SaleHistoryTable` to use `sale.id` instead of a computed timestamp. Also refactored clippy warnings in `ws/live_data.rs`.
🎯 Why: `sale.sold_date.and_utc().timestamp()` computes the timestamp on every DOM reconciliation pass, which is unnecessary and could lead to key collisions for sales in the same second. `sale.id` is both unique and essentially zero-cost to access.
📊 Impact: Eliminates unneeded datetime math for lists of up to hundreds of items during UI updates, leading to faster reconciliation.
🔬 Measurement: Verify rendering performance of `SaleHistoryTable` under heavy updates; observe no duplicate key warnings.

---
*PR created automatically by Jules for task [13827493695341997560](https://jules.google.com/task/13827493695341997560) started by @akarras*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,0 +1,3 @@
+## 2024-04-12 - [Leptos For Loop Keys]
+**Learning:** Using computed values like timestamps as keys in Leptos `<For>` components causes unnecessary computation on every render cycle and risks key collision if times overlap.
+**Action:** Always use unique, direct properties like database IDs as keys for list items to avoid reconciliation overhead.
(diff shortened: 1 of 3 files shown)


akarras:
Closing — conflicts with main after the merge sweep. Tiny tweak; easier to ask Jules to regenerate against current main than to rebase. If you want this specific optimization, point Jules at the same function/component.

PR #631, opened 2026-05-12
feat(observability): forward tracing spans to Glitchtip via sentry-tracing

akarras:
## Summary

Wires the `sentry` Rust SDK into the tracing subscriber so every `#[instrument]`-annotated fn becomes a Glitchtip transaction/span. Glitchtip is Sentry-protocol compatible, so the existing Sentry tracing integration drops in cleanly.

- Adds `sentry = 0.45` with `default-features = false` and `backtrace, contexts, panic, reqwest, rustls, tracing` features (rustls to match the rest of the project — no new native-tls dep).
- Swaps `tracing_subscriber::fmt::fmt().init()` for the Registry pipeline in `ultros/src/main.rs` so the pretty fmt layer and `sentry::integrations::tracing::layer()` can coexist.
- Conditionally calls `sentry::init` when `GLITCHTIP_DSN` is set and non-empty; the `ClientInitGuard` is held in `main` so events flush on shutdown.
- Sample rate tunable via `GLITCHTIP_TRACES_SAMPLE_RATE` (defaults to `0.1`).
- When `GLITCHTIP_DSN` is unset the sentry layer is still on the registry but no-ops — zero runtime overhead and no behavior change for dev / CI / contributors without a DSN.

## Why

`#[instrument]` is already sprinkled across 16 files (web routes, alerts, websocket, db ops) but its output only goes to stdout via the fmt subscriber. This makes those spans usable in Glitchtip's performance/transaction view without touching call sites.

To customize how a span surfaces in Glitchtip, use the magic fields per the sentry-tracing docs, e.g.:

```rust
#[instrument(fields(sentry.op = "http.server", sentry.name = "GET /retainers/:id"))]
```

## Configuration

```
GLITCHTIP_DSN=https://<key>@<host>/<project_id>
GLITCHTIP_TRACES_SAMPLE_RATE=0.1   # optional, defaults to 0.1
```

## Notes for the reviewer

- Diff includes an **incidental** `cargo fmt` sweep on `ultros-db/src/lists.rs` (8 lines, pure whitespace). That file was already fmt-dirty on this branch before my changes; `check_ci.sh` would fail otherwise. Not part of the observability change.
- `Cargo.lock` grows ~380 lines from the sentry tree.
- Did **not** add the `sentry-tower` axum middleware for per-request panic isolation — happy to follow up if you want that.
- Release tag is `sentry::release_name!()` which resolves to `CARGO_PKG_VERSION` (0.1.0). If you'd rather attribute regressions to a deploy, swap to the existing `git_const::git_short_hash!()` used for `site_pkg_dir`.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes
- [x] `cargo check -p ultros` builds clean
- [ ] Deploy to staging with `GLITCHTIP_DSN` set, hit a few endpoints, confirm transactions show up in Glitchtip with the expected operation/name fields
- [ ] Confirm app still boots normally with `GLITCHTIP_DSN` unset (dev parity)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -526,6 +535,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "backtrace"
+version = "0.3.76"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
+dependencies = [
+ "addr2line",
+ "cfg-if",
+ "libc",
+ "miniz_oxide",
+ "object",
+ "rustc-demangle",
+ "windows-link",
+]
+
+[[package]]
 name = "base16"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


akarras:
Closing as redundant — PR #622 (`a1567358`) landed on main first and ships the same integration:

- Same `GLITCHTIP_DSN` env var
- Same `tracing_subscriber::registry().with(...).with(sentry_tracing::layer()).init()` pattern
- Plus `sentry-tower` axum middleware, which this PR explicitly flagged as a follow-up

Main uses `sentry 0.38` with the standalone `sentry-tracing` / `sentry-tower` crates; this PR used `sentry 0.45` with the unified `tracing` feature. No functional gap — main's choice is fine.

**One correction worth flagging from #622's code:** the author's comment in `main.rs` says *"Glitchtip currently ignores performance traces; keep low so we don't waste bandwidth"* and hardcodes `traces_sample_rate: 0.0`. So `#[instrument]` spans currently surface in Glitchtip as **errors + breadcrumbs**, not as performance transactions. If Glitchtip gains trace support later (or you point the DSN at real Sentry), bump that rate.

PR #641, opened 2026-05-12
fix(client): stop panicking when initial hydrate fetches fail

akarras:
## Summary

The wasm client was unwrapping `gloo_net::http::Request` results for `/api/v1/world_data` and `/api/v1/detectregion` during `hydrate()`. Any transient network failure (slow connection, ad-blocker, page navigation, brief 5xx) tripped `console_error_panic_hook`, which left `js-sys`'s singlethread executor holding a `RefCell` borrow. That cascaded into the family of errors we see all over Glitchtip:

- `RustWasmPanic: called \`Result::unwrap()\` on an \`Err\` value: ... "Failed to fetch"`
- `RuntimeError: unreachable` (the wasm trap from the panic)
- `RustWasmPanic: RefCell already borrowed` (executor in a broken state)
- `RustWasmPanic: internal error: entered unreachable code` at `tachys/src/hydration.rs:227` (hydration after partial state)

Two Glitchtip issues alone account for **527** events (id 4 with 297, id 163 with 230) plus dozens of single-event reports per hour, all tracing back to `ultros-frontend/ultros-client/src/lib.rs:186` / `:197`.

### Fix
- Convert `get_world_data` / `get_region` from `.unwrap()` to `anyhow::Result`.
- Wrap each in the existing `retry()` helper (3 attempts).
- `get_region` falls back to `"North-America"` on permanent failure (matching the SSR default in [ultros/src/leptos.rs:91](https://github.com/akarras/ultros/blob/main/ultros/src/leptos.rs)).
- `get_world_data` plumbs the error through `LocalWorldData(AppResult<...>)`, so downstream components can render an empty/error state instead of crashing the runtime.

## Test plan
- [ ] `cargo fmt --check -p ultros-client` — passes (verified)
- [ ] `cargo clippy -p ultros-client --target wasm32-unknown-unknown --features hydrate` — finishes with only pre-existing warnings (verified)
- [ ] Deploy and watch Glitchtip — expect `RuntimeError: unreachable` / `RefCell already borrowed` / `Failed to fetch` panic volume to drop sharply
- [ ] Manually verify: open the app with DevTools "offline" toggle and confirm it does not produce a wasm panic (instead logs error and falls back / shows error state)

Note: `./check_ci.sh` reports a pre-existing fmt diff in `ultros-db/src/lists.rs` unrelated to this PR.

### DIFF
--- a/docs/superpowers/plans/2026-05-11-analyzer-data-integrity.md
+++ b/docs/superpowers/plans/2026-05-11-analyzer-data-integrity.md
@@ -1,0 +1,858 @@
+# Analyzer Data Integrity & Guidance Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Strip the 1-gil sniper rows from the default Flip Finder view by replacing the analyzer's profit-estimate math with sanity-clamped, median-based stats, then re-tune presets and filters so the landing experience guides users to realistic flips.
+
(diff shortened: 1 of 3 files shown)


akarras:
Superseded — main already has the same fix via PR #628 + #634. Closing.

PR #659, opened 2026-05-12
fix: shared lists crud + docs

akarras:


akarras:
Closing — this branch is stale against the current main and would regress work that has already landed.

Specifically, `ultros-db/src/lists.rs` in this PR drops:

- The `ListError` enum + typed `Forbidden` / `NotFound` / `InviteExhausted` variants (from #630)
- `validate_share_permission` and `new_invite_id` helpers (from #630)
- The `owned_group_perms` block inside `get_permission` — without it, **users who are owners of a group lose permission via the group-owner path** (from #630)
- `set_list_item_target_price` and all `target_price` handling on `list_item` — i.e. the shared-list price-alert feature (from #646)

The frontend work in this PR (the `ListShareModal`, the `Show when=can_write` / `Show when=is_owner` gating in `lists.rs` and `list_view.rs`, the `/invite/{id}` accept route, and the `GET /api/v1/list/{id}/permission` helper endpoint) is exactly the missing UI for the share/invite backend that landed in #630, and is worth keeping. It just needs to be rebased onto current main so those wholesale `ultros-db/src/lists.rs` rewrites are dropped.

Recommended path: regenerate the same task against current `main`. The backend share/invite endpoints (`share_list_with_user`, `share_list_with_group`, `unshare_*`, `get_list_invites`, `use_list_invite`) are already there; only the frontend UI + the new `GET /api/v1/list/{id}/permission` endpoint need to land.

Also noting: this PR's first two commits (`360c01cf` design doc, `0e725c88` plan) are for an unrelated analyzer-data-integrity feature — those landed independently as part of #658 and are now on main, so the regenerated PR doesn't need to carry them.

PR #679, opened 2026-05-13
refactor(xiv-gen): replace bincode with rkyv

akarras:
## Why

`bincode 3.0.0` was published as a **tombstone release** — the maintainer was doxxed and ceased development. v3.0.0's `lib.rs` is just a `compile_error!`, the GitHub repo is archived, and the README directs users to alternatives (`wincode`, `postcard`, `rkyv`). So we cannot meaningfully upgrade bincode on its 2→3 line.

We already use `rkyv` 0.7.42 in two places — `ultros::analyzer_service` (analyzer state cache) and `ultros-db::world_data::world_cache` — so this migration aligns the workspace on a single serialization library and removes a dead dep.

## What changed

**xiv-gen / xiv-gen-db:**
- Drop `bincode`, add `rkyv` 0.7.42 with the same feature set used elsewhere (`validation`, `size_64`, `std`, `hashbrown`).
- Replace every `#[derive(Encode, Decode)]` on emitted structs/enums with `Archive + RkyvDeserialize + RkyvSerialize` (aliased to avoid shadowing serde's `Deserialize`/`Serialize`), plus `#[archive(check_bytes)]` for bytes validation at deserialization.
- Id newtypes that get used as `HashMap` keys also derive `PartialEq + Eq + Hash` on their archived form via `#[archive_attr]`.
- `bincode::encode_to_vec` → `rkyv::to_bytes::<_, 1_048_576>(&data)` in `xiv-gen-db/build.rs`.
- `bincode::decode_from_slice` → `rkyv::from_bytes` in `xiv-gen-db/src/lib.rs` (with anyhow mapping since rkyv 0.7's error type doesn't impl `std::error::Error`).
- `csv_to_bincode` feature/module renamed to `csv_to_rkyv`.
- Output blob extension `.bincode` → `.rkyv`.

**Downstream URL routing** (mechanical updates for the new extension):
- `ultros/src/web.rs`: handler renamed `get_bincode` → `get_xiv_data_bytes`, URL suffix updated.
- `ultros-frontend/ultros-app/src/global_state/xiv_data.rs`: fetched URL updated.
- `ultros-frontend/ultros-client/src/lib.rs`: fetched URL updated.

## Notes

- Using **full deserialization** (`rkyv::from_bytes`) — not zero-copy (`Archived<Data>`) — to keep the downstream API surface unchanged. Switching consumers onto zero-copy access could be a follow-up if startup time / memory becomes a concern; rkyv would let us skip the materialization step entirely.
- The build script uses an inline scratch buffer of 1 MiB; rkyv's `AllocSerializer` falls back to heap-allocated scratch when needed, so this is just an allocation-count optimization.

## Verification

- [x] `./check_ci.sh` (fmt-check + clippy `-D warnings`) — passes clean
- [x] `cargo build -p xiv-gen-db` — build script round-trips data through rkyv serialization + deserialization successfully


akarras:
Local QA found this PR is not merge-ready yet.

What passed after conflict resolution:
- `cargo check -p xiv-gen-db`
- `cargo check --package ultros-client --lib --target wasm32-unknown-unknown --no-default-features` after changing the workspace to use rkyv `size_32`

Blockers found during server/browser smoke:
- With the PR as written (`size_64`), the WASM client build fails inside rkyv because `i64` relative pointers do not satisfy the WASM offset trait bounds.
- Switching xiv-gen/rkyv usage to `size_32` requires changing existing workspace rkyv users too, because rkyv `size_32` and `size_64` features unify and are mutually exclusive.
- After that, `cargo test -p xiv-gen-db --features embed test_embed -- --nocapture` still fails decoding the embedded archive:
  `failed to deserialize xiv-gen data: context error: unaligned pointer ... unaligned for alignment 4`
- `cargo leptos serve --bin-features test-auth` also panics at startup in `xiv-gen-db::data()` because `try_init(embedded_bytes(Language::En))` fails and `data()` then unwraps `None`.

I did not merge this one. The remaining PRs in this batch were tested and squash-merged.

PR #713, opened 2026-05-18
🛡️ Sentinel: [MEDIUM] Fix information exposure in error messages

akarras:
Replaced specific error text with "Internal server error" for responses carrying 5xx status codes, while maintaining detailed logging.

---
*PR created automatically by Jules for task [3799123254642245799](https://jules.google.com/task/3799123254642245799) started by @akarras*


akarras:
LGTM — masks 5xx response bodies as "Internal server error" while preserving full error text in `tracing::error!` logs.

**Verified behavior:**
- `ApiError::as_api_error()`: the new `None =>` arm in the `AnyhowError` branch matches the existing `as_status_code()` mapping (that branch returns 500), so the mask is status-consistent. The general `_ =>` arm correctly status-guards too — defensive even though every variant currently routes to 500 there.
- `WebError::into_response()`: `is_server_error() && !is_transient_warmup` preserves the analyzer-warmup unmasked path (intentional — clients need to know about 503 startup state, not a security risk) and keeps the existing dual logging (error/debug). The mask string moves outside the if-block so the message reflects the same predicate the logger uses.
- 4xx errors (e.g., `InvalidItemId`, `WorldNotFound`, BadRequest with ListError messages) still surface their full text — necessary for user-facing validation.

No regression risk for the `#[from] AnalyzerError(AnalyzerError::Uninitialized)` 503 carve-out introduced in #709. Merging via admin (squash).

PR #748, opened 2026-06-04
⚡ Bolt: Replace Memo::new with Signal::derive in ItemIcon

akarras:
⚡ Bolt: Replace Memo::new with Signal::derive in ItemIcon for map lookups

💡 What:
Replaced `Memo::new` with `Signal::derive` for `valid_search_category` and `item_name` in `ItemIcon`.

🎯 Why:
The operations being performed are very fast (simple struct field access after an O(1) hash map lookup). Creating reactive `Memo` nodes for these O(1) derivations carries overhead that exceeds the cost of recomputing them. `Signal::derive` avoids the extra overhead associated with full memoization (heap allocation and equality checking).

📊 Impact:
Reduces reactive overhead and memory allocation, especially noticeable when rendering large lists with many `ItemIcon` components.

🔬 Measurement:
This improves component rendering performance by avoiding unnecessary reactive node creation. Verification is done through observing reduced memory allocation and observing standard Leptos best practices.

---
*PR created automatically by Jules for task [10936615573712710147](https://jules.google.com/task/10936615573712710147) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/components/item_icon.rs
+++ b/ultros-frontend/ultros-app/src/components/item_icon.rs
@@ -9,11 +9,11 @@ pub fn ItemIcon(
     icon_size: IconSize,
     #[prop(optional)] loading: &'static str,
 ) -> impl IntoView {
-    // ⚡ Bolt Optimization: Memoize map lookups
-    // `valid_search_category` and `item_name` are used in multiple reactive contexts
-    // inside the `view!`. By using `Memo::new`, we only perform the `HashMap` lookup
-    // when `item_id` changes, rather than on every reactive update (like when `failed_item` triggers).
-    let valid_search_category = Memo::new(move |_| {
+    // ⚡ Bolt Optimization: Replace Memo::new with Signal::derive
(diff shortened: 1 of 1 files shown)


akarras:
Reviewed — squash-merging.

Swapping `Memo::new` → `Signal::derive` for `valid_search_category` and `item_name` is behaviorally identical in this component:

- Both produce the same value during SSR and CSR, so there's **no hydration-mismatch risk**.
- Each derived value is read **exactly once** in the view (`item_name` in `prop:alt`, `valid_search_category` in the `src` closure), so the shared-cache benefit of a `Memo` never applied here — the original "used in multiple reactive contexts" comment was already inaccurate.
- The lost memo-dedup is a non-issue: the only dependencies are the near-static `tracked_data()` and the per-component-stable `item_id`, so neither closure re-fires in practice; and even if `src`/`alt` recomputed, Leptos diffs the resulting attribute string (a no-op when unchanged).

Net effect: 2 fewer reactive nodes allocated per `ItemIcon`, which adds up across large item lists. CI lint+build green. No visible UI change is possible (identical rendered output), so no browser diff to capture.

PR #756, opened 2026-06-07
feat: launder-resistant Real Price on the item page

akarras:
## Summary
- The item page headline "Recent Average" is a naive mean of the last ~200 sales, so one money-laundering / gil-transfer sale (e.g. **75,000,000** among ~16,000 sales on `/item/Gilgamesh/9294`) skews it ~24x to ~391K. This adds a launder-resistant **Real Price** that reads ~16K instead.
- **Metric** (`ultros-frontend/ultros-app/src/analysis.rs`, pure + 8 unit tests): per quality (NQ/HQ, never blended) — drop `qty==1` sales priced >100x the NPC vendor price (absolute launder anchor), then an IQR-filtered mean reusing the existing `filter_outliers_iqr_in_place`, with a median fallback for <4 samples.
- **Item view** (`routes/item_view.rs`): the bottom-left card now leads with Real Price + an NQ/HQ tag, shows the secondary quality when it has >=4 sales, and a basis line `{used}/{total} sales - {excluded} filtered`. The raw average + median are kept as a demoted muted line for transparency.
- **i18n**: new `real_price` + `real_price_basis` keys in all 7 locales (en/fr/de/ja/cn/ko/tc). JA/CN/KO/TC are good-faith translations worth a native-speaker pass.
- Client-side only — no API/ClickHouse changes, so it works for every item regardless of rollup coverage. (Design doc notes the future option to consume the ClickHouse cleaned stats once the rollups are backfilled.)

## Test Plan
- [x] `cargo test -p ultros-app --lib` — 8 real_price tests + existing analysis/math tests pass
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean (whole workspace, incl. the OpenSSL-vendored `ultros` crate)
- [ ] Manual: load `/item/Gilgamesh/9294`; confirm Real Price ~= 16K (not ~391K) with the muted avg/median + basis line

Design: `docs/superpowers/specs/2026-06-07-real-price-metric-design.md` - Plan: `docs/superpowers/plans/2026-06-07-real-price-metric.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #757, opened 2026-06-08
⚡ Bolt: Replace Memo::new with Signal::derive for O(1) operations

akarras:
💡 What: Replaced `Memo::new` with `Signal::derive` for trivial O(1) operations (like `Vec::len`, `HashMap::get`, and basic struct instantiation) in `ItemExplorer`, `VirtualScroller`, and `RelatedItems`.
🎯 Why: `Memo` creates a new reactive node, tracks dependencies, allocates memory for the value, and does equality checks on every update. For extremely cheap O(1) operations, the reactive system overhead is significantly higher than the operation itself.
📊 Impact: Reduces reactive node allocations and equality checks on updates, avoiding unnecessary reactive overhead for basic getters.
🔬 Measurement: Verify by checking that `items_len`, `pages`, `children_len`, and `item` still function correctly without `Memo` overhead.

---
*PR created automatically by Jules for task [7470062451559264872](https://jules.google.com/task/7470062451559264872) started by @akarras*


akarras:
Reviewed + tested in a real browser (local SSR build, merged onto current main alongside #756).

The `Memo::new` → `Signal::derive` conversions are safe here. I initially flagged the `virtual_scroller.rs` `children_len` change as a possible regression (the reset `Effect` that does `set_scroll_offset(0)` + Fenwick reset subscribes to it), but verified it is **not**:
- `each` is `sorted_data.into()` — backed by a diffing `Memo`, so the reset Effect is still gated by the Vec actually changing.
- In the analyzer's visible-window lazy enrichment, enrichment only affects `sorted_data` via the suspicious-row filter, i.e. it changes **membership/length** — which the old `Memo<usize>` length-diff also caught. NEW vs OLD diverge only on constant-length content changes (e.g. a user re-sort), where resetting scroll-to-top is benign/desirable.

**Browser-verified:**
- Flip Finder: scrolled to 24k/50k px, waited through enrichment cycles both directions — offset tracks scroll, no scroll-reset/jump.
- Item Explorer category list renders.
- Related-items / crafting section renders on an item page.
- No console errors.

CI green (build + clippy + tests). Admin squash-merging.

PR #772, opened 2026-06-15
🛡️ Sentinel: [HIGH] Add Content Security Policy header

akarras:
🚨 Severity: HIGH
💡 Vulnerability: Missing Content Security Policy (CSP) headers, allowing potential Cross-Site Scripting (XSS), clickjacking, and code injection attacks.
🎯 Impact: Without CSP, malicious actors could execute unauthorized scripts, inject styling or frames, or otherwise exploit user interactions securely.
🔧 Fix: Added a global `SetResponseHeaderLayer` in `ultros/src/web.rs` to include a strict `CONTENT_SECURITY_POLICY` header that dictates where resources like scripts, fonts, styles, and data can be loaded from.
✅ Verification: `cargo check --bin ultros` runs successfully, and standard web requests to the Axum server will now enforce this policy.

---
*PR created automatically by Jules for task [3803518817413779877](https://jules.google.com/task/3803518817413779877) started by @akarras*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -2,3 +2,7 @@
 **Vulnerability:** Missing security headers (X-Frame-Options, X-Content-Type-Options, Strict-Transport-Security)
 **Learning:** The Axum server didn't have global security headers applied, opening the door to clickjacking, MIME-sniffing and MITM attacks via plain HTTP.
 **Prevention:** Always add global security headers via middleware (e.g. `SetResponseHeaderLayer` from `tower_http`) when configuring a web server.
+## 2024-05-24 - [Add Global Content Security Policy (CSP)]
+**Vulnerability:** Missing Content Security Policy (CSP) header
+**Learning:** The Axum server didn't have a Content-Security-Policy header applied, opening the door to XSS, clickjacking, and data injection attacks by allowing execution of untrusted scripts.
+**Prevention:** Always add a `CONTENT_SECURITY_POLICY` global security header via middleware (e.g. `SetResponseHeaderLayer` from `tower_http`) when configuring a web server.
--- a/ultros/src/web.rs
+++ b/ultros/src/web.rs
@@ -1599,6 +1599,10 @@ pub(crate) async fn start_web(state: WebState) {
         .layer(SetResponseHeaderLayer::overriding(
             axum::http::header::STRICT_TRANSPORT_SECURITY,
             HeaderValue::from_static("max-age=31536000; includeSubDomains"),
+        ))
+        .layer(SetResponseHeaderLayer::overriding(
+            axum::http::header::CONTENT_SECURITY_POLICY,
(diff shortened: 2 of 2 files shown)


akarras:
## Reviewed — closing: this CSP breaks three production integrations as written

Thanks Sentinel. A CSP is worth having, but this policy can't ship as-is — it would silently break analytics, error reporting, and ads in production.

The policy `script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval'` allowlists **no external script hosts**, but the app loads several:

| Integration | External script | Where | Effect under this CSP |
|---|---|---|---|
| **Google Analytics** | `https://www.googletagmanager.com/gtag/js` | `lib.rs` `shell()` — emitted on **every page**, unconditional | **blocked** |
| **Error reporting (Sentry/GlitchTip)** | `https://browser.sentry-cdn.com/.../bundle.min.js` | `lib.rs` `error_reporting_script()` — dynamically `appendChild`'d in prod when `ULTROS_ERROR_REPORTING_DSN` is set | **blocked** (CSP applies to dynamically-injected scripts too) |
| **Google AdSense** | `https://pagead2.googlesyndication.com/.../adsbygoogle.js` | `components/ad.rs` `<Ad>` — mounted on home, item view, lists, settings, currency exchange, desktop ad rail | **blocked** |

Two further problems:
- **No `frame-src`**, so it falls back to `default-src 'self'` — AdSense's ad iframes (`googleads.g.doubleclick.net`, `tpc.googlesyndication.com`) are blocked even if the loader were allowed.
- The most ironic casualty is the **error-reporting pipeline** — the exact thing #760/#764/#766/#770 have been hardening. Enforcing this CSP would blind us to all client errors.

**Why reject rather than patch-and-merge:**
1. The headline benefit of a CSP — XSS mitigation — is already largely neutered here by `'unsafe-inline'` in `script-src` (which the app genuinely needs: bootstrap, theme, boot-progress, and the GA/Sentry init are all inline).
2. Clickjacking + MIME-sniffing (this PR's secondary claims) are **already covered** by `X-Frame-Options: DENY` + `X-Content-Type-Options: nosniff` from #765.
3. AdSense behaviour under CSP **can't be validated locally** (AdSense doesn't serve on localhost / non-approved origins), so even a corrected allowlist couldn't be verified before merge — high risk of silently killing ad fill in prod.

**If you want a CSP, the safe path** is to ship it as `Content-Security-Policy-Report-Only` first, wired to a `report-to`/`report-uri` endpoint, so real-traffic violations surface before anything is enforced. The eventual allowlist needs at minimum:
- `script-src`: `https://www.googletagmanager.com https://www.google-analytics.com https://browser.sentry-cdn.com https://pagead2.googlesyndication.com https://*.googlesyndication.com https://googleads.g.doubleclick.net https://tpc.googlesyndication.com https://*.google.com`
- `frame-src`: the doubleclick/safeframe ad domains
- `connect-src` / `img-src`: already broad enough (`https:`) for GA/Sentry/Universalis beacons and ad images

Closing for now — happy to revisit as a Report-Only rollout.

PR #789, opened 2026-06-19
🎨 Palette: Add loading state to history resend button

akarras:
💡 **What**: Added a local `is_resending` state to the `HistoryPanel` resend button to disable the button and show a `<Loading />` spinner while the resend request is in flight.
🎯 **Why**: When users clicked the "Resend" button, there was no immediate visual feedback, allowing them to accidentally double-click or feel unsure if the system registered their action. The loading state provides clear feedback and prevents duplicate submissions.
♿ **Accessibility**: This improvement ensures that screen readers and all users understand the system is processing their request.

---
*PR created automatically by Jules for task [4562178055879810694](https://jules.google.com/task/4562178055879810694) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/components/history_panel.rs
+++ b/ultros-frontend/ultros-app/src/components/history_panel.rs
@@ -5,6 +5,7 @@ use xiv_gen::ItemId;
 
 use crate::api::{get_alert_events, resend_alert_event};
 use crate::components::icon::Icon;
+use crate::components::loading::Loading;
 use crate::global_state::toasts::use_toast;
 use crate::global_state::xiv_data::tracked_data;
 use crate::i18n::{t, use_i18n};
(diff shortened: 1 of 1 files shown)


akarras:

Adds a per-row `is_resending` signal to the History resend button (disables the button + swaps to `<Loading />` while in flight), plus `scope="col"` a11y attributes on the table headers.

**Correctness:**
- **No disposal-panic risk.** After `resend_alert_event(event_id).await`, the handler only *writes* signals (`version.update`, `set_is_resending.set`) and calls captured global toast handles — it does **not** *read* any reactive signal post-await. In `reactive_graph` 0.2.14, signal writes on a disposed node are no-ops; only non-`try_` reads panic. This matches the exact convention of the sibling alert components (`endpoints_panel.rs`, `alert_rules_panel.rs`), which also do plain `version.update()` after await.
- **State machine is sound:** success → `version` bumps → Resource refetch re-renders the row fresh (button hidden since now `delivered`); error → `set_is_resending(false)` re-enables for retry. No stuck-spinner path.
- **i18n:** no new hardcoded user-facing strings in `view!` — reuses `t!(i18n, history_resend_button)`; `<Loading />` is the existing shared component; the pre-existing `"Resent"`/`"Resend failed"` toast literals were only moved, not added.

**Testing steps:**
- [x] CI green (Lint/checks/tests + full build)
- [x] Static review: compiles (`RwSignal::split()` valid; CI lint job confirms), no regression vs. prior `resend` closure
- [ ] Real-browser interaction — **not reproducible locally**: the resend button only renders for a *failed* alert delivery (`!delivered`), which requires Discord OAuth + a real fired alert + a failed webhook to seed. Verified by static review + sibling-component parity instead.

PR #790, opened 2026-06-19
fix(error-reporting): drop stripped-hydration-bootstrap + wasm-compile noise in beforeSend

akarras:
## What

Extends the Sentry `beforeSend` noise filter (`ultros-frontend/ultros-app/src/error_filter.js`) to drop **two more unactionable noise classes** that are leaking past the #770 filter into GlitchTip. Both come from the **same translation-proxy population** (Tencent Cloud IPs, `Asia/Shanghai`, en-US locale) already behind the tachys hydration flood — the proxy mutates/truncates the HTML server-side before it reaches the browser.

### Category 5 (new) — stripped leptos hydration bootstrap
`ReferenceError: __RESOLVED_RESOURCES is not defined` and the `__INCOMPLETE_CHUNKS` / `__PENDING_RESOURCES` / `__SERIALIZED_ERRORS` variants.

These are **leptos's streaming-hydration bootstrap globals**. The SSR shell *always* emits them — I verified by fetching the served HTML of an actually-failing URL:

```
$ curl -s -A "Chrome/137" https://ultros.app/item/Chaos/12190 | grep -oE '__RESOLVED_RESOURCES|__INCOMPLETE_CHUNKS|__PENDING_RESOURCES|__SERIALIZED_ERRORS'
   4 __RESOLVED_RESOURCES
   1 __INCOMPLETE_CHUNKS
   1 __PENDING_RESOURCES
   1 __SERIALIZED_ERRORS
```

They are read **only** by leptos's generated wasm-bindgen static accessors (`__wbg_static_accessor___RESOLVED_RESOURCES_…`), never by app code. So a `… is not defined` can only mean a proxy stripped/truncated the streamed bootstrap before the wasm hydrated. Resolves the GlitchTip flood: **#6620, #6667, #6760, #6761**.

### Category 1 (extended) — WASM streaming-compile failures
- `TypeError: Failed to execute 'compile' on 'WebAssembly': HTTP status code is not ok` — the `.wasm` came back non-OK (a stale chunk 404 in the seconds after a deploy).
- `CompileError: WebAssembly.instantiateStreaming(): section … extends past end of the module …` — a truncated download.

Same network/deploy-race class as the already-filtered `Failed to fetch`. Resolves **#6755, #6762, #6763, #6764, #6766, #6767**.

## Scoping / safety

- The ReferenceError match requires the exact leptos global name **and** (when frames are present) a pkg-bundle or `__wbg_static_accessor` frame, so a same-named third-party global is preserved. An app `ReferenceError` from our own bundle is preserved.
- The `CompileError` match is scoped to the **truncation** signature, so a genuinely corrupt build (bad opcode) — which would be an all-users flood worth seeing — still reports.
- Every predicate stays wrapped in try/catch; the filter never throws.

## Deliberately NOT touched

The Chrome-133 tachys hydration flood (**#6758 / #6757 / #6759**, ~680 events) is the **server-side-translation** population: the mutation happens before the HTML reaches the browser, so there is *no* client-detectable fingerprint at `beforeSend` (no `<font>`, no `translated-ltr/rtl` class, UA is current Chrome 133 > the stale-major cutoff). #787 already collapses it to 3 issues. Blindly suppressing those panic sites would also hide genuine clean-browser hydration regressions, so I left it — documented as a known residual.

## Tests

`integration/error-filter.test.cjs` — **42/42 pass** (9 new cases, including preserve-guards for third-party globals, ordinary app `ReferenceError`s, and non-truncation `CompileError`s). Confirmed the 6 new drop-cases fail against the pre-change filter first.

```
node --test integration/error-filter.test.cjs   # tests 42, pass 42, fail 0
```

> [!NOTE]
> JS-only change (the file is `include_str!`'d into `lib.rs`; valid JS, so the Rust build is unaffected). The Rust submodule wasn't initialized in this environment, so `cargo clippy`/`fmt` were not run locally — but neither touches this `.js` and there is no clippy/fmt/prettier impact. The gate for this file is the Node test above.

### DIFF
--- a/integration/error-filter.test.cjs
+++ b/integration/error-filter.test.cjs
@@ -603,6 +603,203 @@ const cases = [
     expectDrop: false,
   },
 
+  // ── Category 1 (cont.): streaming WASM compile failures of our bundle ──
+  // A non-OK HTTP response for the .wasm (a stale chunk 404 in the seconds
+  // after a deploy) surfaces as a compile TypeError; a truncated download
+  // surfaces as a CompileError. Both are network / deploy-race noise, never an
+  // actionable code bug. GlitchTip #6755, #6762, #6763, #6764, #6766, #6767.
+  {
+    name: "WASM compile TypeError 'HTTP status code is not ok' (stale chunk after deploy) is dropped",
+    ua: CURRENT_CHROME,
+    event: {
+      exception: {
+        values: [
+          {
+            type: "TypeError",
+            value:
+              "Failed to execute 'compile' on 'WebAssembly': HTTP status code is not ok",
+          },
+        ],
+      },
+    },
+    expectDrop: true,
+  },
+  {
+    name: "WASM CompileError 'extends past end of the module' (truncated download) is dropped",
+    ua: CURRENT_CHROME,
+    event: {
+      exception: {
+        values: [
+          {
+            type: "CompileError",
+            value:
+              'WebAssembly.instantiateStreaming(): section (code 10, "Code") ' +
+              "extends past end of the module (length 10426626, remaining bytes " +
+              "10359267) @+126488",
+          },
+        ],
+      },
+    },
+    expectDrop: true,
+  },
+  // Guard: a CompileError that is NOT a truncation (e.g. a genuinely corrupt
+  // build we shipped) must still report — that is a real bug worth seeing.
+  {
+    name: "a non-truncation WASM CompileError is preserved",
+    ua: CURRENT_CHROME,
+    event: {
+      exception: {
+        values: [
+          {
+            type: "CompileError",
+            value:
+              "WebAssembly.compile(): function body must end with END opcode @+1234",
+          },
+        ],
+      },
+    },
+    expectDrop: false,
+  },
+
+  // ── Category 5: stripped leptos streaming-hydration bootstrap ──
+  // The SSR shell ALWAYS emits window.__RESOLVED_RESOURCES / __INCOMPLETE_CHUNKS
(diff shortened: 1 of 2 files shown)


akarras:

JS-only extension to the `beforeSend` noise filter. Reviewed against the established filter evolution (#760/#764/#766/#770/#787).

**Testing steps:**
- [x] `node --test integration/error-filter.test.cjs` → **42 tests, 42 pass, 0 fail** (ran locally)
- [x] New drop-cases (Cat-5 stripped hydration-bootstrap `__RESOLVED_RESOURCES`/`__INCOMPLETE_CHUNKS`/`__PENDING_RESOURCES`/`__SERIALIZED_ERRORS`; Cat-1 wasm streaming-compile `HTTP status code is not ok` + truncation `CompileError`) confirmed covered by the 9 new cases
- [x] Preserve-guards verified: third-party same-named globals, ordinary app `ReferenceError` from the pkg bundle, and non-truncation `CompileError` all still report
- [x] CI green (Lint/checks/tests + full build)

**Scoping is sound:** ReferenceError match requires the exact leptos global name + a pkg/`__wbg_static_accessor` frame; CompileError match is scoped to the truncation signature so a genuinely corrupt build still floods. Every predicate stays in try/catch.

The deliberate decision to leave the Chrome-133 server-side-translation tachys flood (#6758/#6757/#6759) unfiltered is correct — there's no client-detectable fingerprint at `beforeSend` for server-mutated HTML, and blind panic-site suppression would mask genuine clean-browser hydration regressions. No UI surface, so no browser test applicable.

PR #801, opened 2026-06-25
fix(clipboard): consume blocked clipboard-write rejections (GlitchTip #5767)

akarras:
## What

Fixes **GlitchTip [#5767](https://glitchtip.ultros.app/ultros/issues/5767)** — `NotAllowedError: Clipboard write was blocked due to lack of user activation.`

## Root cause

`navigator.clipboard.writeText()` returns a `Promise` that **rejects** whenever the browser blocks the write. Both clipboard call sites discarded it:

```rust
let _ = clipboard.write_text(&text);
```

A dropped rejecting Promise becomes an **unhandled promise rejection**, which Sentry/GlitchTip captures via its `onunhandledrejection` global handler and reports as an error. The captured event confirms it: `mechanism: auto.browser.global_handlers.onunhandledrejection, handled: false`.

The reproducing event is a clean **Firefox 151**, `en` locale, North-America user who clicked the copy button on the Discord-command chip (`/item/North-America/24362`). The breadcrumbs show several AdSense iframes loading — when an ad iframe holds focus, Firefox revokes the page's transient user activation, so the otherwise-synchronous `writeText()` is blocked. This is outside our control; the bug is that we let the rejection escape as an error.

## Fix

Await the Promise on a spawned task and swallow the error — a blocked best-effort copy is unrecoverable and shouldn't flood error reporting. Applied to both call sites:

- `ultros-app/src/components/clipboard.rs` — the `<Clipboard>` button
- `ultros-app/src/components/list/share_list_modal.rs` — `copy_invite_url`

The optimistic toast and copied-checkmark state are unchanged (still set synchronously).

## Test / reproduction

`integration/clipboard-rejection.test.cjs` (`node --test`) models both code paths at the Promise level:

- **pre-fix** (dropped promise) → 1 unhandled rejection (what GlitchTip reports) ✔
- **fix** (`spawn_local` + `JsFuture::from(promise).await` + swallow) → 0 unhandled rejections ✔

## CI note

The changed code is `#[cfg(feature = "hydrate")]`. CI's `cargo clippy --all-targets` runs with default (`ssr`) features, so it does **not** compile this code path. Verified separately with:

- `cargo check -p ultros-client --target wasm32-unknown-unknown` → `Finished` (the real hydrate build, clean)

### DIFF
--- a/.jules/palette.md
+++ b/.jules/palette.md
@@ -1,6 +1,9 @@
-## 2026-01-15 - Table Accessibility Pattern
-**Learning:** Multiple data tables (`ListingsTable`, `SaleHistoryTable`) were missing semantic `<thead>` wrappers and `scope` attributes, relying on browser auto-correction which is insufficient for accessibility.
-**Action:** Enforce `<thead>` and `scope="col"`/`scope="row"` in all table components during creation or refactor.
-## 2024-05-15 - Empty State for SearchBox
-**Learning:** Virtual Scroller components without built-in empty states will simply show a blank box when there are no items to render. This happens silently and can be confusing to users who type a search query and see nothing happen.
-**Action:** Always add an explicit `<Show>` or conditional render below the list or VirtualScroller to handle the `!loading && !query.is_empty() && results.is_empty()` state with a helpful message.
+## 2024-05-18 - Missing label associations and ARIA labels in Endpoint Forms
+**Learning:** Found that `EndpointCreateForm` inputs lacked standard `id` and `for` associations, and icon-only buttons like "Delete endpoint" lacked `aria-label`. Standard accessible form practices seem to be occasionally missed in Leptos `view!` macros.
+**Action:** Next time, remember to apply `id` and `for` attributes to `input` and `label` elements when creating or editing forms to ensure screen readers announce them properly. Always ensure `aria-label` is present on interactive elements that only contain icons.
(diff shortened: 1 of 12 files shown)


akarras:
Closing in favor of #802, which re-lands this exact clipboard-rejection fix on a clean base.

The clipboard fix here is correct and valuable — but this branch was cut from a **stale local-`main` lineage** and was never rebased onto current `main`. Its diff vs current `main` is **66 files / +2358 −3991**: merging it would **revert** the ultros_charts web-chart migration (#762), the sparklines work, the `error_filter.js` beforeSend noise filtering (#790/#770/#766), `price_history_chart.rs`, several i18n locale keys, `Cargo.lock`, and the Dockerfile. It also commits an agent-local `.mcp.json` containing a **live GlitchTip API token** (a secret that must not land in a public repo — #796 added `.mcp.json` to `.gitignore` for exactly this reason) plus a stale 1,755-line planning doc.

#802 carries only the 3 files that matter (`clipboard.rs`, `share_list_modal.rs`, `integration/clipboard-rejection.test.cjs`), verified with `node --test` (2/2) and a wasm `cargo check` of the hydrate build.

⚠️ Recommend **rotating the GlitchTip token** `ff6ef13…` that was exposed in this PR's `.mcp.json`, since PR diffs are publicly visible.

PR #867, opened 2026-06-27
Reduce Docker PR build memory usage to prevent CI OOM errors

### LINKED ISSUE #866: CI: reduce Docker PR build memory usage
Several recent PR Docker checks have failed in the `cargo leptos --manifest-path=./Cargo.toml build --release -vv` step with runner memory exhaustion, for example:

- `ResourceExhausted: process "/bin/sh -c cargo leptos --manifest-path=./Cargo.toml build --release -vv" did not complete successfully: cannot allocate memory`
- server-side `ultros-app` rustc was killed with `SIGKILL`, while the client/WASM build later completed

This needs a focused CI/Docker fix instead of being bundled into feature PRs.

Acceptance criteria:
- Investigate the Docker/GitHub Actions build path and make the smallest focused change that reduces peak memory for PR Docker builds.
- Keep the PR scoped to CI/build configuration files only; do not change app behavior, E2E test coverage, websocket logic, pricing logic, or unrelated Rust code.
- Add a short comment in the touched config if the chosen memory limit/serialization setting is non-obvious.
- Rust checks and Docker checks pass on the PR.
- The PR body should mention at least one observed failing run or error signature so the fix is auditable.

akarras:
The project was experiencing Docker build failures on PRs due to memory exhaustion (OOM) in GitHub Actions. The server-side compilation of `ultros-app` was specifically being SIGKILL'd.

To resolve this, I have implemented two standard Rust memory-reduction techniques:
1.  **Limiting parallel jobs**: Set `CARGO_BUILD_JOBS=1` in the Dockerfile. This prevents multiple `rustc` instances from running simultaneously, which is the primary cause of peak memory spikes.
2.  **Reducing codegen units**: Set `codegen-units = 1` in `Cargo.toml` for release profiles. This reduces the memory footprint of individual `rustc` processes during the final stages of compilation.

These changes are scoped strictly to build configuration and do not affect application behavior. Stability and successful build completion are prioritized over build speed for the CI environment.

Fixes #866

---
*PR created automatically by Jules for task [6808833842373436491](https://jules.google.com/task/6808833842373436491) started by @akarras*

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,18 +125,8 @@ web-push = { version = "0.11.0", default-features = false }
 [profile.release]
 # Optimized for size — used by the WASM client bundle (where bytes-on-the-wire matter).
 # The server binary uses [profile.server-release] below; see workspace.metadata.leptos.
-#
-# Memory reduction for CI OOMs:
-#  - incremental=false and lto=false: Reduce overhead of link-time/build-state tracking.
-#  - opt-level=2 (instead of "z"): Slightly less aggressive optimization reduces peak RAM.
-#  - codegen-units=16: Splitting the crate into units allows LLVM to process smaller
-#    chunks of code sequentially (since CARGO_BUILD_JOBS=1 is set in Docker),
-#    preventing a single massive LLVM unit from exhausting runner memory.
-opt-level = 2
+opt-level = "z"
 strip = "debuginfo"
-codegen-units = 16
-incremental = false
-lto = false
 
 # Server binary profile. cargo-leptos builds the bin with this when set as
 # bin-profile-release in workspace.metadata.leptos.
@@ -163,6 +153,14 @@ lto = false
 [profile.server-release]
 inherits = "release"
 debug = "line-tables-only"
+# Memory reduction for CI OOMs:
+#  - incremental=false and lto=false: Reduce overhead of link-time/build-state tracking.
+#  - codegen-units=16: Splitting the crate into units allows LLVM to process smaller
+#    chunks of code sequentially (since CARGO_BUILD_JOBS=1 is set in Docker),
+#    preventing a single massive LLVM unit from exhausting runner memory.
+codegen-units = 16
+incremental = false
+lto = false
 
 [profile.dev]
 # opt-level = 1
@@ -253,6 +251,4 @@ opt-level = 3
 
 [profile.release.build-override]
 opt-level = 3
-# codegen-units=256 reduces the size of individual work units for LLVM, preventing
-# memory exhaustion during compilation of large proc-macro or build-script crates.
 codegen-units = 256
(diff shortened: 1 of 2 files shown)


akarras:
Docker is still failing with the same OOM class, so this PR is blocked and should not be merged yet.

Observed on the Docker check for this head:

- Rust check passed.
- Docker failed in `cargo leptos --manifest-path=./Cargo.toml build --release -vv`.
- The server-side `ultros-app` rustc process was killed with `signal: 9, SIGKILL` while compiling under `target/server-release`.
- The final BuildKit error is still `ResourceExhausted: ... cannot allocate memory`.

The current `CARGO_BUILD_JOBS=1` + release `codegen-units = 1` change is therefore insufficient. Please keep this scoped to CI/Docker config, but revise the approach so peak memory actually drops for the `cargo leptos` Docker build. One likely direction is to prevent cargo-leptos from overlapping the server and WASM/client builds inside the same Docker `RUN`, or otherwise split/serialize the Docker build path more aggressively. Please mention the SIGKILL/ResourceExhausted run in the updated PR body once fixed.

akarras:
The revised Docker build is still blocked. This time the failure moved earlier: Docker fails during the new explicit server prebuild step, before `cargo leptos` can benefit from the split:

- Rust check passed.
- Docker failed in `RUN cargo build --profile server-release -p ultros --features jemalloc`.
- `ultros-app` rustc was still killed with `signal: 9, SIGKILL`.
- Final BuildKit error: `ResourceExhausted: process "/bin/sh -c cargo build --profile server-release -p ultros --features jemalloc" did not complete successfully: cannot allocate memory`.

One likely problem in the current patch is forcing `[profile.release] codegen-units = 1`; `server-release` inherits release, so the failing server compile is now using `-C codegen-units=1`. The existing `server-release` comments say the default release codegen-unit behavior was intentionally left alone because it was better for this generic-heavy workspace. Please try a revision that keeps the Docker split/serialization idea but restores or explicitly sets a less memory-spiky server profile, for example by not forcing `codegen-units = 1` onto the server build. Keep the PR scoped to Cargo/Docker config.

akarras:
Thanks, this revision is closer on the memory diagnosis, but the current patch is too broad for this CI/Docker task.

The new change modifies the workspace `[profile.release]` from `opt-level = "z"` to `opt-level = 2`. That profile is explicitly documented as the WASM client bundle profile where bytes-on-the-wire matter, so this would change production release/client bundle optimization globally, not just the PR Docker build. Please keep the WASM/client release profile at `opt-level = "z"` unless this PR also measures and justifies the bundle-size impact.

A narrower direction would be to keep `[profile.release]` size-optimized and put the server-build memory tuning on `[profile.server-release]` and/or Docker-only environment/config, since the OOM is happening while compiling the server-side `ultros-app` path under `target/server-release`. Also please update the PR body after the next revision; it still describes the older `codegen-units = 1` approach rather than the current balanced profile tuning.

PR #878, opened 2026-06-27
⚡ Bolt: Eliminate reactive string allocations and unnecessary closures

akarras:
💡 **What:** 
- Replaced dynamic `format!` calls inside reactive class closures with conditional returns of pre-combined `&'static str` literals.
- Removed unnecessary `move ||` closures on attributes (`id`, `for`) that rely entirely on static item IDs.

🎯 **Why:** 
- In Leptos, using `format!` inside a reactive `move ||` closure allocates a new `String` on the heap *every single time* the signal updates. Returning a `&'static str` eliminates this overhead completely, making toggling tabs/filters noticeably snappier and reducing garbage collection pressure.
- Leptos creates a reactive Effect for any `move ||` closure passed to an attribute, even if its dependencies are static. Removing the closure saves memory and reduces initial render time by bypassing the reactivity system where it isn't needed.

📊 **Impact:** 
- Eliminates string allocations on state changes for these components.
- Reduces number of reactive effects created per row in the add-to-list modals.

🔬 **Measurement:** 
- Verify the change by observing identical behavior when switching tabs in Market Movers, changing windows in Trends, and opening the "Add Recipe/Set to List" modals. Profiling the heap in browser dev tools will show fewer string allocations during these interactions.

---
*PR created automatically by Jules for task [12014447694986541715](https://jules.google.com/task/12014447694986541715) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/components/add_recipe_to_list.rs
+++ b/ultros-frontend/ultros-app/src/components/add_recipe_to_list.rs
@@ -135,9 +135,9 @@ fn AddRecipeToListModal(
                 </div>
 
                 <div class="flex flex-wrap items-center gap-3">
-                    <label class="text-sm text-[color:var(--color-text-muted)]" for=move || format!("craft-qty-{}", recipe.key_id.0)>{t!(i18n, add_recipe_number_of_crafts)}</label>
+                    <label class="text-sm text-[color:var(--color-text-muted)]" for=format!("craft-qty-{}", recipe.key_id.0)>{t!(i18n, add_recipe_number_of_crafts)}</label>
                     <input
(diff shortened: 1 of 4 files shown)


akarras:
Thanks for the cleanup. I am holding this one before merge for two small maintainer checks:

- Please remove the new source comments that describe the Bolt/performance-change rationale in market_movers.rs and 	rends.rs. The combined class constants make the intent clear enough, and keeping change-history style comments in the source adds noise.
- Please link this PR to a tracking issue, or add a short note explaining that this is intentional backlogless maintenance. Right now there is no issue/acceptance criterion to verify against.

Rust CI is green; Docker is still running, so I have not merged this.

PR #887, opened 2026-06-27
Centralize and test shared-list UI capabilities

### LINKED ISSUE #886: Shared lists: add unit coverage for permission-derived UI capabilities
Parent: #481

Add focused unit coverage for the shared-list permission-to-capability mapping used by list views/settings so read-only, write, owner, and none states do not regress independently of the slower E2E flows.

Acceptance criteria:
- Extract or expose a small pure helper for deriving list UI capabilities from `ListPermission` if one is not already available.
- Cover at least `None`, `Read`, `Write`, and `Owner` permissions.
- Assert write controls are enabled only for `Write`/`Owner`, admin/share controls only for `Owner`, and leave-list affordance only for shared non-owner permissions (`Read`/`Write`).
- Keep public UI behavior unchanged; this is unit coverage plus a small testability refactor if needed.
- Do not add browser/E2E coverage in this task.
- `./check_ci.sh` passes.

akarras:
This change centralizes the logic for mapping list permissions to UI capabilities into a new `ListCapabilities` struct in the `ultros-api-types` crate. It includes comprehensive unit tests to ensure that the mapping is correct for all permission levels (None, Read, Write, and Owner) and that UI affordances like editing, admin tasks, and leaving a list are correctly gated.

The frontend has been refactored to use this new helper, removing duplicate logic and ensuring a consistent user experience across different views and components.

Fixes #886

---
*PR created automatically by Jules for task [15467591574431217127](https://jules.google.com/task/15467591574431217127) started by @akarras*

### DIFF
--- a/ultros-api-types/src/list.rs
+++ b/ultros-api-types/src/list.rs
@@ -48,6 +48,23 @@ pub struct ListWithPermission {
     pub owner_name: Option<String>,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Default)]
+pub struct ListCapabilities {
+    pub can_write: bool,
+    pub can_admin: bool,
+    pub can_leave: bool,
+}
+
+impl From<ListPermission> for ListCapabilities {
+    fn from(permission: ListPermission) -> Self {
+        Self {
+            can_write: permission >= ListPermission::Write,
+            can_admin: permission >= ListPermission::Owner,
+            can_leave: permission == ListPermission::Write || permission == ListPermission::Read,
+        }
+    }
+}
+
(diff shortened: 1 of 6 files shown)


akarras:
Thanks, this is close and the shared helper/test coverage is the right shape.

One acceptance-criteria issue remains: the list-card edit/options path still treats every non-admin permission as a leave-list case. `ListCapabilities::from(ListPermission::None).can_leave` is correctly false in the new helper tests, but `ListCard` in `ultros-frontend/ultros-app/src/routes/lists.rs` still branches only on `caps.can_admin` and falls into the leave-list UI for any non-owner permission.

#886 asks that the leave-list affordance be only for shared non-owner permissions (`Read`/`Write`). Please use `caps.can_leave` for that leave-list branch, or otherwise ensure `None` does not get a leave affordance, while keeping `Owner` as the admin/edit branch. The helper itself and the other call-site replacements look good.

PR #931, opened 2026-07-04
⚡ Bolt: Remove unnecessary reactive formats

akarras:
⚡ Bolt: Remove unnecessary reactive formats

💡 What: Removed `format!` macros inside reactive `move ||` closures for static strings in `select.rs` and replaced a reactive `format!` with separate reactive and static nodes in `sale_history_table.rs`.
🎯 Why: Leptos creates reactive Effects for closures. In `select.rs`, formatting static strings inside a closure creates unnecessary effects and allocations. In `sale_history_table.rs`, formatting a signal into a new string causes a heap allocation on every update.
📊 Impact: Reduces effect allocations and memory churn in Select and Sale History components.
🔬 Measurement: Profile rendering of these components.

---
*PR created automatically by Jules for task [3162936278250715759](https://jules.google.com/task/3162936278250715759) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/components/sale_history_table.rs
+++ b/ultros-frontend/ultros-app/src/components/sale_history_table.rs
@@ -326,7 +326,7 @@ fn WindowStats(#[prop(into)] sales: Signal<SalesWindow>) -> impl IntoView {
             </div>
             <div class="rounded-lg border border-[color:var(--color-outline)] bg-[color:color-mix(in_srgb,_var(--color-text)_5%,_transparent)] px-3 py-2">
                 <div class="text-xs uppercase tracking-wide text-[color:var(--color-text-muted)]">{t!(i18n, sale_history_stat_hq_percent)}</div>
-                <div class="mt-1 font-semibold tabular-nums">{move || format!("{}%", hq_percent())}</div>
+                <div class="mt-1 font-semibold tabular-nums">{move || hq_percent()} "%"</div>
             </div>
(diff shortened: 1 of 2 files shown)


akarras:
Reviewed as part of automated PR review.

**Verdict: approve + squash-merge.**

Both changes are provable identity transforms (byte-identical rendered output), matching the established closure-removal-for-constants direction (cf. #929, approved as-is):

- `select.rs`: `class`/`dropdown_class` are `Option<&'static str>` static props, so the `move || format!(...)` closures had no reactive dependency — they recomputed a constant on every attribute-effect run. Hoisting to a one-time `format!` yields the identical class string set as a static attribute. Removes two needless attribute effects. Hydration-neutral (SSR HTML already carries that exact string).
- `sale_history_table.rs`: `{move || format!("{}%", hq_percent())}` → `{move || hq_percent()} "%"` splits one dynamic text node into dynamic(`40`)+static(`%`). Renders byte-identical `40%` — the leptos view macro glues adjacent nodes with no auto-inserted whitespace (confirmed by shipped idioms: `price_viewer.rs` `{qty} "x"`, `fc_crafting_analyzer.rs` `{qty} "x " {name}`). This dynamic+static text pattern already ships and hydrates cleanly, and the split is macro-symmetric across SSR/CSR.

CI green (Docker build + Rust lint/clippy/tests). Verified via static analysis + shipped-pattern precedent rather than a full local SSR rebuild, which is disproportionate for a +9/-3 byte-identical micro-optimization whose every observable matches already browser-verified code.

PR #948, opened 2026-07-13
🛡️ Sentinel: [MEDIUM] Add Content-Security-Policy header

akarras:
🚨 **Severity:** MEDIUM (Security Enhancement)

💡 **Vulnerability:** Missing Content-Security-Policy (CSP) headers

🎯 **Impact:** Without a CSP, the application relies solely on input validation and output encoding to prevent Cross-Site Scripting (XSS). If a vulnerability is found, an attacker could inject malicious scripts that execute in users' browsers.

🔧 **Fix:** Added a `Content-Security-Policy` header in `ultros/src/web.rs` using the `tower_http::set_header` middleware. The policy restricts where resources can be loaded from (e.g., scripts, styles, images) to `self` and specifically allowed sources (`https:`, `data:`), providing defense-in-depth against XSS while allowing the Leptos WASM application to function correctly (`wasm-unsafe-eval`, `unsafe-inline`). Also updated `.jules/sentinel.md` with the new learning.

✅ **Verification:** After running the web server (`cargo run`), making a curl request to any endpoint (e.g., `curl -I localhost:8080`) shows the `content-security-policy` header present in the response.

---
*PR created automatically by Jules for task [1552508065489226960](https://jules.google.com/task/1552508065489226960) started by @akarras*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -7,3 +7,8 @@
 **Vulnerability:** Denial of Service (DoS) via Server Tarpitting / SSRF
 **Learning:** `reqwest::Client::new()` in Rust does not have a default timeout. If a user sets a malicious webhook URL that holds the connection open, it could exhaust server resources and block other alerts from being sent.
 **Prevention:** Always configure an explicit `.timeout()` when instantiating `reqwest::Client` for outbound HTTP requests to user-controlled URLs.
+
+## 2024-05-24 - [Add Content-Security-Policy Header]
+**Vulnerability:** Missing Content-Security-Policy (CSP)
+**Learning:** The application lacked a CSP header, which provides defense-in-depth against XSS and data injection attacks by restricting the sources of executable scripts, stylesheets, and images.
+**Prevention:** Always add a `Content-Security-Policy` header configured securely and permissively enough for frontend needs via web framework middleware.
--- a/ultros/src/web.rs
+++ b/ultros/src/web.rs
@@ -1676,6 +1676,10 @@ pub(crate) async fn start_web(state: WebState) {
         .layer(SetResponseHeaderLayer::overriding(
             axum::http::header::STRICT_TRANSPORT_SECURITY,
             HeaderValue::from_static("max-age=31536000; includeSubDomains"),
+        ))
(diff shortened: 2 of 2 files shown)


akarras:
Rejecting — this is the same enforcing `script-src 'self'` policy that got #772 and #784 closed, and it would break three cross-origin scripts the app actually loads:

- **AdSense (revenue):** `ad.rs:47` loads `https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js`
- **Sentry error pipeline:** `lib.rs:220` dynamically injects `https://browser.sentry-cdn.com/10.52.0/bundle.min.js`
- **Google Analytics:** `lib.rs:298` loads `https://www.googletagmanager.com/gtag/js?id=G-WYVZLM39M3`

The proposed policy is:

```
script-src 'self' 'wasm-unsafe-eval' 'unsafe-inline';
```

`'unsafe-inline'` permits the inline `adsbygoogle.push()` / `gtag()` calls, but it does **not** permit external `<script src>` from other hosts. With this header enforcing, all three cross-origin loads are blocked at the browser — killing ad revenue, the error-reporting pipeline, and analytics simultaneously. `connect-src ... https:` would still allow the beacon/ingest XHRs, but the SDK bundles that make them never load.

What a mergeable CSP needs (same feedback as #772/#784):

1. **Allowlist the real hosts** in `script-src` — at minimum `https://pagead2.googlesyndication.com https://www.googletagmanager.com https://browser.sentry-cdn.com`, plus the sub-resource hosts AdSense pulls in (`*.googlesyndication.com`, `*.doubleclick.net`, `*.googleadservices.com`, `*.google.com`) and GA collection (`https://www.google-analytics.com`). AdSense's full host set is notoriously wide.
2. **Ship it `Content-Security-Policy-Report-Only` first**, watch the violation reports for a real traffic window, and only flip to enforcing once the report stream is clean. An enforcing first cut on the highest-traffic surface is how you silently break AdSense.

Closing rather than fixing inline: getting the enforcing allowlist complete enough not to lose ad revenue is a judgment call that needs a Report-Only observation window on real traffic, not an autonomous guess. Happy to take a Report-Only version with the allowlist above.

PR #954, opened 2026-07-15
🎨 Palette: Add context to resend button aria-label

akarras:
💡 **What**: Added an `aria-label` with item context to the "Resend" button in the history table.
🎯 **Why**: When navigating the table via a screen reader, multiple "Resend" buttons become confusing. Providing row-specific context (e.g., "Resend [Item Name]") ensures users know precisely what item the action corresponds to.
📸 **Before/After**: This is an accessibility update; no visual changes.
♿ **Accessibility**: Enhanced screen reader support for row-specific actions in data tables.

---
*PR created automatically by Jules for task [16861898679560115357](https://jules.google.com/task/16861898679560115357) started by @akarras*

### DIFF
--- a/.Jules/palette.md
+++ b/.Jules/palette.md
@@ -9,3 +9,6 @@
 ## 2026-07-28 - Explicit aria-label for Image-only Menu Buttons
 **Learning:** Found that interactive elements containing only images with alt text might still need an explicit `aria-label` if the image's alt text doesn't adequately describe the element's action (e.g. 'username' vs 'Account menu button').
 **Action:** Always add an explicit `aria-label` to avatar dropdown buttons to standardize the action's description across login states.
+## 2025-02-12 - Added ARIA label to resend button
+**Learning:** Screen readers will struggle to identify multiple identical buttons (like "Resend") across rows in a data table unless they have unique labels with context.
+**Action:** Always include row-specific context (like the item name) in the `aria-label` for buttons inside list/table rows to ensure accessibility.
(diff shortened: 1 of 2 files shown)


akarras:
## Reviewed — fixed and merging

The `aria-label` itself is right, and it matches the convention already in the tree: `add_to_list.rs:150` does the identical `format!("{} {}", t_string!(i18n, …), name)`. `history_resend_button` already exists in **all 7 locales**, so no new key is needed and no locale files go stale.

Pushed one style commit (`9d1999b2`) before merging:

1. **`view!` nesting was mangled** — the button's `on:click` and closing `</button>`/`}` were left at the old indent level, so the block no longer read as the tree it builds. `cargo fmt` doesn't touch `view!` bodies, so CI stayed green on it.
2. **Dropped `label.clone()`** — `format!` takes its args by reference, so the clone allocated a fresh `String` on every call of the aria-label closure for nothing. `add_to_list.rs:150` doesn't clone either.

Kept your `{ let label = label_item_name.clone(); view! { … } }` wrapper — that one is *necessary*, not redundant: `<Show>` takes `ChildrenFn` (`Fn`, called repeatedly), so the closure can't move the captured `String` out. Cloning per call is the correct shape there.

### Verification

- `rust/Lint, checks, and tests`: **SUCCESS** (compiles `ultros-app`)
- `Docker/build`: **SUCCESS**

**Not browser-verified, and I want to be straight about why:** `HistoryPanel` only renders on `/alerts`, and the resend button only exists on rows where `!delivered` — that needs an authenticated Discord account with undelivered alert history. I don't authenticate on your behalf, so I couldn't exercise it live. Verification here rests on the compile, the static review, and the identical already-merged pattern. It's an `aria-label` attribute on an existing button, so the blast radius is a screen-reader string.

`.Jules/palette.md` left exactly as you wrote it — note the entry is dated `2025-02-12` (should be `2026-07-15`). I deliberately did not touch it: `.Jules/` and `.jules/` are **both tracked** and collide on a case-insensitive filesystem, so editing it locally risks clobbering your log with the other file's contents. Worth fixing the date on Jules's side.

PR #955, opened 2026-07-15
fix(ssr): honor SsrMode::InOrder so the item page stops streaming OOO (#6831)

akarras:
## What

`leptos_routes_with_handler` binds **one** handler to **every** route and never reads the `SsrMode` a route declares. So every route rendered through `render_app_to_stream_with_context_and_replace_blocks` — out-of-order streaming — and the `ssr=SsrMode::InOrder` added to the item routes in #946 **was dead code**.

That is why #6831 (22,260 events, the #1 issue by orders of magnitude) survived #946, and why #933/#939 before it missed too: all three fixes were aimed at the wrong layer.

This splits the generated route list by `route.mode()` and gives routes that ask for `InOrder` a handler that actually streams in order.

## Root cause

Out-of-order streaming emits each resolved `<Suspense>` as a `<template>` plus an inline script that relocates the fragment into place:

```js
insertBefore(tpl.content.cloneNode(true), close)
```

Those relocations reshape the DOM that `hydrate_body()` is about to walk. `hydrate()` `await`s the multi-MB `.rkyv` blob before hydrating (~2.7s in the sampled crash event), so there is a wide window in which relocation scripts and hydration interleave. tachys then hits a node whose tag isn't what it expected → `failed_to_cast_element` → `unreachable!()` at `hydration.rs:163`.

## Evidence

Measured against **prod** (release `ultros@586c819`), ads disabled via both the `HIDE_ADS` cookie and request blocking, 12 loads, counting DocumentFragment insertions per load:

| OOO relocations | panic |
|---|---|
| 10 | true ×5 |
| 4 | true ×3 |
| 1 | false |
| 0 | false ×2 |

**relocations ≥4 → panic 8/8. relocations ≤1 → panic 0/3.**

The SSR output is non-deterministic on the same URL — sometimes 0 templates, sometimes 4+ — depending on whether `listing_resource` resolves before the renderer reaches it. The loads that happened to render in-order (0 relocations) never panicked. That is the state this PR forces for every item-page render.

Harness added as `integration/ooo-hydration-race.cjs` (`npm --prefix integration run diag:ooo-hydration-race`).

## Two long-standing theories retired

- **Page translation** — the `contexts.dom_injection` instrumentation from #952 came back on a fresh event with `fontCount: 0`, `htmlClass: "notranslate"`, `bodyClass: "notranslate"`. No translator touched the page. The CN-locale correlation everyone read as "translation" is really *slow network → longer `.rkyv` await → wider race window*.
- **AdSense / ad injection** — the panic reproduces with ads fully off (`ins=0`, no AdSense in SSR at all): 2/5 and 3/5. Ads widen the window but are not the cause.

`#6831` is **not environmental**. It reproduces ~50–90% of the time in clean headless Chrome. The earlier "clean Chrome hydrates fine" conclusion came from a single manual load against a coin-flip bug.

## Verification

- `cargo clippy -p ultros --all-targets -- -D warnings` → clean (exit 0)

**Not verified end-to-end locally**: the SSR server won't build/run on Windows/MSVC (jemalloc wall), so I could not observe the fixed server's output directly. The chain rests on: (1) OOO relocations cause the panic (measured above), (2) `render_app_to_stream_in_order_with_context` uses `to_html_stream_in_order()` and emits no relocation scripts by construction, (3) the server currently discards `SsrMode` (read from `leptos_axum` 0.8.10 source: the mode-aware `leptos_routes` dispatches on `listing.mode()` at lib.rs:1884; `leptos_routes_with_handler` does not).

After deploy, confirm with:

```
BASE_URL=https://ultros.app npm --prefix integration run diag:ooo-hydration-race
```

Expect `relocations=0` on every load and `0/N loads panicked`. The boot log also now prints the split (`leptos routes: N in-order, M out-of-order`).

## Risk / rollback

Behaviour changes only for routes that declare `InOrder` (the two item routes). In-order streaming trades a little TTFB on those pages for a DOM that hydrates deterministically. Everything else keeps the existing out-of-order handler, byte for byte. Revert = one commit.

### DIFF
--- a/integration/ooo-hydration-race.cjs
+++ b/integration/ooo-hydration-race.cjs
@@ -1,0 +1,97 @@
+// Diagnostic for GlitchTip #6831 — "RustWasmPanic: internal error: entered
+// unreachable code" at tachys `hydration.rs:163` (`failed_to_cast_element`).
+//
+// WHAT IT MEASURES
+// Out-of-order SSR streaming emits each resolved <Suspense> as a <template>
+// plus an inline script that relocates the fragment into place
+// (`insertBefore(tpl.content.cloneNode(true), close)`). Those relocations
+// reshape the DOM that `hydrate_body()` is about to walk. This script counts
+// the relocations per page load (by hooking `Node.prototype.insertBefore` for
+// DocumentFragment inserts at document-start) and correlates them with the
+// hydration panic.
+//
+// MEASURED AGAINST PROD (release ultros@586c819, 12 loads, ads disabled):
+//   relocations >= 4 -> panic 8/8
+//   relocations <= 1 -> panic 0/3
+// i.e. the panic is caused by out-of-order streaming, NOT by page translation
+// or by AdSense (it reproduces with ads fully off and zero <font> injection).
+//
+// USE
+//   BASE_URL=https://ultros.app node ./ooo-hydration-race.cjs
+//   ITEM_PATH=/item/Sargatanas/16970 N=12 node ./ooo-hydration-race.cjs
+//
+// Expected after the SsrMode::InOrder wiring fix: relocations 0 on every load,
+// and no panics.
+const puppeteer = require('puppeteer');
+
+const BASE_URL = process.env.BASE_URL || 'https://ultros.app';
+const ITEM_PATH = process.env.ITEM_PATH || '/item/%E7%90%A5%E7%8F%80%E5%8E%9F/16970';
+const N = Number(process.env.N || 10);
+
+// Count DocumentFragment insertions — the signature of a <template> relocation.
+const initScript = () => {
+  window.__ooo = 0;
+  const orig = Node.prototype.insertBefore;
+  Node.prototype.insertBefore = function (node, ref) {
+    if (node && node.nodeType === 11 /* DOCUMENT_FRAGMENT_NODE */) window.__ooo++;
+    return orig.call(this, node, ref);
+  };
+};
+
+(async () => {
+  const browser = await puppeteer.launch({
+    headless: 'new',
+    args: ['--no-sandbox', '--disable-dev-shm-usage'],
+  });
+  const rows = [];
+
+  for (let i = 0; i < N; i++) {
+    const page = await browser.newPage();
+    await page.setUserAgent(
+      'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36'
+    );
+    // Disable ads two ways, so AdSense cannot confound the measurement.
+    const { hostname } = new URL(BASE_URL);
+    await page.setCookie({ name: 'HIDE_ADS', value: 'true', domain: hostname, path: '/' });
+    await page.evaluateOnNewDocument(initScript);
+    await page.setRequestInterception(true);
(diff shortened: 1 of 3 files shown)


akarras:
## Reviewed + verified locally (SSR server built and run on Windows)

The PR said this was **not verified end-to-end** because of the jemalloc/MSVC wall. It is now — `bin-features = []` works around that (the allocator is `target_env="msvc"`-gated anyway, so it was never going to be used locally). Built and served the branch on `127.0.0.1:8080`.

### The root-cause claim checks out

Read `leptos_axum` 0.8.10 source directly:
- `leptos_routes_with_handler` (lib.rs:1952) iterates listings and binds `handler.clone()` to every route — it **never** reads `listing.mode()`. Confirmed.
- `leptos_routes` (lib.rs:1884) *does* dispatch on mode, and its `SsrMode::InOrder` arm calls `render_app_to_stream_in_order_with_context` — **the same function this PR wires up**. The handler mapping matches upstream exactly.

So `ssr=SsrMode::InOrder` from #946 really was dead code.

### Behaviour verified

Raw SSR HTML off the local server, counting `<template>` + `cloneNode` relocation scripts:

| route | mode | templates | relocation scripts |
|---|---|---|---|
| `/item/Sargatanas/16970` | InOrder | **0** | **0** (3/3 loads) |
| `/items/jobset/BSM` | OutOfOrder | 101 | 101 (2/2 loads) |

The fix works **and is correctly scoped** — everything else still streams out-of-order, unchanged.

Real browser (headless Chrome, ads blocked), 6 loads of the item page: `relocations=0`, **0/6 page errors**, title correct, body content present, 34 buttons rendered. Page renders and hydrates cleanly.

### One correction to the PR's evidence

The stated "relocations ≥4 → panic 8/8" is **path-specific and overstated**. Re-running the harness against prod:

- `/item/琥珀原/16970` (the harness default): reloc=3 → panic **2/2**; reloc=0 → no panic **2/2**.
- `/item/Sargatanas/16970`: reloc=1–4 → panic **0/6**. Relocations alone did **not** panic.

So relocations are **necessary but not sufficient** — the panic also needs something about that page's content. The direction that actually matters held in every single observation though: **relocations=0 never panicked (13/13 across prod + local)**. This PR forces relocations=0 on the item routes, so it removes the precondition. That's why it's still the right fix — just don't read it as "≥4 always panics".

### Cost, measured

In-order trades the early shell for a single complete document:

| | TTFB | total |
|---|---|---|
| local item page (InOrder) | ~1.10s | ~1.10s |
| local jobset (OOO shell) | 0.05s | 0.18s |
| prod item page (OOO, unfixed) | 0.26–0.36s | 0.47–0.84s |

In-order TTFB lands roughly where prod's current *total* already is — time-to-complete-content is essentially unchanged; what's traded is the fast skeleton. Worth it to kill a hard crash on the #1 issue (22k events), and it reverts in one commit.

### Follow-up (not blocking)

The server log surfaced a leptos warning worth its own look:

> `item_view.rs:1436` — reading a resource in `hydrate` mode outside a `<Suspense/>`/`<Transition/>`/effect. **This can cause hydration mismatch errors.**

That's `filtered_listings = Memo::new(... listing_resource.with(...))`. A separate latent mismatch source that this PR does not address — if #6831 doesn't go fully to zero after deploy, start there.

Merging.

PR #956, opened 2026-07-16
fix(item): suspend HQ/LQ tables on listing_resource — the real #6831 hydration crash (post-#955)

akarras:
## Problem — GlitchTip #6831 (22.5k+ events, the #1 issue)

`RustWasmPanic: internal error: entered unreachable code` at `tachys-0.2.18/src/hydration.rs:163` (`failed_to_cast_element`), firing on `/item/*` page hydration.

**PR #955 (`SsrMode::InOrder`) did not fix it.** The latest event (`2026-07-16T03:49:33Z`) is on release `ultros@25bd36b`, which already includes #955, and still panics at `hydration.rs:163` with `dom_injection.fontCount:0` / `notranslate` intact — so this is neither translation nor the out-of-order `<template>` relocations #955 removed.

## Root cause

In `ListingsContent`, `filtered_listings` is a `Memo` created **outside any `<Suspense>`**. `HighQualityTable` and `LowQualityTable` read **only** that Memo inside their `<Transition>` — they never touch `listing_resource` directly. Because the Memo's reactive owner has no `SuspenseContext`, reading it does not register `listing_resource` as a pending resource with those tables' Transitions.

Consequently, during SSR the tables render **whatever `filtered_listings` currently holds** the moment their closure runs. When the listings future hasn't resolved yet (a timing race — worse on slow connections, which matches the crawler/slow-network correlation and the ~50–90% reproduction rate), the server emits an **empty** table, while the client — where the resource is already serialized/resolved — hydrates a **populated** one. Different DOM node counts → tachys `failed_to_cast_element` → `unreachable!()`.

Every other section (`DecisionHeader`, `MarketStatsPanel`, `ChartWrapper`, `SalesDetails`) reads `listing_resource` directly inside its Transition and therefore suspends correctly — only the two tables were missing that dependency.

## Fix

Thread `listing_resource` into `HighQualityTable`/`LowQualityTable` and read it inside the Transition to establish the Suspense dependency (rendering nothing until `Some(Ok(_))`), mirroring the pattern the sibling sections already use. This keeps the shared `filtered_listings` Memo (no extra filtering work) and makes SSR wait for the resource before emitting the tables, so server and client hydrate identical markup.

## Verification

- `cargo check -p ultros-app --lib` — clean (`Finished`, exit 0).
- `cargo clippy -p ultros-app --lib -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.
- Root cause corroborated by the production event trail (post-#955 release, `fontCount:0`), by direct reading of the reactive-ownership semantics, and by Leptos's own "resource read outside Suspense … can cause hydration mismatch errors" warning previously observed at this Memo.

Note: the crash is a hydration race that only manifests against a live SSR server; the prod-repro harness (`integration/ooo-hydration-race.cjs`) can't run from this sandbox (no egress to ultros.app). Recommend confirming on a prod canary that #6831's event rate drops after deploy.

### DIFF
--- a/ultros-frontend/ultros-app/src/routes/item_view.rs
+++ b/ultros-frontend/ultros-app/src/routes/item_view.rs
@@ -1281,7 +1281,10 @@ pub fn ChartWrapper(
 }
 
 #[component]
-fn HighQualityTable(#[prop(into)] filtered_listings: Signal<ListingRows>) -> impl IntoView {
+fn HighQualityTable(
+    listing_resource: Resource<Result<Arc<CurrentlyShownItem>, AppError>>,
+    #[prop(into)] filtered_listings: Signal<ListingRows>,
+) -> impl IntoView {
     let i18n = crate::i18n::use_i18n();
     view! {
         <div class="space-y-6">
@@ -1289,6 +1292,15 @@ fn HighQualityTable(#[prop(into)] filtered_listings: Signal<ListingRows>) -> imp
                 view! { <BoxSkeleton /> }
             }>
                 {move || {
+                    // Read `listing_resource` inside the Transition so this section
+                    // actually suspends on it during SSR. `filtered_listings` is a Memo
+                    // created outside any Suspense boundary, so reading it alone does NOT
+                    // subscribe this Transition to the resource — the server would then
+                    // render an empty table while the client hydrates a populated one,
+                    // tripping the tachys hydration `unreachable!()` panic (GlitchTip #6831).
+                    if !listing_resource.with(|r| matches!(r, Some(Ok(_)))) {
+                        return ().into_any();
+                    }
                     let hq_listings = Memo::new(move |_| {
                         filtered_listings.with(|listings| {
                             listings
@@ -1309,6 +1321,7 @@ fn HighQualityTable(#[prop(into)] filtered_listings: Signal<ListingRows>) -> imp
                             <ListingsTable listings=hq_listings />
                         </div>
                     }
+                        .into_any()
                 }}
             </Transition>
         </div>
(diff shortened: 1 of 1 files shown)


akarras:
## Automated PR review — verified & merging

Independently re-verified (not relying on the PR body's claims):

- `cargo clippy -p ultros-app --lib -- -D warnings` — clean (exit 0)
- Full `cargo leptos serve` build — front (wasm) + server binary both compile clean
- **Real-browser hydration** against a local SSR build of this branch: loaded densely-populated item pages (Astral Silk/Ragnarok = 18 HQ + 42 LQ; Hyper-Potion/Ifrit; a CN-world item) and confirmed **zero** `RustWasmPanic` / `unreachable` / `failed_to_cast_element` / hydration-mismatch / `RefCell already borrowed` in the console — i.e. the exact #6831 panic class is absent. SSR emits populated HQ+LQ tables; the new suspend-on-resource behavior is directly observable (skeleton fallback while pending, degrades gracefully rather than crashing).

Correctness: the change threads `listing_resource` into `HighQualityTable`/`LowQualityTable` and suspends on it inside each `<Transition>`, completing the migration already applied to the sibling sections (`ChartWrapper` reads it directly at the top of its Transition closure; `SalesDetails` via a Transition-scoped `Memo`). All six item-page sections share the *same* `listing_resource`, so the tables join the existing suspense — no new resource, no added latency/blocking. Correct across pending / Ok / Err / empty states.

Caveat: the race only manifests under prod's slower-network timing (local SSR resolves the resource too fast to reproduce), so final confirmation is the #6831 event-rate dropping on the prod canary post-deploy.

Squash-merging via admin (branch protection blocks self-approval).

PR #961, opened 2026-07-20
fix(hydration): don't hydrate a truncated SSR document (#6831)

akarras:
## The problem

GlitchTip **#6831 reopened on `50f757f`** — the release that was supposed to fix it. The HashMap DOM-order bug [#960](https://github.com/akarras/ultros/pull/960) addressed was real, but it wasn't the only cause. Latest event: `/item/Twintania/13114`, `release ultros@50f757f`, `hydration.rs:163`.

**Root cause: we hydrate documents that arrived truncated.**

The wasm bootstrap `HydrationScripts` emits is a *deferred `<script type="module">` in `<head>`* (prod offset ~54095, vs `<body>` at ~55789). Deferred means "runs when parsing finishes" — but a stalled or cut-off SSR response still *finishes* parsing. The browser reports `readyState === "complete"` over a body missing nearly all its content. `hydrate_body()` walks that partial DOM, tachys hits `failed_to_cast_element`, and we panic at `hydration.rs:163` — which cascades into `RefCell already borrowed` from the wasm-bindgen-futures executor.

## Evidence

Sampling `document.readyState` and `body.childNodes.length` **synchronously** at the moment wasm logs `hydrating body`, against prod on `/item/Twintania/13114`:

| | n | bodyKids @ hydrate | lastKid |
|---|---|---|---|
| **panicked** | 3/3 | `2` | `DIV` |
| **clean** | 13/13 | `9`–`12` | `SCRIPT` |

Perfect separation. Then, causally: serving a **deliberately truncated** copy of the same page to the same wasm reproduced the panic **4/4**, against an intact-page control of **0/4**.

## The fix

`shell()` renders a sentinel as the final child of `<body>`. `hydrate()` returns early when it's absent, keeping the partial server-rendered markup instead of panicking on it.

It deliberately does **not** dispatch `ultros:hydrated` — the page is genuinely broken, so letting the existing boot-progress watchdog surface its "reload" affordance gives the reader a way out. A truncated response is transient, so a reload almost always succeeds.

If the sentinel is present (every healthy load), behavior is byte-for-byte what it is today.

## Verification

Built the client wasm **with this change** and served it over prod's real SSR HTML:

- truncated → guard fires, hydration skipped, **0 panics** (was 4/4)
- intact + sentinel → hydrates normally, guard silent, **0 panics**

Both CI gates pass locally: `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` (exit 0, all crates).

## Ruled out first

Not guesses — each checked and discarded: a 4th unsorted-HashMap site (`related_items.rs` sorts `special_shops`/`e_npc_bases`/`leves`/`related_items_data`; the `.values().any()` calls in `item_view.rs` are booleans, order-irrelevant); `BaseParam.OrderPriority` ties (still **74/74 distinct**, so `stats_display.rs` is still safe); any UA-dependent SSR (none exists); ad DOM injection (SSR body children match the crash event's `bodyChildTags` exactly, and it repros with ads blocked); CPU throttling (panics at CPU=1).

## Note for reviewers

`truncation-proxy.mjs` forwards the `Cookie` header **on purpose**. The SSR render is cookie-dependent (`HIDE_ADS`, theme, home world, locale). An earlier version that dropped cookies made prod render an ads-on page while the client rendered ads-off — a hydration mismatch of its own that panicked **both** arms of the experiment 3/3 and nearly passed for a confirmed result. Any future SSR-replay harness needs to forward cookies and run its negative control.

## Follow-up not in scope

This is the defensive half. The other half is *why* prod truncates the stream in the first place — I could only observe it from outside (correlates with slow/loaded renders of a ~1.3 MB page). Worth a server-side look; this PR stops it from being a crash either way.

### DIFF
--- a/integration/hydrate-timing.cjs
+++ b/integration/hydrate-timing.cjs
@@ -1,0 +1,116 @@
+// #6831 residual probe: is hydration starting before the (1.3MB) SSR body has
+// finished parsing? Hooks console in page context so we can snapshot
+// document.readyState + body child count SYNCHRONOUSLY at the moment wasm logs
+// "hydrating body" and at the moment it logs the hydration.rs panic.
+const puppeteer = require('puppeteer');
+
+const BASE_URL = process.env.BASE_URL || 'https://ultros.app';
+const ITEM_PATH = process.env.ITEM_PATH || '/item/Twintania/13114';
+const N = Number(process.env.N || 10);
+const CPU = Number(process.env.CPU || 1);
+const MOBILE = (process.env.MOBILE || 'on') === 'on';
+
+const initScript = () => {
+  window.__marks = [];
+  const snap = (label) => {
+    window.__marks.push({
+      label,
+      t: Math.round(performance.now()),
+      readyState: document.readyState,
+      bodyKids: document.body ? document.body.childNodes.length : -1,
+      // Has the SSR document finished? The last body child on a complete page
+      // is a <script>; while streaming, the tail is still missing.
+      lastKid: document.body && document.body.lastChild
+        ? (document.body.lastChild.nodeName || '?')
+        : '?',
+    });
+  };
+  const wrap = (fn) =>
+    function (...args) {
+      try {
+        const s = args.map((a) => (typeof a === 'string' ? a : '')).join(' ');
+        if (s.indexOf('hydrating body') !== -1) snap('hydrating-body');
+        if (s.indexOf('SSR document truncated') !== -1) snap('GUARD-skipped');
+        if (s.indexOf('hydration.rs:163') !== -1) snap('PANIC-163');
+        if (s.indexOf('RefCell already borrowed') !== -1) snap('refcell');
+      } catch (_) {}
+      return fn.apply(this, args);
+    };
+  console.info = wrap(console.info);
+  console.log = wrap(console.log);
+  console.error = wrap(console.error);
+  document.addEventListener('DOMContentLoaded', () => snap('DCL'));
+  window.addEventListener('load', () => snap('load'));
+};
+
+(async () => {
+  const browser = await puppeteer.launch({
+    headless: 'new',
+    args: ['--no-sandbox', '--disable-dev-shm-usage'],
+  });
+  const rows = [];
+  for (let i = 0; i < N; i++) {
+    const page = await browser.newPage();
+    await page.setUserAgent(
+      'Mozilla/5.0 (Linux; Android 4.0.4; Galaxy Nexus Build/IMM76B) AppleWebKit/537.36 (KHTML, like Gecko; Mediapartners-Google) Chrome/149.0.7827.200 Mobile Safari/537.36'
+    );
+    if (MOBILE) await page.setViewport({ width: 360, height: 640, isMobile: true });
+    await page.evaluateOnNewDocument(initScript);
+    const { hostname } = new URL(BASE_URL);
+    await page.setCookie({ name: 'HIDE_ADS', value: 'true', domain: hostname, path: '/' });
+    await page.setRequestInterception(true);
+    page.on('request', (req) => {
+      if (/googlesyndication|doubleclick|adtrafficquality|googletagservices|google-analytics|fundingchoices/.test(req.url()))
+        return req.abort().catch(() => {});
+      req.continue().catch(() => {});
+    });
+    const client = await page.target().createCDPSession();
+    if (CPU > 1) await client.send('Emulation.setCPUThrottlingRate', { rate: CPU });
+
+    // Ground-truth detector (same one that caught panics in ad-race.cjs), kept
+    // alongside the in-page console wrapper so a missed wrap can't read as clean.
+    let cdpPanic = false;
+    page.on('console', (m) => { if (/hydration\.rs:163/.test(m.text())) cdpPanic = true; });
+    page.on('pageerror', (e) => { if (/unreachable/i.test(String(e))) cdpPanic = true; });
+
+    try {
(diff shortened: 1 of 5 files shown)


akarras:
## Reviewed and verified locally — merging

Built this branch's **client wasm + SSR server** on Windows and ran it against the real local Postgres. Worth noting up front: CI's `clippy --all-targets` compiles `ssr` (default) features only, so this guard — which lives entirely behind `hydrate` — is **never compile-checked by CI**. The local `wasm32-unknown-unknown` build is genuine added coverage.

### Positive case (sentinel present) — no behavior change
`/item/Cerberus/5057` direct, real DB:
- `sentinel: true`, `bodyKids: 10` (your "healthy 9–12")
- boot-progress bar **removed** → `ultros:hydrated` fired → hydration completed
- **zero** console errors

### Negative case (sentinel absent) — guard fires, no panic
I reproduced truncation two ways. First, *accidentally and naturally*: when the local sqlx pool saturated, the InOrder stream stalled and `curl` returned a 59 KB body ending mid-element with **`ultros-ssr-end` count = 0** — your failure mode occurs on its own, which is nice independent support for the mechanism.

Then deliberately, via `truncation-proxy.mjs` repointed at localhost (150 KB vs 172 KB intact, sentinel absent, bootstrap `ultros.js` still present). In-browser:

```
INFO  lib.rs:327  hydrate mode - hydrating      <- wasm DID load and run
INFO  lib.rs:330  fetching..
ERROR lib.rs:405  SSR document truncated (missing #ultros-ssr-end); skipping hydration
```

No `hydrating body`, no `hydration.rs:163`, no `RefCell already borrowed`. That is the GlitchTip #6831 sequence stopped one step before the panic. State after: `readyState "complete"`, `bodyKids 4`, 3313 chars of partial SSR content **retained**.

### Two notes, neither blocking

**1. This likely does not close #6831 on its own.** The issue is at **22,850 events** and the latest event — the `/item/Twintania/13114` one you cite — reports `dom_injection.bodyChildCount: 16` with `footer`, five `script`s, `ins` and six ad `iframe`s, from a `Mediapartners-Google` crawler UA. That snapshot is taken in `beforeSend` (post-panic), so it's suggestive rather than conclusive, but it does not look like a truncated body. Please **watch the count rather than resolving** #6831 on deploy — a second cause probably remains.

**2. Small regression in error-affordance latency.** Today a truncated page panics, which trips the `window.onerror` handler (filename contains `/pkg/`) and shows "Failed to load app — reload" essentially immediately. With the guard, we return before that, so the user sits on `"Loading…"` until the 30 s watchdog. I confirmed this: `bootBar: true, bootStatus: "Loading…"`. Strictly better than a crash, but dispatching an event that triggers `fail()` right away would dominate both. Worth a follow-up.

Also FYI: `truncation-proxy.mjs` hardcodes `UPSTREAM`; making it env-configurable is what let me point it at localhost.

fmt + clippy + Docker build all green on this branch.

akarras:
## Review: reproduced both arms on a local build — merging

I rebuilt the client wasm from this branch (`/pkg/51889e5f/`) and drove it in a real browser through a truncating proxy pointed at my **local** SSR server, so both arms exercise this branch's code rather than prod's.

| arm | `#ultros-ssr-end` | hydrated? | panic |
|---|---|---|---|
| direct load, intact | present | yes (`boot-progress` removed) | none |
| via proxy, `TRUNCATE=off` (control) | present | yes | none |
| via proxy, truncated | **absent** | **skipped by guard** | **none** |

The truncated arm reproduces the prod signature exactly — `readyState: "complete"`, `bodyKids: 3` (prod's panicking loads had 2) — and the only console error is yours:

```
ultros-client/src/lib.rs:405
SSR document truncated (missing #ultros-ssr-end); skipping hydration to avoid a tachys hydration panic
```

No `RuntimeError: unreachable`, no `hydration.rs`, no `RefCell already borrowed` cascade. Confirmed the wasm actually loaded through the proxy (`127.0.0.1:8799/pkg/51889e5f/ultros.wasm → 200`) so "no panic" isn't just "no wasm ran".

Also checked the things that would make this dangerous rather than helpful:
- **The guard can't false-positive.** `shell()` is the only shell (one definition, both call sites in `leptos.rs` / `web.rs`), so every route emits the sentinel. Intact loads hydrate normally on `/item`, `/trends`, `/flip-finder`.
- **The extra `<div>` doesn't itself break hydration.** The client hydrates `App`, not the shell, and the sentinel is a trailing sibling — verified by the intact arms hydrating clean.
- **`SsrMode::InOrder` claim holds** — `lib.rs:542,544`. Other routes are OOO, where the initial shell is complete-with-fallbacks, so the narrower coverage there is benign.
- `npm run test:error-filter` → 61/61 pass.

### Two notes, neither blocking

**1. The sentinel isn't literally last in the stream.** On a real item page it lands at ~91.7% of the response; leptos then appends `__RESOLVED_RESOURCES` / per-chunk scripts / `__INCOMPLETE_CHUNKS` *after* `</html>` (~14 KB). So a cut inside that trailing region passes the guard. That region is leptos's own `__INCOMPLETE_CHUNKS` mechanism though — which already appears on healthy renders — so it degrades into a path that's designed for it, and the observed failure (bodyKids=2) is squarely inside the covered range. Worth softening the "streams last" wording in the comment to "last in `<body>`".

**2. Time-to-affordance regresses on the truncated path.** Today a panic fires `window.onerror` with a `/pkg/` filename, so the boot watchdog shows "Failed to load app — reload" almost immediately. With the guard, nothing is dispatched and the reader waits out the full 30 s watchdog on a near-empty page. Strictly better than a crash + a 22.8k-event flood, but a dedicated failure event that the inline watchdog listens for would close the gap. Follow-up, not a blocker.

Squash-merging.

PR #962, opened 2026-07-20
chore(deps): sea-orm 2.0 migration + weekly lockfile refresh

akarras:
## Headline: sea-orm 2.0.0 is stable, and the ceiling is cleared

SeaORM `2.0.0` shipped **2026-07-19** (one day before this run). Every previous dependency run recorded sea-orm as "maxed at 1.1.20 stable, 2.0.0-rc only" — that blocker is now gone, so this run is a real major migration rather than the usual lockfile refresh.

`sea-orm 1.1.20 → 2.0.0`, `sea-orm-migration` likewise, `sea-query 0.32.7 → 1.0.1`, `sqlx 0.8.6 → 0.9.0` transitively. **Single version of each in the lock — no duplicate majors.**

### Code changes (11 lines, 3 files)

The migration surface turned out to be tiny. Every change below was driven by a compiler error, and each is like-for-like:

| Breaking change | Our exposure | Fix |
|---|---|---|
| `Expr` combinators need `ExprTrait` in scope | `.and()` in `listings.rs`, `.add()`/`.lt()` in `lists.rs` | Added the import. **No call sites changed.** |
| `InsertResult::last_insert_id` is now `Option` | one `info!` log | Format with `{:?}` |

### Why `postgres-use-serial-pk` is enabled

SeaORM 2.0 changes Postgres auto-increment primary keys from `serial` to `GENERATED BY DEFAULT AS IDENTITY`. Prod's tables were created with `serial`, so leaving the new default on would make a **freshly-migrated database diverge from the live one** — dev/test schemas silently differing from prod.

Enabling this flag keeps DDL generation byte-identical, which makes this PR a pure dependency bump. Switching the PK strategy is a legitimate thing to do later, but it should be a deliberate schema migration, not a side effect of a dep run.

### Breaking changes that do NOT apply to us

Checked each item in the [official 2.0 migration guide](https://www.sea-ql.org/blog/2026-01-12-sea-orm-2.0/):

- **No `Alias::new`, no `IntoCondition`, no `impl Iden`** anywhere in the codebase.
- **`execute` → `execute_raw` rename**: our raw SQL goes through `find_by_statement`, which is unaffected. `execute_unprepared` (used in the migration crate) is unchanged in 2.0.
- **`insert_many` overhaul**: all four call sites already guard with `is_empty()`, so the empty-iterator change is a no-op, and none of them use `exec_with_returning`.
- **`sqlx` 0.9**: pulled *only* by sea-orm — no direct usage in our code, so the major bump touches nothing of ours.
- **Edition 2024 / MSRV 1.85**: we're on nightly 1.94, edition 2024 already.

## Also: in-range lockfile refresh

77 packages advanced — tokio 1.52.3→1.53.0, sentry 0.48.4→0.48.5, serde 1.0.229, uuid 1.24.0, rustls 0.23.42, hyper 1.11.0, futures 0.3.33.

- **lodestone held at `192faccd`** as every prior run has done — a plain `cargo update` advances it to a reqwest-0.12 commit that breaks the 0.11 client in `ultros`.
- **wasm-bindgen did not move** (0.2.126), so the `Dockerfile` CLI pin needs no bump this run.
- `npm audit` in `integration/` → **0 vulnerabilities**.

## Other ceilings — all re-verified, all unchanged

| Dep | Ours | Latest | Status |
|---|---|---|---|
| leptos | 0.8.20 | 0.9.0-**beta** | Blocked: `leptos_i18n 0.6.2` pins `leptos ^0.8` |
| leptos-use | 0.18.3 | 0.19.0 | Duplicate trap: `leptos_i18n` pins `^0.18` |
| axum | 0.8.9 | 0.8.9 | At latest |
| tower-http | 0.6.10 | 0.7.0 | Duplicate trap: `leptos_axum` pins `^0.6` |
| itertools | 0.14.0 | 0.15.0 | Duplicate trap: tantivy pins `^0.14` |
| reqwest 0.13 | — | 0.13.4 | Pulled only by sentry; irreducible |

leptos 0.9 moved alpha → **beta**, but no ecosystem crate supports it yet, so the frontend ceiling is unchanged.

## Dependabot

4 open alerts (1 high / 1 medium / 2 low), **all rustls-webpki** — unchanged by this PR and not fixable by a lockfile bump. They trace to residual webpki `0.101.7` (rustls 0.21) and `0.102.8` (rustls 0.22), both pinned by serenity 0.12.5 / poise 0.6.2. The first-party rustls 0.23.42 path is already on the patched `0.103.13`. **This PR does not fix them.**

## Verification (local, Windows)

- `cargo clippy --all-targets -- -D warnings` → **exit 0, zero warnings**
- `cargo check -p ultros -p ultros-app` → exit 0
- `cargo test -p ultros-db -p ultros-api-types` → **150 passed, 0 failed** (CI does not run tests, and this is a DB-layer change)

**Not covered locally:** the `jemalloc`/`profiling` features can't build on Windows MSVC — `jemalloc` is exercised by the Docker PR build. Nothing in this PR touches that path. No runtime database was exercised, so the sea-orm 2.0 query-builder changes are verified by compile + unit tests only; worth a look at staging before deploy.


akarras:
## Reviewed and verified locally — merging

Compile-checking a major ORM bump isn't enough, so I ran the app against the **real local Postgres** (~3M `active_listing`, ~7.2M `sale_history`) with `sea-orm 2.0.0` + `sqlx 0.9`.

### Runtime verification (the part CI can't do)
`Migrator::up` runs at startup (`ultros-db/src/lib.rs:67`) — it completed with **no migration error and no `DbErr`**. The full ingestion pipeline then ran live:

- `updated listings` / `Removed listings` — the `.and()` filter path in `listings.rs`
- `Stored sale data, added_sales: [...]` — `insert_many` in `sales.rs`
- a window-function query returned **1,300,357 rows in 14.8 s** via the raw-SQL `find_by_statement` path

`/item/Cerberus/5057` rendered fully (200, 172 KB, 1.17 s) with real data — "Iron Ingot – Cerberus", real listings and sale rows. Read and write paths both exercised.

### Verified your "does not apply" claims independently
- `Alias::new` / `IntoCondition` / `impl Iden` — **zero** hits, confirmed.
- `.execute(` — only hit is `alerts/delivery.rs:290`, an HTTP client, not sea-orm. Confirmed.
- **`ExprTrait` is provably SQL-identical.** In 0.32.7 the inherent `Expr::add` already delegated to `ExprTrait::add`; 1.0 just removed the forwarder. Operators map to the same `BinOper` in both versions (`and`→`And`, `lt`→`SmallerThan`, `add`→`Add`). So the import is pure API surface with zero SQL change — which also covers the auth-gated `lists.rs` invite-redemption path I couldn't reach through the UI.
- **`postgres-use-serial-pk` is right, and I checked it against the live schema:** `list.id` is `nextval('list_id_seq'::regclass)`, `is_identity = NO`. Genuine `serial`, so without the flag fresh DBs would diverge from prod. Good catch.

### One correction to the PR body
> all four call sites already guard with `is_empty()`

There are **six** `insert_many` sites, and `lists.rs:573` is **not** guarded — `insert_queue` is empty whenever every added item already exists in the list (an ordinary duplicate-add, which lands in `updated_models` instead).

This lands in your favour. In 1.1.20 `Insert::exec_without_returning` had no empty guard and went straight to `exec_insert_without_returning`, producing a malformed `INSERT ... VALUES`. In 2.0 `InsertMany::exec_without_returning` short-circuits with `if self.empty { return Ok(0) }` before the asserting `into_one()`. So 2.0 doesn't just avoid a regression here — it **fixes a latent duplicate-add bug**. Conclusion unchanged, reasoning worth correcting.

### Other checks
lodestone held at `192faccd` (confirmed in the build), wasm-bindgen still 0.2.126 so no Dockerfile pin bump, single major of sea-orm/sea-query/sqlx in the lock.

The 4 rustls-webpki Dependabot alerts are correctly described as out of scope.

akarras:
## Review: verified against a real database — merging

Reviewed the 11 code lines, the lockfile, and exercised sea-orm 2.0 at runtime against the local prod-shaped Postgres (3.36M `active_listing`, 10.5M `sale_history`).

**Lockfile** — confirmed single major of each, no duplicate trap:
`sea-orm 2.0.0`, `sea-orm-migration 2.0.0`, `sea-query 1.0.1`, `sea-schema 0.18.1`, `sqlx 0.9.0`, `sqlx-postgres 0.9.0`. lodestone still pinned at `192faccd` on reqwest 0.11.27 as claimed.

**`postgres-use-serial-pk` is the right call, and I confirmed why empirically.** The live table really is serial-backed:
```
\d active_listing
id | integer | not null | nextval('active_listing_id_seq'::regclass)
```
Without the flag a fresh migrate would emit `GENERATED BY DEFAULT AS IDENTITY` and diverge from that. Good catch keeping this a pure dep bump.

**Runtime verification (the gap the PR flagged as not covered):**
- Reads: `/item/Cerberus/5057` server-rendered 172 KB of real listing + sale-history rows.
- Writes: with the full ingestion firehose running, `max(timestamp)` on `active_listing` tracked live to the current minute — so `insert_many` and the now-`Option` `last_insert_id` path are both exercised, not just compiled.
- `cargo test -p ultros-db -p ultros-api-types` → **150 passed, 0 failed** (131 + 19), matching your numbers.
- `cargo fmt --all -- --check` and `cargo clippy -p ultros --all-targets -- -D warnings` → clean.

No conflicts with #961/#963/#964 (verified by merging all four locally — zero conflicts, disjoint file sets).

Squash-merging.

PR #965, opened 2026-07-21
🛡️ Sentry: [test coverage improvement]

akarras:
💡 What: Extracted the Tailwind CSS class mapping for `ConfidenceTone` in `ultros-frontend/ultros-app/src/components/confidence_badge.rs` into a pure testable `css_classes(&self)` function. Added a unit test to verify the output for each variant.
🎯 Why: The reliability gap it covers is that embedding conditional styling directly in Leptos macros causes testability issues and creates dummy reactive effects for static enum values. Extracting the mapping allows for deterministic unit testing without Leptos reactive context, as per `sentry.md` guidelines.
📊 Impact: `ultros-frontend/ultros-app/src/components/confidence_badge.rs` (`ConfidenceTone::css_classes` and `ConfidenceBadge`) is now completely covered for its styling mapping logic.
🔬 Verification: Run `cargo test -p ultros-app components::confidence_badge` to verify the new test.

---
*PR created automatically by Jules for task [12721976015404446407](https://jules.google.com/task/12721976015404446407) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/components/confidence_badge.rs
+++ b/ultros-frontend/ultros-app/src/components/confidence_badge.rs
@@ -18,6 +18,25 @@ pub enum ConfidenceTone {
     Error,
 }
 
+impl ConfidenceTone {
+    pub fn css_classes(&self) -> &'static str {
+        match self {
+            Self::Success => {
+                "text-emerald-300 border-emerald-400/40 bg-[color:color-mix(in_srgb,#10b981_14%,transparent)]"
+            }
+            Self::Neutral => {
+                "text-[color:var(--color-text)] border-[color:var(--color-outline)] bg-[color:color-mix(in_srgb,var(--brand-ring)_10%,transparent)]"
+            }
+            Self::Warning => {
+                "text-amber-300 border-amber-400/40 bg-[color:color-mix(in_srgb,#f59e0b_12%,transparent)]"
+            }
+            Self::Error => {
+                "text-red-300 border-red-400/40 bg-[color:color-mix(in_srgb,#ef4444_12%,transparent)]"
+            }
+        }
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ConfidenceLabel {
     High,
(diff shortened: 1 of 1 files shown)


akarras:
## Review: right direction, low risk — merging

This is the shape I like: it **removes** 12 `class=(..)` toggles that were closing over a *constant*. `band: ConfidenceBand` is a plain value (not a signal), `tone` is derived once from it, so every one of those `move || tone == ConfidenceTone::X` closures was a reactive effect that could never fire. Hoisting to a single `format!` is strictly less work per badge and makes the class attribute deterministic — which slightly *shrinks* the SSR/CSR mismatch surface rather than growing it.

Semantics are unchanged: leptos class toggles only ever *add* on `true`, so old output was `base + matching tone classes` and new output is the same set via `format!`. Class order within the attribute doesn't affect specificity.

**Note on the diff:** `git diff origin/main pr-965` looks alarming (13 files, −1157) but that's only because main moved under it while #961–#964 merged. Against its actual merge base (`080cb9cf`) it is one file, +42/−13. Re-merged it onto current main locally — clean, still one file.

### Verified

- `cargo test -p ultros-app --lib` → **202 passed, 0 failed** (CI doesn't run tests, so this needed a local run).
- `cargo clippy -p ultros-app --all-targets -- -D warnings` and `cargo fmt --all -- --check` → clean.
- **The real risk of this refactor is Tailwind purging class literals that moved out of the `view!` macro.** It doesn't: built the app and grepped the served `ultros.css` (199 KB, byte-identical to the fresh build output) for all four tone groups — `text-emerald-300`, `border-emerald-400\/40`, `#10b981`, `text-amber-300`, `border-amber-400\/40`, `#f59e0b`, `text-red-300`, `border-red-400\/40`, `#ef4444`, plus the neutral `--brand-ring`/`--color-outline` vars — **all present**. The JIT still sees them in `css_classes()`.
- App builds and serves; other pages hydrate clean.

### What I could not verify, honestly

I could not get this badge to render in the browser. It only draws when `get_confidence_verdict_display(band)` returns `Some`, and local ClickHouse is weeks stale so every band comes back `Unknown` → the component returns `().into_any()`. The Trends page *does* have `medium`/`low` bands in its data (471/29 of 500), but it can't render them because of a **pre-existing, unrelated bug**: `get_trends_v2` in `api.rs` sends `show_suspicious=1`, and the handler wants a bool, so every v2 trends fetch 400s and the page shows "0 results — Error loading trends: … provided string was not `true` or `false`". Filed separately; nothing to do with this PR.

Given that, I'm leaning on the static argument plus the CSS check rather than a screenshot. For a pure class-string extraction with identical literals on both sides, that's proportionate.

### Minor

The new test is a change-detector — it asserts `css_classes()` returns the exact literals `css_classes()` contains, so it can only catch an unintended edit, not a wrong mapping. Harmless, and it does pin the tone→class contract. Not worth another round trip.

Also worth knowing: `components/tool_help.rs` defines a *second, unrelated* `ConfidenceBadge` (`total_sales`/`daily_sales`) that renders the "Verify In-Game" / "Slow mover" chips and still uses the old toggle style. Same name, different component — easy to confuse when grepping. A follow-up could give it the same treatment or rename it.

Squash-merging.

PR #973, opened 2026-07-24
feat(item): rework the item view market panel

akarras:
Six related tweaks to the item view's market panel, from a design pass on the "Cheapest Found" / "Sale History" area.

## What changed

**1. The freshness chip explains itself.** It read `Caution (4h 28m)` — the verdict word alone told you nothing. It now reads `Data 4h 28m old`, with a `title` tooltip carrying the verdict *and* what to do about it ("Getting old for this item's sales rate — prices may have changed."). Tone colors are unchanged, so the at-a-glance signal survives.

**2. No more two-column split.** The `xl:grid-cols-[0.85fr_1.45fr]` layout left a tall empty left column whenever the stats panel was shorter than the chart. `ListingsContent` now stacks: stats strip → market share → chart, all full width. `id="history"` moved onto the chart wrapper so the Real Price cell still deep-links correctly.

**3. Flat stat strip instead of four cards.** The NQ / HQ / Real Price / Active Listings cells lose their `rounded-lg border … min-h-24` boxes and become a separated row — `grid-cols-2` with hairline dividers on mobile, `lg:grid-cols-4` with left dividers on desktop. The Real Price cell's basis / recent average / median lines collapse into one muted `text-[10px]` line. Header is a single baseline-aligned flex row; the decorative coin icon is gone.

**4. Extended sale history auto-loads, button removed.** An `Effect` (which never runs during SSR) fires `spawn_local(get_extended_sale_history(id, world, 10_000))` after hydration and overrides the chart's sales signal — the same swap the button used to perform, so SSR and the hydrating client still agree on the 200-sale view and the render path is unchanged. The server cap goes `5_000 → 10_000`.

Because the fetch now auto-fires on navigation, results are checked against the current item/world before being applied, and all writes use `try_set` (the component can be disposed mid-flight — cf. #6864/#6867):

```rust
let still_current = item_id.try_get_untracked() == Some(id)
    && world.try_with_untracked(|w| w == &world_name) == Some(true);
if !still_current { return; }
```

A small `Loading full history…` note sits next to the "based on N sales" line while it's in flight.

**5. Market share by world (new).** When the selected scope is a datacenter or region, a panel of horizontal bars shows each world's share of the currently listed quantity, with the listing count alongside. It reads `filtered_listings`, so DC exclusion toggles and `?exclude-worlds=` apply for free. On a single world it renders nothing. It subscribes to `listing_resource` inside its own `Transition` before touching `filtered_listings`, matching `HighQualityTable` — the SSR/CSR structural contract behind #6831.

**6. Selected world actually looks selected.** The active region/DC/world in the selector bar was a faint border tint. It's now a filled brand pill (`bg-[color:var(--brand-bg)] / text-[color:var(--brand-fg)]`) with `aria-current="page"`. The per-type color and hover classes are applied only when *not* selected so they can't fight the pill on specificity.

## i18n

Added to all seven locales with real translations: `freshness_data_age`, `freshness_tooltip_{fresh,caution,verify,no_data}`, `extended_history_loading`, `market_share_{title,subtitle,row}`. Removed: `freshness_label_with_age`, `load_extended_history`, `load_extended_history_help`, `extended_history_loaded`.

## Validation

Done:
- `./check_ci.sh` — `cargo fmt --all -- --check` + `cargo clippy --all-targets -D warnings`, clean.
- `cargo test -p ultros-app freshness` — 3 passed.
- Full `cargo leptos serve` build (server binary + WASM + Tailwind) completes.

**Not done — please exercise this manually before merging.** I could not get a running instance up locally: the machine ran out of disk mid-build and the Docker postgres container would not stay up, so the app never bound to a port. The unverified behaviors are the runtime ones:

- Cold full-page loads of a world / datacenter / region item URL with no tachys hydration panic in the console. Items 4 and 5 are the ones that touch the SSR/CSR boundary, so this is the check that matters most.
- Exactly one client-side `GET /api/v1/extended_history/…?limit=10000` after hydration (and none during SSR), with the sales count climbing above 200.
- Market-share bars appearing only on DC/region pages and reacting to the exclusion toggles.
- Visual pass on the stats strip at ~375px (2×2 + hairlines) and ≥1024px (1×4).


akarras:
## Reviewed, fixed one real bug, verified the rest in a browser

I worked the "not done — please exercise this manually" list against a local build of this branch merged with #973/#974/#975/#976 on current `main`.

### Bug found and fixed (pushed as 11abd2d)

**Item 6 did not work.** The selected world/DC/region rendered *identically* to an unselected one. `style/tailwind.css` has a global

```css
a:not(.nav-link):not(.btn):not(.btn-primary):not(.btn-secondary):not(.btn-ghost) {
    @apply hover:underline rounded-md;
    background-color: transparent;
}
```

At specificity **(0,5,1)** that beats a plain **(0,1,0)** utility, so both `bg-[color:var(--brand-bg)]` and `rounded-full` were overridden. Measured on the real element:

```
selected "Materia"  →  bg rgba(0,0,0,0)   radius 6px
unselected "Ravana" →  bg rgba(0,0,0,0)   radius 6px
```

Isolation test confirmed the mechanism — the *same* class list gives `rgb(76,29,149)` on a `<div>`/`<span>` but transparent on an `<a>`. The PR body's reasoning ("colors applied only when not selected so they can't fight the pill on specificity") correctly handled the *intra*-class-list conflict but not this global anchor rule.

Fix marks the three conflicting utilities `!important` — the idiom already used across the analyzer / leve / recipe / FC-crafting tabs (`active_classes="!text-[color:var(--brand-fg)]"`). Now:

```
selected   → bg rgb(76,29,149) (= --brand-bg)   radius 2.2e7px (full)   aria-current="page"
unselected → unchanged
```

### Your checklist

- [x] **Cold loads of world / DC / region URLs, no hydration panic.** `/item/Materia/46185` (DC) and `/item/Zurvan/46185` (world) both hydrate cleanly — `boot-progress` absent, no `tachys` / `RefCell` / `unreachable` in console. Items 4 and 5 are the SSR/CSR-boundary ones and both are clean.
- [x] **Exactly one client-side `extended_history` call, none during SSR.** Network shows a single `GET /api/v1/extended_history/Materia/46185?limit=10000 [200]`. Sales count climbs **200 → 318**, which matches the row count the endpoint actually returns for that scope — so the swap-in demonstrably happened.
- [x] **Market share only on DC/region, reacts to filters.** Renders on `/item/Materia/…` with 5 world bars sorted by quantity desc; **absent** on `/item/Zurvan/…`. Reads `filtered_listings`, so exclusions apply.
- [x] **Stats strip at both widths.** Mobile: `grid-template-columns: 222px 222px` (2×2) with hairlines on `nth-child(even)` (`border-left: 0.667px`) and `nth-child(n+3)` (`border-top`), no horizontal overflow. Desktop ≥1024: one row of 4 with left dividers.
- [x] Freshness chip reads `Data 1d 4h old` with tooltip `Caution: Getting old for this item's sales rate — prices may have changed.`

### Payload cost of auto-loading

Worst case measured — region scope on a hot item (`North-America/46185`) hits the 10k cap: **969KB raw, 76KB gzip, 0.21s** server-side. Responses go out **brotli**-compressed, so real wire cost is lower. Typical world/DC scopes are 8–150KB raw / 1–9KB compressed. Acceptable for an unconditional fetch, but worth knowing the region-scope tail exists now that it's not behind a button.

`cargo test -p ultros-app --lib` → **210 passed, 0 failed** (run locally; `cargo test` is commented out in `rust.yml`), including the 3 `freshness::tests`. `cargo fmt --all --check` and `cargo clippy -p ultros-app --all-targets -D warnings` both clean with the fix applied.

Unrelated pre-existing note: `/api/v1/item_stats/{dc}/{item}` returns **400** on every DC/region item page — by design (`item_stats.rs:37` rejects non-world selectors), but the client fires it unconditionally, so each DC page view logs a console error. Predates this PR; worth a separate cleanup.

PR #974, opened 2026-07-24
feat(item-explorer): world picker, row-based results, grouped subcategory popovers

akarras:
Reworks `/items/*` around three things: an on-page world picker, a row-based results list, and grouped subcategory dropdowns.

## World picker

Previously the explorer silently scoped prices to the global `PRICE_ZONE` cookie — settable only in Settings — with no on-page control.

`?world=` now scopes prices per page. `use_explorer_price_scope` resolves `?world=` → `PRICE_ZONE` cookie → `"North-America"`, validating the param against `WorldHelper` so typos and stale links fall through to the default rather than producing a scope the API can't serve. **The picker only ever writes the query param** — the cookie is read-only here, so scoping one browse session doesn't change prices everywhere else in the app.

`ItemExplorer` shadows the `CheapestPrices` context with a resource keyed on that scope, so `ItemList`, `CheapestPrice`, `JobSetCard` and `JobSetDetail` pick it up through `use_context` with no signature changes. When the resolved scope equals the cookie zone (the common no-`?world=` case) the scoped resource awaits the root one instead of issuing a duplicate request.

## Row-based results

The card grid becomes a `role="table"` row list so prices line up in a scannable column, plus:

- a cheapest-listing **World** column, hidden when the scope is already a single world
- sortable column headers (name / iLvl / price) that toggle `dir` and reset `page`, built from `use_location()` so `world`, `per_page` and friends survive
- a rows-per-page selector (25/50/100), clamped so a hand-edited `?per_page=` can't produce a surprising page size

Three responsive tiers rather than one breakpoint. The full column set can't appear before `xl`: next to the 268px app sidebar the fixed columns overflow the container and `minmax(0,1fr)` collapses the item name to zero width — item names vanish entirely. Below `lg` the extra columns fold into a compact line under the name.

## Grouped subcategory popovers

Weapons and Job Sets render a `GroupedNavPopover` with sections by role (Tank / Healer / Melee / Physical Ranged / Magical Ranged / DoH / DoL). Armor, Items and Housing keep the inline chips at ≤8 subcategories and otherwise get the same popover with one headerless section.

The game data we ship has no per-job role field (`ClassJob` carries only indices and priorities), so the grouping is a hand-written abbreviation table. Weapon categories resolve through their `class_job`, so one table serves both tabs and categories added by future expansions group automatically as long as their job's abbreviation is known. Tests assert every visible job and every weapon category maps to a real role, so a new job fails CI rather than silently landing in "Other" — this already caught Beastmaster.

## Reviewer notes

- **Hydration** was the main risk. Existing `hydrated` gates are preserved, conditionals stay inside always-present wrappers, and the popover panel is always in the DOM (toggled by class) so the SSR and first-CSR view trees keep the same shape.
- **i18n**: 13 new keys with real translations across all seven locales. Category and job names come pre-localized from xiv data, so the toolbar's render closure calls `tracked_data()` to re-render on locale swap.
- `/items` index now renders the already-localized `DefaultItems` instead of a hardcoded English string.
- `CheapestPrice` gains an optional `show_world` prop for layouts that show the world in a separate column. Default is unchanged.
- **Unrelated drive-by**: `scripts/run_e2e.sh` aborted immediately on macOS's stock bash 3.2, where `"${arr[@]}"` on an empty array trips `set -u`. Fixed with the `${arr[@]+"${arr[@]}"}` guard so the harness runs locally at all.

## Verification

- `cargo test -p ultros-app` — 208 passed, 0 failed
- `./check_ci.sh` (fmt + clippy `-D warnings`) — exit 0
- Manual click-through against a local server: popover renders all 33 weapon categories in role order, the trigger label switches to the active subcategory, sort-header hrefs preserve query params, rows render correctly at 1440px. **This is what surfaced the collapsed-name-column bug at mid widths**, which is fixed here.
- E2E: the full-suite run at the harness default `CONCURRENCY=16` exits 1 with navigation timeouts on nearly every route — including `/privacy` and `/cookie-policy`, which contain no explorer code. SSR responds in 0.15–0.36s per route when the server is idle, so this is 16 headless Chrome instances starving the CPU rather than slow rendering. The FC-crafting E2E passed in the same run. A low-concurrency re-run was still in flight when this PR was opened; **I'll confirm it here rather than leaving it implied.**


akarras:
## Reviewed — verified in a real browser, approved to merge

Built merged with #973/#975/#976 on current `main`, exercised at 1440px. **Hydration clean on every explorer route** (`boot-progress` absent; no tachys / RefCell / unreachable errors — only the pre-existing dev-build `leptos_i18n` tracking warnings that `main` also emits).

**Popover**: `aria-expanded` false→true, panel `hidden` → visible (always in DOM, class-toggled, so SSR and first CSR agree). **33 weapon categories** in **7 role sections in order** (Tank, Healer, Melee, Phys Ranged, Magical Ranged, DoH, DoL), grouped correctly.

**World picker never writes the cookie** — selecting Ravana: url gained `?world=Ravana`, `PRICE_ZONE` stayed `Cerberus`.

**`?world=` validation**: `?world=NotARealWorld` falls back to the cookie zone (links render `/item/Cerberus/…`), 51 rows, no error state.

**World column gating** verified both directions by computed style: scope `Cerberus` → `display:none`; `?world=Materia` → `display:block` with cells Ravana/Sophia/Bismarck/Zurvan.

**`cargo test -p ultros-app --lib` → 210 passed, 0 failed**, run locally because `cargo test` is commented out in `rust.yml` (green CI only proves the tests compile). Includes `every_visible_job_maps_to_a_role`, `every_weapon_category_maps_to_a_role`, and `test_job_filtering` (the one that broke on the Beastmaster bump — passes).

The completeness assertions are the right call: they fail CI on a game-data bump instead of silently dumping new jobs into "Other".

i18n: all 7 locales at **1511 keys, exact parity**, valid JSON, no duplicate keys after merging with #973. Translations are real — the only two matching English (`Lv` ja, `Tank` fr) are correct FFXIV terms.

`${arr[@]+"${arr[@]}"}` is the correct bash 3.2 + `set -u` guard, and `show_world` defaults true so existing `CheapestPrice` callers are unaffected.

PR #975, opened 2026-07-24
feat(universalis): identify ourselves + close catch-up gaps

akarras:
## Why

Two things came out of an audit of how we talk to Universalis:

1. **We weren't identifying ourselves properly.** The REST client sent a bare `"ultros"` User-Agent, and the websocket sent none at all — our firehose connection was anonymous. Universalis asks scripted consumers to identify with a version and a contact point.
2. **The catch-up path had a correctness hole.** `most-recently-updated` was diffed against our recency table *by item id only*. If an item appeared anywhere in our window, it looked in-sync — even when Universalis had a newer upload for it. Those missed events were silently dropped and never recovered.

## What changed

**User-Agent**

- `UNIVERSALIS_USER_AGENT` in `main.rs`: `ultros/<CARGO_PKG_VERSION> (+https://ultros.app)`.
- `WebsocketClient::connect` now takes a user-agent and sets it on the handshake via `IntoClientRequest`, threaded through both the initial connect and the reconnect loop. An invalid header value warns and connects without rather than failing the connection.
- Breaking change to the `universalis` crate: `WebsocketClient::connect()` gained a parameter. Example and doctest updated.

**Shared client**

- `UniversalisClient` derives `Clone` (it wraps a `reqwest::Client`, so clones share the pool) and is constructed once in `main`, then shared through `WebState` with a `FromRef` impl. `refresh_world_item_listings` was building a brand-new client — and connection pool — on every request.

**Catch-up correctness**

- `missed_updates` is now a pure, unit-tested function that compares timestamps in addition to item ids. An item counts as missed when their upload is newer than our ingest by more than `UPLOAD_TIME_SLACK_SECONDS` (120s). The slack exists because `listing_last_updated` stores *ingest* time, not Universalis' upload time — without it, normal ingest latency would look like a miss on every pass.
- **Window saturation.** The endpoint returns at most 200 entries. If all 200 come back as missed, updates have almost certainly scrolled past where we can see them. We now emit `ultros_catchup_window_saturated` and escalate to a full marketable-item sweep of that world, rate-limited to once per world per 6h so a persistent problem can't turn into a request storm.
- `ultros_catchup_items_recovered` (labelled by world) makes the size of the websocket gap visible in metrics instead of invisible.

**Error visibility**

- `check_items` swallowed DB failures in `if let Ok(..)` on both the listing and sale paths. Both now `tracing::error!` with item/world context. The `sentry_tracing` layer already installed in `main` turns ERROR-level events into GlitchTip issues, so these surface without any extra plumbing.

## Reviewer notes

- The 120s slack and the 6h full-sweep cooldown are judgement calls — both are single named constants at the top of `item_update_service.rs` if you want different numbers.
- Six unit tests cover `missed_updates`: unknown item, in-sync item, newer-upload-than-ingest, newer-but-within-slack, our-side-only, and unsorted inputs.
- `./check_ci.sh` passes (fmt + `clippy --all-targets -D warnings`, zero warnings). Clippy needed the `xiv-gen/ffxiv-datamining` and icon submodules, which wouldn't fetch in this sandbox — I borrowed the checkouts from the main clone via symlink for the build only. Nothing submodule-related is in this commit; `git diff` is the seven source files and nothing else.

### DIFF
--- a/ultros/src/item_update_service.rs
+++ b/ultros/src/item_update_service.rs
@@ -1,19 +1,32 @@
-use std::{sync::Arc, time::Duration};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 use futures::{StreamExt, stream};
 use tokio::time::Instant;
 use tokio_util::sync::CancellationToken;
-use tracing::{error, info, instrument};
+use tracing::{error, info, instrument, warn};
 use ultros_api_types::websocket::{ListingEventData, SaleEventData};
 use ultros_db::{
     UltrosDb,
-    common::partial_diff_iterator::PartialDiffIterator,
+    common::partial_diff_iterator::{DiffItem, PartialDiffIterator},
     entity::{listing_last_updated::Model, world},
     world_data::world_cache::WorldCache,
 };
 use universalis::{UniversalisClient, WorldId, WorldItemRecencyView};
 
 use crate::event::{EventProducer, EventType};
+
+/// Universalis' most-recently-updated endpoint returns at most this many entries.
+const RECENTLY_UPDATED_WINDOW: u8 = 200;
+/// Our `listing_last_updated` rows store *ingest* time rather than Universalis'
+/// upload time, so an entry only counts as missed when their upload is newer
+/// than our ingest by more than this slack.
+const UPLOAD_TIME_SLACK_SECONDS: i64 = 120;
+/// Minimum time between saturation-triggered full sweeps of a single world.
+const FULL_SWEEP_COOLDOWN: Duration = Duration::from_secs(6 * 60 * 60);
 
 /// Item update service attempts to keep ultros' data in sync with Universalis' data.
 /// It does this primarily by comparing the recently updated items on Universalis with recently updated items on ultros
@@ -23,6 +36,8 @@ pub(crate) struct UpdateService {
     pub(crate) universalis: UniversalisClient,
     pub(crate) listings: EventProducer<ListingEventData>,
     pub(crate) sales: EventProducer<SaleEventData>,
+    /// Per-world timestamp of the last saturation-triggered full sweep.
+    pub(crate) full_sweep_cooldowns: Mutex<HashMap<i32, Instant>>,
 }
 
 struct CmpListing(Model);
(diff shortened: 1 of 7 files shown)


akarras:
## Reviewed — verified, approved to merge

Checked the two things that could have made this wrong, rather than taking them on trust:

**1. Is `date_time` really UTC?** The new comparison does `ours.0.date_time.and_utc()`, which is only correct if the column stores UTC. It does — `ultros-db/src/recently_updated.rs:19` writes `Utc::now().naive_utc()`. If it had been local-naive, `and_utc()` would have skewed every row by the TZ offset and made *everything* look missed.

**2. Does `listing_last_updated` actually track websocket ingest?** If that row only advanced on the catch-up path it would lag permanently, `theirs.upload > ours.ingest + 120s` would be true constantly, and saturation would fire every pass on every world. It doesn't — `set_last_updated` is called from `update_listings` (`listings.rs:428`) and `update_sales` (`sales.rs:40`), both of which the websocket path uses. So the 120s slack is measured against real ingest time and saturation stays rare.

**Full-sweep blast radius** is bounded, which was my main worry given the trigger correlates across worlds (a websocket outage hits all of them at once). `start_service` iterates worlds **sequentially** and `check_items` awaits its 100-item chunks in order, so at most one full sweep runs at a time — plus the 6h per-world cooldown. No thundering herd.

**Tests: 6/6 pass.** The `ultros` bin test target doesn't build on Windows/MSVC (`rustc-LLVM ERROR: out of memory` this run), so I extracted `missed_updates` into a standalone crate with `partial_diff_iterator.rs` copied verbatim and field-identical `Model` / `WorldItemRecencyView` stand-ins, then ran your tests unmodified:

```
test tests::item_unknown_to_us_is_missed ... ok
test tests::item_we_ingested_after_their_upload_is_in_sync ... ok
test tests::item_with_newer_upload_than_our_ingest_is_missed ... ok
test tests::upload_newer_within_slack_is_in_sync ... ok
test tests::item_only_on_our_side_is_ignored ... ok
test tests::unsorted_inputs_still_diff_correctly ... ok
test result: ok. 12 passed; 0 failed
```
(12 = your 6 + the 6 pre-existing `partial_diff_iterator` tests, which also pass against the copied file.)

Also confirmed `UniversalisClient::new` genuinely applies the UA (`Client::builder().user_agent(...)`, `lib.rs:325`) rather than just storing it, and that `recently_updated_items` takes `u8` so the new `RECENTLY_UPDATED_WINDOW` const is type-compatible.

One naming nit, non-blocking: `ultros_catchup_items_recovered` increments *before* `check_items`, so it counts items **detected as missed**, not recovered. Fine as a gap-size signal — just don't read it as a success count.

PR #978, opened 2026-07-26
feat(analyzer): default the sale-velocity filter to 1d

akarras:
## Why

Open `/flip-finder` cold today and you get every item that clears the profit threshold — including ones that sell once a month. Those rows look like the best flips on the page (high profit, high ROI) and they're exactly what a newcomer buys and then sits on for three weeks.

The filter that fixes this already existed. It was just off by default and buried behind **More Filters**, so the people who most needed it never found it.

## What changed

**Seeded defaults.** A bare visit now lands on a velocity-filtered view:

| Page | Param seeded | Meaning |
|---|---|---|
| `/flip-finder` | `next-sale=1d` | avg sale duration < 1 day |
| `/vendor-resale` | `next-sale=1d` | avg sale duration < 1 day |
| `/recipe-analyzer` | `min-sales=1` | ≥ 1 sale/day |
| `/fc-crafting-analyzer` | `min-sales=1` | ≥ 1 sale/day |

The default lives in the **URL**, not in the filter logic, so chips, Clear All, the input box, and shared links all behave exactly as they do for a hand-typed filter. Seeding fires only when the param is *absent*, so a link carrying `?next-sale=` or `?min-sales=0` still means "no limit" — and both forms are what the input box produces when you empty it.

New `ultros-app/src/query_defaults.rs` holds the two helpers and the constants.

**The control is now visible.** Max Sale Time moves into the primary toolbar on both analyzer and vendor resale — a default you can't see is just a mystery. On vendor resale that emptied the secondary toolbar entirely, so the drawer, its button, and `show_more` are deleted.

## Reviewer notes

**Seeding deliberately lives in the route components, not next to the `query_signal` declarations.** All four filter toolbars render inside a `Suspense` closure that re-runs whenever its resources change — including live market refetches on the analyzer pages. Seeding there would remount and re-seed on every refresh, silently reinstating a filter the user had just cleared. `AnalyzerWorldView` / `VendorWorldView` / `RecipeAnalyzer` / `FCCraftingAnalyzer` mount once per navigation, which is the granularity a default wants.

**Accepted behavior change:** the next-sale filter drops items whose `avg_sale_duration` is unknown (`unwrap_or(false)`, unchanged). Making it a default therefore hides no-data items by default, and on vendor resale — where `sale_summary` is an `Option` — that may shorten the list noticeably. This is intended: an item with no sales history isn't a good recommendation for the audience this default serves. The chip renders with a working X, so it's one click to undo. **Worth a sanity check on a real world that the vendor resale list doesn't get too thin.**

**Two fixes on the lines being touched:**
- Filter params now navigate with `replace: true, scroll: false`. Previously every keystroke in a duration/min-sales box pushed a history entry *and* scrolled the page to top.
- Vendor resale's hardcoded `placeholder="e.g. 7d 12h"` moves to the existing `analyzer_placeholder_7d_12h` key. No new locale entries needed — verified the key exists in all seven files.

**Out of scope:** `"More Filters"` / `"Fewer Filters"` remain hardcoded English in `analyzer.rs` (vendor resale's copy disappeared with the button). That needs 2 keys × 7 locales and is unrelated; filed separately.

## Testing

- `./check_ci.sh` — passes (fmt + clippy clean, whole workspace).
- `cargo test -p ultros-app --lib query_defaults` — passes. The test pins `DEFAULT_MAX_SALE_TIME` to parse as exactly one day; a typo there wouldn't error, it would silently turn the default into no filter at all.
- **Not yet exercised in a browser.** The URL-seeding and back-button behavior are worth a manual pass: bare `/flip-finder` should land on `?next-sale=1d` with the chip showing, the X should stay cleared across a live market refresh, and Back should leave the page rather than stepping to the unfiltered URL.

Design doc: `docs/superpowers/specs/2026-07-25-analyzer-velocity-default-design.md`

### DIFF
--- a/docs/superpowers/specs/2026-07-25-analyzer-velocity-default-design.md
+++ b/docs/superpowers/specs/2026-07-25-analyzer-velocity-default-design.md
@@ -36,9 +36,22 @@ scope. No filter logic changes; no new locale keys.
 ### Seeding the default
 
 New module `ultros-frontend/ultros-app/src/query_defaults.rs`, a sibling of
-`math.rs` and `freshness.rs`, exposing one helper. On mount, if the query
-parameter is absent from the URL, the helper writes the default value. Each of
-the four routes calls it once, next to its existing `query_signal` declaration.
+`math.rs` and `freshness.rs`, exposing two helpers:
+
+- `filter_query_signal(key)` — a `query_signal` carrying the nav options below.
+  Used wherever a filter is read or written.
+- `seed_query_default(key, default)` — writes `default` into the URL if `key` is
+  absent when it mounts.
+
+**Seeding belongs to the route component, not the filter's own component.** The
+filter toolbars live in `AnalyzerTable`, `VendorResaleTable`,
+`RecipeAnalyzerTable`, and `FCCraftingAnalyzerTable`, all of which are rendered
+inside a `Suspense` closure that re-runs whenever its resources change — a world
+switch, or a live market refetch on the analyzer pages. Seeding there would
+remount and re-seed on every refresh, silently reinstating a filter the user had
+just cleared. `seed_query_default` is therefore called from `AnalyzerWorldView`,
+`VendorWorldView`, `RecipeAnalyzer`, and `FCCraftingAnalyzer`, which mount once
+per navigation — the granularity a default wants.
 
 Everything downstream is untouched. The input box, the `Next Sale ≤ 1d` chip and
 its X button, and Clear All all keep operating on the same query signal, and the
--- a/docs/superpowers/specs/2026-07-25-analyzer-velocity-default-design.md
+++ b/docs/superpowers/specs/2026-07-25-analyzer-velocity-default-design.md
@@ -1,0 +1,134 @@
+# Default sale-velocity filter on the analyzer tools
+
+**Date:** 2026-07-25
+**Status:** Approved, ready for implementation
+
+## Problem
+
+A user who opens `/analyzer` or `/vendor-resale` for the first time sees every
+item that clears the profit threshold, including items that sell once a month.
+Those rows look like the best flips on the page — high profit, high ROI — and
+they are the ones a newcomer is most likely to buy and then sit on. The tools
+already have a sale-velocity filter that fixes this, but it is off by default and
+buried behind "More Filters," so the people who need it most never find it.
+
+## Goal
+
+A bare visit to any of the four analyzer tools lands on a velocity-filtered view,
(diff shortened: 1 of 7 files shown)


akarras:
## Review: verified, approved

Built and browser-tested on a branch with #978, #980, #982, #983 merged together (they touch disjoint files; all four merged clean).

**Gates** — `cargo fmt --all --check` clean, `cargo clippy --all-targets -D warnings` clean (full workspace), `cargo test -p ultros-app --lib` = **211 passed**, including `query_defaults::test::default_max_sale_time_parses_to_one_day`.

**Browser (local SSR, Cerberus):**
- `/flip-finder/Cerberus` → lands on `?next-sale=1d`, hydrates clean (`boot-progress` absent).
- `/vendor-resale/Cerberus` → lands on `?next-sale=1d`; **"More Filters" button is gone** as intended.
- Max Sale Time renders in the **primary** toolbar on both, visible, `value="1d"`. The `MAX SALE TIME` header shows in the toolbar text.
- `?next-sale=` (explicit empty) is **not** re-seeded on either page — the contract you documented holds exactly.
- i18n: all four keys you reuse (`analyzer_placeholder_7d_12h`, `analyzer_tooltip_duration_format`, `{analyzer,vendor_resale}_filter_max_sale_time_label`) confirmed present in all 7 locales. No new keys.

**The vendor-resale sanity check you asked for — here's the number:**

| `/vendor-resale/Cerberus` | results |
|---|---|
| `?next-sale=` (unfiltered) | **1341** |
| `?next-sale=1d` (new default) | **34** |

A **97.5% reduction**. 34 rows is still a usable list and this is the tradeoff you explicitly documented (no-sales-history items aren't good recommendations), plus the control is now visible and one click from cleared — so I'm not treating it as a blocker. But it's a much bigger cut than "may shorten the list noticeably" suggests, and it's worth knowing that most of what vendor-resale surfaces today has no `sale_summary`. If that reads as too aggressive in production, `3d` or `7d` would be the lever.

**Not verified:** the input-typing clear path. The in-app browser pane renders at 0x0 here so I could only drive the page via JS, and synthetic `input` events don't reach Leptos's delegated handlers — I confirmed that with a control (the untouched `profit` input ignores them identically), so it's a harness limit, not your change. The URL-level behavior above covers the substantive contract.

Two notes on the PR text: bare `/flip-finder` is the landing page with the world picker, not a table — the seeding correctly lives on `/flip-finder/:world` (`AnalyzerWorldView`), so the description is slightly off but the code is right. And `SsrMode::InOrder` on the item routes is already on main, unrelated to this.

PR #979, opened 2026-07-26
docs: spec first-load boot messaging

akarras:
## What

Adds a design spec for telling users that a first load is slow *because* it is warming caches that make every later load fast. **Spec only — no implementation yet.**

[`docs/superpowers/specs/2026-07-26-first-load-boot-messaging-design.md`](https://github.com/akarras/ultros/blob/claude/first-load-wasm-messaging-9f51d5/docs/superpowers/specs/2026-07-26-first-load-boot-messaging-design.md)

## Why

A cold load downloads both the `.wasm` bundle and the game-data `.rkyv` archive before the page is interactive. Both get cached afterwards — the wasm by the HTTP cache, the archive in IndexedDB — so subsequent loads are dramatically faster. The boot indicator communicates none of this: it says `Loading…` and nothing more, so a slow first load is indistinguishable from a broken site.

Note this is not only a *first ever* load. The IndexedDB key is `{data_version}-{lang}`, so a game-data version bump puts returning users back on the cold path.

## Design summary

- **Signal** — the wasm dispatches a new `ultros:game-data-download` boot event on the IndexedDB miss path in `try_populate_xiv_gen_data_internal`, where it already knows the answer.
- **Gate** — that event arms a 1200ms timer; if `ultros:hydrated` lands first, nothing changes, so a fast cold load stays silent. The 30s watchdog extends to 60s on the cold path, since a first load on slow mobile can legitimately exceed 30s and a false failure message is worse than silence.
- **Presentation** — reuses the existing `#boot-progress-status` label, adding `max-width`/wrapping. No new elements, no layout shift.
- **i18n** — threads `locale` into `shell()` so all five boot strings resolve through `leptos-i18n`'s `td_string!` instead of staying hardcoded English, per the `CLAUDE.md` rule.

## Reviewer notes

**The one non-obvious decision:** the boot script deliberately does *not* probe IndexedDB itself. `indexedDB.open('ultros')` with no version creates the DB at version 1 with **no object stores**; rexie's `Rexie::builder().version(1)` would then see version 1 already present, skip its upgrade callback, and never create `game_data` — permanently breaking data caching for that user. Letting the wasm report it avoids the hazard and is strictly more accurate.

**Accepted limitation, called out in the spec:** the event can only fire after the wasm has downloaded and started, which on a truly cold load is already the slower payload. Someone on bad network sees plain `Loading…` through the worst of the wait, then gets the explanation. Covering the wasm download would mean inferring the cold path from elapsed time alone, which is less accurate.

**Two judgement calls worth a second opinion:**
- The exact wording — currently *"First load: downloading game data. Future visits will be much faster."*
- Not adding `aria-live` to the boot label. A live region that appears, mutates twice, and vanishes mid-load seems likely to produce confusing announcements, and the SSR page is already complete and readable by then. Reasoning is recorded in the spec so it's a decision rather than an oversight; happy to reverse it.

## Not included

`.jules/palette.md` shows as modified in this worktree and is **intentionally not committed**. The repo tracks two case-differing paths for one file — `.Jules/palette.md` (blob `67513b37`) and `.jules/palette.md` (blob `0ec3c590`) — which collide onto a single file on a case-insensitive filesystem, so one of them always reports dirty. Committing it would overwrite `.jules/palette.md`'s tracked content with the other path's. Worth fixing separately; agent note files are currently split across both spellings on `main`.

CI note: only a markdown file changed, so `cargo fmt`/`clippy` are unaffected by this diff.

### DIFF
--- a/docs/superpowers/specs/2026-07-26-first-load-boot-messaging-design.md
+++ b/docs/superpowers/specs/2026-07-26-first-load-boot-messaging-design.md
@@ -1,0 +1,191 @@
+# First-load boot messaging
+
+**Date:** 2026-07-26
+**Status:** Approved, ready for implementation plan
+
+## Problem
+
+A cold load of Ultros downloads two large payloads before the page becomes
+interactive: the `.wasm` bundle and the game-data `.rkyv` archive. Both are
+cached afterwards — the wasm by the HTTP cache, the archive in IndexedDB — so
+every subsequent load is dramatically faster. The user sees none of that. The
+boot indicator says `Loading…` and nothing else, so a slow first load is
+indistinguishable from a broken site, and there is no signal that waiting once
+buys fast loads forever.
+
+Worth stating plainly: this is not only a *first ever* load. The IndexedDB key
+is `{data_version}-{lang}`, so a game-data version bump puts returning users
+back on the same cold path. The messaging must cover that case too.
+
+## Current behaviour
+
+`shell()` in `ultros-frontend/ultros-app/src/lib.rs` injects an inline
+vanilla-JS boot indicator into `<head>`:
+
+- `#boot-progress` — a 2px gradient bar pinned to the top of the viewport.
+- `#boot-progress-status` — a 12px label at `top:8px;right:12px`, text
+  `Loading…`.
+
+It advances on two events dispatched by the wasm's `hydrate()` in
+`ultros-frontend/ultros-client/src/lib.rs` through the `dispatch_boot_event`
+helper:
+
+| Event | Effect |
+| --- | --- |
+| `ultros:wasm-loaded` | bar gets `.mid` (animates 75% → 92%) |
+| `ultros:hydrated` | bar gets `.done`, fades, both elements removed |
+
+A 30s watchdog flips the bar to `.error` and the label to
+`Loading is taking longer than expected — reload`.
+
+The slow work lives in `try_populate_xiv_gen_data_internal`: it opens the
+`game_data` store, looks up `{data_version}-{lang}`, and on a miss fetches
+`/static/data/{version}/{lang}.rkyv` and stores it.
+
+All four boot strings (`Loading…`, `Failed to load app`, `App crashed during
+load`, `Loading is taking longer than expected`) are currently hardcoded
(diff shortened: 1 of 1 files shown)


akarras:
## Review: approved — every technical claim in the spec checks out against the code

Docs-only, no runtime risk. I verified the spec's premises rather than just skimming it, since a spec with wrong premises misleads the implementation:

- `#boot-progress` / `#boot-progress-status` injected by `shell()` — `lib.rs:278` (CSS) and `lib.rs:281` (the inline IIFE). Confirmed, including `top:8px;right:12px` and the `position:fixed` claim that makes the wrapping change layout-shift-free.
- All four boot strings are hardcoded English in that IIFE: `Loading…`, `Failed to load app`, `App crashed during load`, `Loading is taking longer than expected`. Confirmed — the `CLAUDE.md` violation is real.
- 30s watchdog: `setTimeout(..., 30000)`. Confirmed.
- `ultros:wasm-loaded` → `.mid`, `ultros:hydrated` → `finish`; `dispatch_boot_event` at `ultros-client/src/lib.rs:312`, dispatched at `:328` and `:382`/`:405`. Confirmed.
- `try_populate_xiv_gen_data_internal` at `ultros-client/src/lib.rs:90`, keying the `game_data` store on `format!("{}-{}", xiv_gen::data_version(), get_i18n_lang())` at `:92`. Confirmed — so the "a data-version bump puts returning users back on the cold path" point is right, and worth designing for.

The IndexedDB hazard is the strongest part of the spec and is correct: `indexedDB.open('ultros')` with no version does create the DB at version 1 with no object stores, which would make rexie's `version(1)` builder skip its upgrade callback and never create `game_data`. Letting the wasm report the miss is the right call.

On your two flagged judgement calls: not adding `aria-live` is defensible for the reason you give (the SSR page is already complete and readable by then), and having it recorded as a decision is the right outcome either way. The wording reads fine to me.

Also agree with leaving `.jules/palette.md` uncommitted — the repo really does track both `.Jules/palette.md` and `.jules/palette.md`, which collide on a case-insensitive filesystem, and committing it would clobber one with the other's content. Worth its own PR.

PR #980, opened 2026-07-26
fix(sparkline): portal the hover tooltip so table rows can't clip it

akarras:
The sparkline tooltip on the analyzer table gets clipped by the surrounding rows — the value bubble is sliced in half by the row below it.

## Why z-index couldn't fix it

The analyzer's scroll container ([`analyzer.rs:1239`](https://github.com/akarras/ultros/blob/main/ultros-frontend/ultros-app/src/routes/analyzer.rs#L1239)) carries `overflow-x-auto contain-layout contain-paint`. Both `overflow` and `contain: paint` clip descendants outright — stacking order never enters into it. `contain: paint` also makes the container a containing block for `position: fixed` descendants, so escaping via fixed positioning inside the cell wouldn't have worked either. The only way out is to leave the subtree.

## What changed

`ultros-frontend/ultros-app/src/components/sparkline.rs`:

- The tooltip renders through `leptos::portal::Portal` onto `<body>` with `position: fixed`, instead of `absolute` + `z-20` inside the cell.
- Hover state went from `Option<usize>` to `Option<SparkHover>`, which also carries the sparkline's viewport-space `left`/`top`/`width`, captured from `getBoundingClientRect()` on each `pointermove`. The portalled tooltip anchors off those coordinates.
- The midpoint flip (`translate(-100%,-100%)` past the halfway point) is preserved, so the tooltip still grows back over the trace rather than off toward the screen edge.
- The portal is gated on an `is_hovered` memo rather than the full hover state — otherwise the portal container would be torn down and remounted on every index change as the pointer slides along the trace. Now it mounts once per hover and only the inner style/text react.
- Dropped the now-meaningless `relative` from the wrapper span.

This is the shared `<Sparkline>` component, so Market Movers, Trends, and Recently Viewed get the same fix.

## Reviewer notes

- **One known limitation:** because positioning is captured at pointermove time, scrolling while the pointer sits motionless over a sparkline leaves the tooltip briefly stale. In practice Chrome and Firefox dispatch boundary/pointer events after a scroll changes what's under the cursor, so it re-anchors or dismisses itself. I opted not to add a window scroll listener for this — it would need `cfg`-gating for the non-hydrate build (see `components/tooltip.rs`) for a case that self-corrects.
- The existing `<Tooltip>` component already portals the same way, so this brings the sparkline in line with it. It uses `use_element_bounding` (which does track scroll); the sparkline needs per-sample x positions rather than element-center, hence the hand-rolled rect.

## Verification

- `cargo clippy -p ultros-app --target wasm32-unknown-unknown --no-default-features --features hydrate -- -D warnings` — running at the time of writing; I'll follow up on this PR if it turns up anything. There are no `cfg` branches in the change, and `Portal` is not feature-gated, so ssr and hydrate compile the same code.

Not visually verified in a running browser.

### DIFF
--- a/ultros-frontend/ultros-app/src/components/sparkline.rs
+++ b/ultros-frontend/ultros-app/src/components/sparkline.rs
@@ -6,6 +6,7 @@
 //! a dot on the trace and a micro-tooltip with the value and how long ago
 //! that sample was. Sparkline series are hourly VWAP, oldest first.
 
+use leptos::portal::Portal;
 use leptos::prelude::*;
 
 use ultros_charts::charts::sparkline::build_sparkline;
@@ -13,6 +14,17 @@ use ultros_charts::components::color_attr;
 use ultros_charts::scale::short_number;
 
 use crate::i18n::{t_string, use_i18n};
+
+/// Hover state for the tooltip. Carries the sparkline's viewport-space
+/// geometry alongside the index so the portalled tooltip — which lives on
+/// `<body>`, not next to the `<svg>` — can position itself.
+#[derive(Clone, Copy, PartialEq)]
+struct SparkHover {
+    index: usize,
+    left: f64,
+    top: f64,
+    width: f64,
+}
 
 #[component]
 pub fn Sparkline(
@@ -49,7 +61,10 @@ pub fn Sparkline(
         .join(" ");
     let stroke = color_attr(&model.color);
     let model = StoredValue::new(model);
-    let hover = RwSignal::new(None::<usize>);
+    let hover = RwSignal::new(None::<SparkHover>);
+    // Split out so the portal mounts/unmounts once per hover rather than on
+    // every index change as the pointer travels along the trace.
+    let is_hovered = Memo::new(move |_| hover.with(|h| h.is_some()));
 
     let on_pointer_move = move |evt: web_sys::PointerEvent| {
         use web_sys::wasm_bindgen::JsCast;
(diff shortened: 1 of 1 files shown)


akarras:
## Review: approved — reasoning checks out, with an honest note on what I could not see

**Premise verified.** The scroll container is real and does what you say — `analyzer.rs:1243`:

```
rounded-lg overflow-x-auto border ... content-visible contain-layout contain-paint will-change-scroll forced-layer
```

`overflow` and `contain: paint` both clip descendants outright, and `contain: paint` additionally makes the element a containing block for `position: fixed` — so escaping via fixed positioning *inside* the cell genuinely would not have worked either. Leaving the subtree is the only fix. Your z-index explanation is correct.

**Positioning math is equivalent.** `rect` comes from `evt.current_target()`, which is the wrapper `<span>` carrying the `on:pointermove` — the same element the old `absolute` tooltip was positioned against (it had `relative`). So old `left:{i/(n-1)*100}%` + `top:0` + `translateY(-100%)` maps exactly onto new `left: h.left + frac*h.width` + `top: h.top` + the same transform, just in viewport space. The midpoint flip is preserved. Dropping `relative` is safe — the `<circle>` lives inside the `<svg>`, nothing else was absolutely positioned.

**SSR-safe.** The portal is gated on `is_hovered`, which is `false` during SSR, so nothing renders server-side and there is no SSR/CSR divergence to trip hydration. `Portal` is already used in `components/modal.rs` and `components/tooltip.rs`, so the pattern is established in-repo and compiles under `ssr` — confirmed, `clippy --all-targets -D warnings` is clean on the full workspace with this merged.

Gating the portal on the memo rather than the full hover state is the right call; remounting the portal container on every index change along the trace would have been a real cost.

**What I could not verify: the tooltip actually rendering.** No sparklines render on this machine — Market Movers and Trends both come back with no rows because the local dev ClickHouse's newest `sales_hourly` bucket is ~24h old and the movers/trends windows return nothing usable, and the analyzer table is empty for the same reason. On top of that the in-app browser pane sits at 0x0 here, so I could only drive pages via JS and could not deliver a real pointer event. So your "not visually verified in a running browser" caveat still stands after my pass — I could not close it either.

Merging on that basis: self-contained presentational component, established in-repo pattern, gates clean, and the current behaviour (tooltip sliced by the row below) is definitively broken. Worth a quick look on prod after deploy, since the shared component also feeds Market Movers, Trends and Recently Viewed.

PR #966, opened 2026-07-23
🎨 Palette: Add explicit label associations to forms for accessibility

akarras:
💡 What: The UX enhancement adds explicit `for` and `id` attributes to input fields and labels in groups, lists, and settings. It also adds an `aria-label` to the search list input.
🎯 Why: The user problem it solves is improving accessibility by ensuring screen readers correctly associate labels with inputs and increasing the clickable area for users.
📸 Before/After: Inputs had preceding labels without direct associations.
♿ Accessibility: Improved screen reader compatibility and keyboard navigation.

---
*PR created automatically by Jules for task [463696556744764594](https://jules.google.com/task/463696556744764594) started by @akarras*

PR #981, opened 2026-07-26
ci: stop the Actions cache quota from killing every main Docker build

akarras:
## What was breaking

Every push to `main` since 2026-07-24 12:20 failed the Docker workflow. Five consecutive main builds, ~90 minutes each, identical failure at the last step:

```
#40 exporting to GitHub Actions Cache
#40 preparing build cache for export 289.4s done
#40 sending cache export 0.1s done
#40 ERROR: not_found
ERROR: failed to build: failed to solve: not_found
```

PRs passed the whole time, which was the tell — PRs set `cache-to: ''`, so they only read. A missing blob on import is recoverable; a missing blob on export is fatal.

## Root cause

The shared **10GB Actions cache quota**, not the Dockerfile.

`rust.yml` used a plain `actions/cache`, and Actions caches are **scoped per-ref** — every PR run wrote its own full ~1.2GB copy of the same key. Six copies of `Linux-cargo-4170fe27…` (one per ref, identical hash) held **7.22GB of the 10GB budget**, and the repo sat pinned at 10.44GB. That left under 3GB for buildkit, so a `mode=max` export of the cargo-chef build evicted its own already-uploaded blobs and then 404'd finalizing the manifest.

The same eviction shows up earlier in every failing log on the import side, where `#28 CACHED` is immediately followed by `#28 ERROR: not_found` — the index survived, the blob didn't.

Timeline fits: green through 07-21, then the #966–#977 PR burst deposited those cargo caches on 07-23/24, and main has failed every build since.

**Why #977 didn't fix it:** it serialized docker-vs-docker, but the competitor was `rust.yml`. Run `30175930394` was the only build in flight and still failed. Keeping the concurrency block regardless — three 90-minute builds at once is still bad.

## Changes

- **`rust.yml`** — split into `cache/restore` + `cache/save`, saving only on main pushes. PRs stay warm via `restore-keys` against main's copy. Frees ~6GB. Save runs on lint failure too (`!cancelled()`) so a red run still leaves the next build warm.
- **`docker-publish.yml`** — buildkit cache moved from `type=gha` to a GHCR registry cache, which is outside the Actions quota, so the two workflows stop competing entirely. Registry login is ungated so PRs can read it.
- **`docker-publish.yml`** — dropped `setup-qemu-action`. We build `linux/amd64` on an amd64 runner; it did nothing but park a 32MB binfmt cache on every ref (13 of them, 419MB).

Also pruned out of band: the five merged-PR cargo caches and the stale binfmt entries. **Repo cache is now 4.0GB across 18 entries**, down from 10.44GB/36.

## Verification

Diff is YAML-only, so clippy is unaffected; `cargo fmt --all -- --check` passes. Both files validated as parseable YAML with correct step ordering.

The real proof is a green main build, which can't be had before merge — this workflow only exports cache on push. Two things to watch on the first post-merge run:

1. The registry cache starts empty, so that build is cold end-to-end and will be slow. Subsequent ones should be normal or better.
2. If `ghcr.io/akarras/ultros` is a **private** package, confirm the ungated login gives PRs read access to the `:buildcache` tag. I couldn't check visibility — my token lacks `read:packages`. `cache-from` failures are non-fatal warnings, so a wrong guess costs cache hits, not a red build.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -52,14 +52,15 @@ jobs:
               with:
                   submodules: recursive
 
-            - name: Set up QEMU
-              uses: docker/setup-qemu-action@v3
+            # No QEMU: we build linux/amd64 only, on an amd64 runner. It bought
+            # us nothing and parked a 32MB binfmt cache on every ref.
 
             - name: Set up Docker Buildx
               uses: docker/setup-buildx-action@v3
 
+            # PRs log in too — they don't push, but they read the registry
+            # build cache below, which needs auth if the package is private.
             - name: Log in to registry
-              if: github.event_name != 'pull_request'
               uses: docker/login-action@v3
               with:
                   registry: ${{ env.REGISTRY }}
@@ -81,10 +82,16 @@ jobs:
                   push: ${{ github.event_name != 'pull_request' }}
                   tags: ${{ steps.meta.outputs.tags }}
                   labels: ${{ steps.meta.outputs.labels }}
-                  cache-from: type=gha
+                  # Registry cache, not type=gha. A mode=max export of this
+                  # cargo-chef build is multiple GB and shares the repo's 10GB
+                  # Actions cache quota with rust.yml; once that filled up,
+                  # GitHub evicted blobs mid-export and every main build died
+                  # with `failed to solve: not_found`. GHCR storage is outside
+                  # that quota, so the two workflows stop competing.
+                  cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:buildcache
                   # PRs read main's cache but don't write — avoids thrashing
                   # the shared cache from short-lived branches.
-                  cache-to: ${{ github.event_name != 'pull_request' && 'type=gha,mode=max' || '' }}
+                  cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/{1}:buildcache,mode=max', env.REGISTRY, env.IMAGE_NAME) || '' }}
 
             - name: Install cosign
               if: github.event_name != 'pull_request'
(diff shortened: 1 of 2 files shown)


akarras:
## Review: diagnosis confirmed, and your open question is resolved. Approved.

**Your open question #2 — the package is PUBLIC.** An unauthenticated GHCR token request against `akarras/ultros` followed by `GET /v2/akarras/ultros/tags/list` returns **200**. So PRs can read the `:buildcache` tag with or without the login; the ungated login step is harmless and still useful if visibility ever flips. No cache-hit risk here.

**Failure signature confirmed** on the most recent main run (`30175930394`):

```
#29 ERROR: not_found
#28 ERROR: not_found
#40 ERROR: not_found
ERROR: failed to build: failed to solve: not_found
```

Import-side and export-side, exactly as you described. And it is not a one-off — **8 consecutive main Docker builds have failed**, every one since 2026-07-24.

**Corroborating the QEMU removal:** `gh cache list` right now still shows **four** `docker.io--tonistiigi--binfmt-latest-linux-x64` entries created *today* alone (30.77 MiB each, 07:46 / 08:18 / 09:45 / 15:02). That is exactly the per-ref binfmt litter you are removing, still accruing.

**Config review:** restore/save keys are symmetric (`${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}`) with `restore-keys: ${{ runner.os }}-cargo-`, so PRs prefix-match main's copy and stay warm. The save gate (`!cancelled() && push && refs/heads/main`) is right, and `continue-on-error` correctly absorbs the "key already exists" error on repeat Cargo.lock. Cache scoping works in the needed direction — PR branches can read the default branch's caches, which is exactly the fallback you rely on.

Current usage is **4.13 GB across 22 entries**, down from the 10.44 GB you describe, so the out-of-band prune landed.

Agreed that the only real proof is a green post-merge main build, and that the first one will be cold end-to-end against an empty registry cache. Given main is 8-for-8 red right now, that is clearly the better risk.

PR #982, opened 2026-07-26
⚡ Bolt: Optimization: filter before sorting in `buying_view.rs`

akarras:
💡 What: We now filter out excluded datacenters and unwanted listing qualities (HQ vs. NQ) *before* sorting the array of active listings.
🎯 Why: Sorting an array is an `O(N log N)` operation. Previously, all listings were sorted, and then a loop filtered out excluded results. Applying `O(N)` filters first significantly reduces the `N` size of the vector before the sorting algorithm has to do its work.
📊 Impact: Saves CPU cycles during reactive view rendering. Memory allocation for the `.collect()` is also reduced.
🔬 Measurement: No behavior changes should occur. Verify by profiling CPU usage during list component rendering.

---
*PR created automatically by Jules for task [7361859674854970920](https://jules.google.com/task/7361859674854970920) started by @akarras*


akarras:
## Review: approved — behaviour-preserving, verified

Traced the rewrite for equivalence and it holds. The two `continue` guards were position-independent and neither touched `remaining`, so hoisting them into a `.filter()` ahead of the sort cannot change which listings are visited or in what order; the `break` on `remaining <= 0` still sits at the top of the loop. Strictly less work: fewer elements sorted, smaller `collect()`.

One behavioural nuance worth naming (not a blocker): `sort_unstable_by_key` on the filtered subset can order *equal* `price_per_unit` listings differently than sorting the full set did. That non-determinism already existed the moment the sort was unstable, and ties are interchangeable for the buy-quantity maths, so nothing user-visible.

Gates: `cargo fmt --all --check` clean and `cargo clippy --all-targets -D warnings` clean across the workspace, verified locally on a branch with this merged alongside #978/#980/#983.

Nit, non-blocking: the `.jules/bolt.md` entry is dated **2024-11-20** — should be 2026-07-26. Same date slip as #984's journal entry.

PR #983, opened 2026-07-26
fix(api): accept 1/0 for the show_suspicious query flag (Market Trends 400s on every load)

akarras:
Found while triaging the GlitchTip backlog. Not reported by an error event — it
fails as an HTTP 400 that the frontend swallows — but it takes out a whole page.

## The bug

`/api/v1/trends/{world}` and `/api/v1/best_deals/{world}` declare their
`show_suspicious` flag as a bare `Option<bool>` inside an `axum::extract::Query`
struct. `Query` decodes with `serde_urlencoded`, whose `bool` deserializer is
`str::parse::<bool>()` — it accepts **only** the literals `true` and `false`.

The frontend sends the numeric spelling:

```rust
// ultros-frontend/ultros-app/src/api.rs:220 — get_trends_v2
"/api/v1/trends/{world_name}?window={window_days}&show_suspicious={}",
if show_suspicious { 1 } else { 0 }
```

A `Query` rejection is **not** a defaulted field — it fails the whole extraction
and axum answers `400 Bad Request` for the entire request:

```
Failed to deserialize query string: show_suspicious: provided string was not `true` or `false`
```

| Caller | What it sends | Result |
| --- | --- | --- |
| `routes/trends.rs:323` → `get_trends_v2` | `show_suspicious=0` (or `1`) on **every** load | Market Trends page 400s — renders nothing |
| `components/top_opportunity.rs:43` → `get_best_deals` | `show_suspicious=0`, always | 400, swallowed by `.ok()` → card silently blank |

This is a contract mismatch, not a client mistake — the numeric form is what
both handlers' own doc comments and the design spec specify:

- `ultros/src/web/api/best_deals.rs`: ``/// `1` or `true` skips the cross-cutting `ResaleQualityFilter```
- `ultros/src/web/api/trends.rs`: ``/// `1` / `true` bypasses the cross-cutting `ResaleQualityFilter```
- `docs/superpowers/specs/2026-05-18-clickhouse-surface-integration-design.md:114`:
  `/api/v1/trends/{world_name}?window=7|30|90&show_suspicious=0|1`

Three places document `1`/`0` as valid; the implementation accepts neither.

## The fix

Fix the server, not the client — the server is the side that deviates from its
own documented contract, and fixing it also unbreaks any third-party consumer
that followed the published `0|1` spec.

New `ultros/src/web/api/query.rs` adds `optional_flag`, a `deserialize_with`
helper accepting `1`, `0`, `true`, `false` (case-insensitive, whitespace
tolerant). A valueless `?flag=` yields `None` so the handler default applies;
anything else is still a genuine `400`, now with a message naming the accepted
spellings. Both query structs gain:

```rust
#[serde(default, deserialize_with = "super::query::optional_flag")]
pub show_suspicious: Option<bool>,
```

Client code is unchanged — it was already speaking the documented protocol.

## Verification

**Please read this section — the tests added here were _not_ executed by CI or
on my machine.** The `ultros` bin test binary does not link on Windows/MSVC
(38 unresolved `..._10ultros_app` symbols — pre-existing, unrelated to this
change), and `cargo test` is commented out in `rust.yml`. So I verified the
behaviour out-of-band, with a harness that loads the **real** shipped
`query.rs` via `#[path]` (not a copy) plus verbatim copies of both query
structs, and drives them through axum 0.8.9's real `Query::try_from_uri`:

```
running 11 tests
PRE-FIX rejection: Failed to deserialize query string: show_suspicious: provided string was not `true` or `false`
test extractor_tests::red_the_unfixed_struct_rejects_what_the_frontend_sends ... ok
test extractor_tests::green_trends_numeric_flag ... ok
test extractor_tests::green_trends_literal_flag ... ok
test extractor_tests::green_trends_omitted ... ok
test extractor_tests::green_trends_nonsense_rejected ... ok
test extractor_tests::green_deals_top_opportunity_shape ... ok
test query::tests::accepts_numeric_spelling ... ok
test query::tests::accepts_literal_spelling ... ok
test query::tests::accepts_mixed_case_and_surrounding_space ... ok
test query::tests::valueless_flag_defers_to_the_default ... ok
test query::tests::rejects_nonsense ... ok

test result: ok. 11 passed; 0 failed
```

The red case reproduces the pre-fix 400 against a struct with the old bare
`Option<bool>` shape, so this is red-before-green, not green-only.

That harness run is also what caught a defect in my first attempt: the original
`optional_flag` read `Option::<String>::deserialize`, which works under
`serde_urlencoded` but not under `serde::de::value::StringDeserializer`. Since
`#[serde(default)]` already covers the absent-key case, it now reads `String`
directly — simpler, and testable through both deserializers.

The 12 in-crate tests committed here cover the same matrix and will run wherever
the crate's test binary links.

Gates run: `cargo fmt --all -- --check` clean, `cargo clippy -p ultros
--all-targets -- -D warnings` clean.

## Sweep

Checked every other `bool` in a server `Query` struct — `cheapest_per_world`'s
`hq` is a response DTO field, not a query param. These two were the only
occurrences.

### DIFF
--- a/ultros/src/web/api/best_deals.rs
+++ b/ultros/src/web/api/best_deals.rs
@@ -18,6 +18,7 @@ pub(crate) struct BestDealsQuery {
     /// suspicious rows (Unusable / high-launder) appear with a chip.
     /// Default false — these are the rows the Flip Finder used to surface
     /// gil-trader laundering on (e.g. Copper Wristlets 3 → 18.9M).
+    #[serde(default, deserialize_with = "super::query::optional_flag")]
     pub(crate) show_suspicious: Option<bool>,
     /// Cap on returned rows. Default 50, clamped to [1, 200].
     pub(crate) limit: Option<u32>,
@@ -97,4 +98,53 @@ pub(crate) async fn get_best_deals(
         .collect();
 
     Ok(Json(dtos))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::BestDealsQuery;
+    use axum::extract::Query;
+    use axum::http::Uri;
+
+    fn extract(query: &str) -> Result<BestDealsQuery, String> {
+        let uri: Uri = format!("http://ultros.app/api/v1/best_deals/Sargatanas?{query}")
+            .parse()
+            .expect("test URI should parse");
+        Query::<BestDealsQuery>::try_from_uri(&uri)
+            .map(|Query(q)| q)
+            .map_err(|rejection| rejection.body_text())
+    }
+
+    /// `TopOpportunity` always sends `show_suspicious=0` alongside the other
+    /// params; before the flexible flag parser that 400'd the request and the
+    /// component silently rendered nothing.
+    #[test]
+    fn numeric_show_suspicious_is_accepted_with_other_params() {
+        let q = extract("min_profit=1000&filter_sale=Week&limit=5&show_suspicious=0")
+            .expect("`0` must extract");
+        assert_eq!(q.min_profit, Some(1000));
+        assert_eq!(q.filter_sale.as_deref(), Some("Week"));
+        assert_eq!(q.limit, Some(5));
+        assert_eq!(q.show_suspicious, Some(false));
+    }
+
+    #[test]
+    fn literal_show_suspicious_still_works() {
+        assert_eq!(
+            extract("show_suspicious=true")
+                .expect("`true` must extract")
+                .show_suspicious,
+            Some(true)
+        );
+    }
+
+    #[test]
+    fn omitted_flag_defaults_to_none() {
+        assert_eq!(
+            extract("limit=10")
+                .expect("omitted flag must extract")
+                .show_suspicious,
+            None
+        );
+    }
 }
--- a/ultros/src/web/api/mod.rs
+++ b/ultros/src/web/api/mod.rs
@@ -9,6 +9,7 @@ mod market_heat;
 mod market_pulse;
 mod movers;
 pub(crate) mod push;
+mod query;
 pub(crate) mod real_time_data;
 mod recent_sales;
 mod resale_quality;
--- a/ultros/src/web/api/query.rs
+++ b/ultros/src/web/api/query.rs
@@ -1,0 +1,82 @@
+//! Query-string helpers shared by the JSON API handlers.
+
+use serde::{Deserialize, Deserializer, de};
+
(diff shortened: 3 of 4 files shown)


akarras:
## Review: verified end-to-end in a browser, approved

The diagnosis is exactly right, and I confirmed the payoff in a real browser rather than only at the unit level.

**Independent confirmation of the bug:** frontend sends the numeric spelling at `api.rs:157` (`get_best_deals`) and `api.rs:220` (`get_trends_v2`); `serde_urlencoded`'s `bool` is `str::parse::<bool>()`, so `Query` extraction failed the whole request.

**Live server, running this branch:**

```
/api/v1/trends/Cerberus?window=7&show_suspicious=0        -> 200
                                 show_suspicious=1        -> 200
                                 show_suspicious=true     -> 200
                                 (omitted)                -> 200
                                 show_suspicious=banana   -> 400
/api/v1/best_deals/Cerberus?min_profit=1000&filter_sale=Week&limit=5&show_suspicious=0 -> 200   (exact TopOpportunity shape)
```

400 body reads: `Failed to deserialize query string: show_suspicious: invalid boolean flag "banana": expected one of 1, 0, true, false` — the message does name the accepted spellings.

**The real proof:** loading `/trends/Balmung` in the browser, the network log shows the page's own call

```
GET /api/v1/trends/Balmung?window=30&show_suspicious=0 → 200 OK
```

That is precisely the request that returned 400 before this change, and it now returns 500 items. Market Trends' data path is unblocked.

(Unrelated: the page still renders "0 results" despite receiving those 500 items — client-side filtering in `routes/trends.rs`, which **no** PR in this batch touches (`git diff origin/main` on that file is empty), and the item ids coming back from the local dev ClickHouse are odd low ids. Pre-existing and separate from this fix.)

**Sweep independently re-run and confirmed:** the only `Query<…>` extractors server-side are `best_deals`, `trends`, and `movers`; `MoversQuery` has no bool, and none of `ExtendedHistoryQuery` / `ListActivityQuery` / `SearchQuery` / `IconQuery` / `LoginParameters` / `RedirectParameters` / `TestLoginParams` contain a bool. These two were the only occurrences.

Gates: fmt clean, `clippy --all-targets -D warnings` clean across the workspace. Your note that the in-crate tests can't run on Windows/MSVC is correct and worth keeping — but the live-server matrix above covers the same ground with the real extractor.

PR #984, opened 2026-07-26
🛡️ Sentinel: [HIGH] Fix Open Redirect bypass in login parameter

akarras:
🛡️ Sentinel: [HIGH] Fix Open Redirect Bypass via Whitespace

🚨 Severity: HIGH
💡 Vulnerability: Open Redirect bypass in the `safe_login_next` validation function.
🎯 Impact: Attackers could craft a malicious login redirect URL using whitespace characters (e.g., `/\t/evil.com` or `/ /evil.com`) which passes the check for starting with `/` and avoiding `//`, but browsers might normalize this and redirect the user to a malicious external site.
🔧 Fix: Modified `safe_login_next` to also reject redirect URLs containing spaces and tabs (`\t`), and added test cases. Added learning to Sentinel's journal.
✅ Verification: Ran `cargo test -p ultros safe_login_next` to ensure the new tests pass.

---
*PR created automatically by Jules for task [15296084953595665188](https://jules.google.com/task/15296084953595665188) started by @akarras*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -12,7 +12,7 @@
 **Vulnerability:** Incomplete Cookie Scope Configuration
 **Learning:** The `discord_auth` cookie was being set during a `/redirect` endpoint without an explicitly set `Path=/`. This scopes the cookie to the `/redirect` path, meaning the browser wouldn't send the auth cookie to other paths (like `/api/v1/user`), effectively breaking authentication outside that route. Other cookies in the same file were properly using `cookie.set_path("/")` or `CookieBuilder` with `.path("/")`.
 **Prevention:** Always explicitly set `cookie.set_path("/")` for application-wide authentication or session cookies to ensure they are sent to all relevant routes.
-## 2024-05-24 - [Fix Open Redirect Bypass via Whitespace]
+## 2026-07-26 - [Fix Open Redirect Bypass via Whitespace]
 **Vulnerability:** Open Redirect bypass
 **Learning:** Browsers sometimes normalize or ignore whitespace characters like tabs (`\t`) and spaces (` `) in URL paths. A URL like `/\t/evil.com` or `/ /evil.com` starts with `/` and thus bypasses a strict `starts_with("//")` protocol-relative check, but the browser may resolve it to `//evil.com` and redirect the user.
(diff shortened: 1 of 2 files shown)


akarras:
## Review: real vulnerability, verified with an executed red-before-green proof. Approved.

This one is genuine, and the PR body slightly undersells why it is reachable. Recording the confirmation:

**Why it reaches the check:** `next` arrives through `Query<LoginParameters>`, and `serde_urlencoded` **percent-decodes**. So a percent-encoded tab in the query string reaches `safe_login_next` as a literal tab character. `next.trim()` does not help, because the tab is *internal*, not leading or trailing.

**Why the browser completes it:** the WHATWG URL parser strips **all** ASCII tab/LF/CR anywhere in the input. A path whose second character is a tab therefore normalizes to a protocol-relative URL and leaves the site. LF and CR were already rejected here; tab was the one remaining hole, so this closes the class rather than patching a single string.

**Impact confirmed:** `safe_login_next` is the only gate on both ends — the `login_next` cookie write in `begin_login` (oauth.rs:165) and the post-login `redirect_to` in the OAuth callback (oauth.rs:243). A bypass is a full open redirect after Discord login.

**Executed proof.** The `ultros` crate's test binary does not link on Windows/MSVC, so I ran the function standalone (pure `str` logic, so a verbatim copy is faithful) against both the pre-fix and post-fix bodies, plus a helper that reproduces the browser's tab/LF/CR stripping:

- **RED:** the pre-fix body *accepts* the tab-bearing path, and the browser-normalization helper turns the accepted value into a protocol-relative off-site URL. Vulnerability reproduced.
- **GREEN:** the post-fix body rejects it, rejects the space variant, and rejects all three browser-stripped characters.
- Every pre-existing assertion still passes: normal paths (`/list/5`, `/`) still round-trip, and the protocol-relative / CRLF-header / absolute-URL / backslash cases still return `None`.

11 checks, all green, red before green in one run.

**On the space check:** it is not load-bearing for this particular bypass (browsers percent-encode an internal space, which stays same-origin) — it is defense-in-depth, and harmless here: no Ultros route carries a raw space (item/list/retainer paths are numeric ids), and a rejection falls back to `/` rather than erroring.

Nit, non-blocking: the `.jules/sentinel.md` entry is dated **2024-05-24**; should be 2026-07-26. Same class of bot date slip as #982's `2024-11-20`.

PR #989, opened 2026-07-28
Item view layout: merged listings table, jump nav, and section reorder (Phase 1)

akarras:
Phase 1 of an item-view layout revamp. Fixes four reported problems on `/item/:world/:id` without introducing the lens concept the design doc plans for later.

Design: `docs/superpowers/specs/2026-07-26-item-view-layout-design.md`
Plan: `docs/superpowers/plans/2026-07-26-item-view-layout-phase-1.md`

## What changed

**Stat names no longer truncate.** `ItemStats` rendered `lg:grid-cols-4` inside the header's `minmax(320px,1.2fr)` track, leaving ~40px per name, so "Vitality" showed as "Vi…". Now two columns with a tighter gap.

**The two listings tables are one.** `HighQualityTable` and `LowQualityTable` had identical columns, stacked, each with its own "Show more" that expanded to *every* listing — the single largest contributor to page height. They're replaced by `ListingsPanel`: one table with an All/HQ/NQ segmented filter, a row count, and the datacenter-exclusion controls folded into the header instead of occupying their own section. Cross-quality price comparison no longer requires scrolling between two tables.

**The table is bounded.** `ListingsTable` is capped at 26rem with an internal scroller and a sticky header row, so "Show more" grows the box rather than the page.

**There's a way to jump between sections.** Two-tier chrome: the ~30 crawlable sibling-world links in `WorldMenu` stop being sticky and stay in normal flow, and a new slim `SectionNav` below them carries an in-page jump nav (Overview / Listings / History / Sources / Related). It pins itself via `position: sticky` as the pills scroll past — deliberately **no scroll listener and no client-only state**, so this adds zero new SSR/hydration surface on a route with a long history of tachys hydration panics (#6831, #6864, #6865).

**Market share moved to the bottom.** Per-world supply distribution answers a research question; it was rendering third, above the chart. It now sits below the sale history.

## Two bugs found in passing

Neither was in scope; both are real and fixed here.

1. **`WorldButton` dropped the query string** — switching worlds silently discarded the reader's `?exclude-worlds=` filter. Now preserved via a tested `item_href` helper.

2. **World pill hrefs diverged between server and client.** `leptos_router`'s `Url::escape` uses `NON_ALPHANUMERIC` under the `ssr` feature but `encodeURIComponent` on the client, so every hyphenated scope name — which is *all* the region names — rendered as `North%2DAmerica` server-side and `North-America` after hydration. `item_href` encodes against `encodeURIComponent`'s unreserved set so both sides agree.

## Also

`CLAUDE.md`'s submodule instructions were wrong and cost real time. `git submodule update --init --recursive --depth=1` fails three separate ways here: the shallow fetch leaves `universalis-assets` empty while reporting success, `ffxiv-datamining`'s clone dies with `Connection reset by peer`, and because that abort is command-wide `classjob-icons` ends up registered with zero files while `git status` calls it clean. Replaced with per-module `--reference` against the main clone plus explicit post-init verification, since exit codes can't be trusted there. Also documented reading `check_ci.sh`'s exit code without a pipe, and that clippy can be OOM-killed (137) rather than failing a lint.

## Reviewer notes

**These changes have not been viewed in a browser.** This is the main gap. Local verification was blocked — no reachable postgres (the Homebrew `postgresql@14` install is broken against icu4c@77) and Docker Desktop wouldn't finish starting. Everything automatable passes, but nobody has looked at the rendered page, and this is a layout change. Please pull the branch and eyeball it before merging.

**Accessibility tradeoff worth a look:** the jump nav is plain anchor links with `scroll-mt-16` on each target. Scrollspy highlighting of the active section was deliberately left out — it needs an `IntersectionObserver` behind a client-only guard, and adding client-only reactive state to this route seemed a poor trade for a highlight.

## Test plan

- [x] `cargo test -p ultros-app --lib` — 228 passed, 0 failed (18 new)
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] Locale parity — 1520 keys present in all 7 files (9 new strings, real translations, no English stubs)
- [x] All five jump-nav anchors verified to resolve to real ids
- [ ] **Visual check of the rendered page — not done, see above**
- [ ] Confirm hydration is clean in the browser console on `/item/<world>/<id>` and after a world switch


akarras:
**Browser-verified.** Built and served this branch locally (`cargo leptos serve`, postgres + ClickHouse live) and drove it through `/item/Cerberus/40644`. Both unchecked boxes in your test plan are now closed — plus one real finding.

## Test plan — the two open items

- [x] **Visual check of the rendered page.** SSR returns 200 (1.17MB). All five section anchors (`#overview`/`#listings`/`#history`/`#sources`/`#related`) exist exactly once and all five jump-nav hrefs resolve to real elements.
- [x] **Hydration clean on `/item/<world>/<id>` and after a world switch.** `boot-progress` is absent on load *and* after a client-side world switch (Cerberus → Europe), i.e. hydration completed both times. Console was clean on initial load.

## Verified working

| claim | result |
|---|---|
| Two listings tables merged, All/HQ/NQ segmented filter | ✅ all three buttons present |
| Table bounded at 26rem with internal scroller | ✅ `max-height: 416px`/`432px` (= 26 × 16px), scrollHeight 499 > clientHeight 432 |
| Sticky header row inside the table | ✅ pins at the container's top edge under scroll |
| Stat names no longer truncate | ✅ no clipped stat labels (only a 1px-wide hidden "Add To List" node) |
| `WorldButton` preserves the query string | ✅ 17 world links carry `?exclude-worlds=` |
| `item_href` SSR/CSR encoding | ✅ `/item/North-America/40644` renders with a literal hyphen — **zero** `%2D` occurrences |

The `COMPONENT_UNRESERVED` set matches `encodeURIComponent`'s unreserved set (`A-Za-z0-9-_.!~*'()`) exactly, and `SectionNav` is genuinely inert hydration-wise — static markup, no scroll listener, const-array iteration so no ordering hazard.

## 🔴 The jump nav does not actually stick

`SectionNav` renders, but `position: sticky` never engages, because this branch keeps `body { overflow-x: hidden }`:

```
getComputedStyle(body) -> overflowX: "hidden", overflowY: "auto"   // y forced to auto
scroll 1200px -> nav rect.top: 474 -> -726                          // scrolls away; a working sticky pins at 0
```

This is exactly the mechanism **#991** documents and fixes: a non-`visible` value on one axis forces the other to compute to `auto`, making `body` a scroll container that never scrolls, so every sticky descendant resolves against a dead scrollport. The bar scrolls away like ordinary content, so "it pins itself as the pills scroll past" doesn't happen — the jump links still work, but the persistent-nav part of the feature is inert standalone.

The table's sticky header is unaffected because it lives inside its own `overflow-y: auto` container, which is a real scrollport.

**Suggested sequencing: land #991 first, then rebase this.** #991 already audited the blast radius of removing that declaration, and duplicating the one-line CSS fix here would just collide with it. (These two also both rewrite `item_view.rs`, so a rebase is needed regardless.)

## One thing I could not attribute

A client-side **world switch** emits ~11 `RefCell already borrowed` panics from `js-sys` `futures/task/singlethread.rs:142`. Instrumented attribution:

| action | panics |
|---|---|
| item page load | 0 |
| SectionNav anchor jump | 0 |
| home → /flip-finder client nav | 0 |
| **item-page world switch** | **11** |

Hydration still completes and the app stays functional, and this PR touches no async machinery — the only `Resource` lines in the diff relocate the `listing_resource` parameter as the tables merged. It matches the known "signals touched after `.await` in a disposable route component" class on this route (a world switch disposes and recreates it mid-flight), and `RefCell already borrowed` cascades are a known debug-build artifact here that release tolerates. **I did not build main to prove it pre-existing, so flagging rather than asserting.**

PR #990, opened 2026-07-28
feat: server-side sale aggregation and LOD chart rendering

akarras:
Moves the item-page price chart from client-side bucketing of raw sales to a pre-bucketed ClickHouse aggregate. This is step 1 of 4 in a sale-history chart revamp; it is invisible to users but unblocks the rest.

## The problem

The chart fetched up to 10,000 raw sales from `/api/v1/extended_history`, bucketed them in WASM, and emitted one `Node::Circle` per sale. Three consequences:

- **History was capped by row count, not time.** The 10k limit is ordered by recency, so a popular item at region scope reached back only days. Long-run history was unreachable at any scope.
- **The DOM was the bottleneck before the database was.** 10,000 sales meant 10,000 SVG elements — a ceiling lower than the data ceiling.
- **Bucketing was recomputed in the browser** on every resize step, toggle and grouping change.

Meanwhile `sales` in ClickHouse is a full-history mirror ordered by `(item_id, hq, world_id, sold_date, pg_id)`, so a per-item aggregate is a prefix scan — the cheapest shape that table supports. Nothing used it.

## What changed

**New endpoint** `GET /api/v1/price_series/{world}/{itemid}` — params `from`, `to`, `bucket`, `group`, `hq`. Returns per-series OHLC, VWAP inputs and quantiles per time bucket. Scope names resolve exactly as `extended_sale_history` does, so world/DC/region all work.

**Payload size is now flat in history depth.** The server picks a bucket width from the requested span, so four years of weekly buckets across 8 worlds is ~1,664 points — a sixth of what we shipped before. The 10k cap stops limiting *time* and starts limiting *detail*.

**Grouping moved server-side.** Quantiles forced this: a datacenter's p50 is not any function of its worlds' p50s, so the range band could not be re-aggregated in the browser. Rather than ship columns with two different composition rules, the server groups. Cost is that changing group-by is now a fetch; the response is small and cached on both sides.

**Rendering.** New batched `Node::Path` scene primitive collapses many marks sharing a fill into one node; raw sale dots become one path per series instead of one circle per sale, and only ship at all when the window holds ≤ 2,000 sales.

**Caching.** In-process TTL cache keyed on the full parameter tuple, plus a matching `Cache-Control`. Open-ended windows snap `to` down to the current bucket boundary so live views share an entry instead of minting a key per second.

## Deliberate decisions worth reviewing

- **No `FINAL` on the aggregate.** `sales` is a `ReplacingMergeTree` whose duplicates are exact repeats; at aggregate scale an unmerged duplicate shifts a bucket's VWAP imperceptibly, while `FINAL` over a full-history scan is expensive. Accuracy-for-cost trade, documented at the query.
- **No rollup table.** A `sales_daily` mirroring `sales_hourly` would follow the established pattern, but the aggregate is already on the primary-key prefix, and a rollup adds a backfill to own plus a second definition of "what a bucket is" that must stay consistent or the chart visibly changes shape as you zoom across the tier boundary. Ship on-demand, measure, promote if p99 justifies it — the endpoint contract does not change.
- **Item card now shows a fixed 30-day window** instead of whatever span the last 200 raw sales happened to cover. Behavior change.
- **`filter_outliers` toggle removed** rather than left inert — it operated on raw sales, and neither the layout nor the server applies it now.
- **`MAX_BUCKETS` is a soft cap.** Exceeding it widens the bucket a ladder step and retries rather than truncating, since truncation drops the oldest data — which is the data the caller asked for. Once the ladder is exhausted the response ships as-is; the doc comment says so.

## A bug caught in review

The first cut fetched raw sales from Postgres via `get_compact_sale_history`, which has no date bound — it returns the most recent N *as of now* — and then filtered to `[from, to)` in Rust. For any historical window this returned `Some(vec![])`: "raw data available, and there is none," while the bucket series showed real activity in the same window.

Fixed by sourcing raw sales from ClickHouse through a shared `window_predicate()` helper, so the aggregate and the raw fetch build the identical `WHERE` clause from one function and cannot drift apart. `raw_sales_respects_the_requested_window` in `price_series_smoke.rs` is the test that would have caught it.

## ⚠️ The ClickHouse SQL has never been executed

There is no Docker or ClickHouse on the machine this was developed on. `argMin`/`argMax` orientation, `transform()` arity, `toStartOfInterval` syntax and `quantileExact` semantics were verified by reading ClickHouse's source, **not by running the query**.

`ultros-clickhouse/tests/price_series_smoke.rs` contains five gated tests that characterize all of it — including that `open`/`close` key on `sold_date` rather than insertion order, and that a datacenter's p50 is not derivable from its worlds' p50s. They compile and skip here.

**Please run these against a live ClickHouse before merging:**

```
docker run --rm -d -p 8123:8123 -e CLICKHOUSE_DB=ultros \
  -e CLICKHOUSE_USER=ultros -e CLICKHOUSE_PASSWORD= \
  --name ch-test clickhouse/clickhouse-server
ULTROS_CH_INTEGRATION=1 cargo test -p ultros-clickhouse
```

## Verification

- `./check_ci.sh` exits 0 (fmt + clippy `-D warnings`, full workspace, all targets).
- Per-crate suites: `ultros-charts` 72 + 2, `ultros-api-types` 134, `ultros-clickhouse` 22 (+5 gated), `ultros` 6, `ultros-app` 209.
- Not verified: the endpoint end-to-end, the chart in a browser, and the item-card PNG output — all need a running ClickHouse.

## Follow-on work

Specs for the remaining three steps are in `docs/superpowers/specs/`: chart modes and toolbar (candlestick, percentile band, density heatmap), comparison via small multiples with a shared crosshair, and a game-history service with per-region patch tracks for milestone bands.


akarras:
**Ran the ClickHouse tests against a live server — and they found a real defect. Fixed in e2acc9cb.**

## The SQL is correct ✅

Ran `ULTROS_CH_INTEGRATION=1 cargo test -p ultros-clickhouse --test price_series_smoke` against a live ClickHouse (the dev `ultros-clickhouse` container). All five characterizations hold — so `argMin`/`argMax` orientation, `transform()` arity, `toStartOfInterval` syntax and `quantileExact` semantics are now **executed-verified**, not just read from the source:

- `ohlc_keys_on_sold_date_not_insertion_order` — open/close key on `sold_date`, not insertion order
- `gil_and_units_reproduce_vwap` — 900 gil / 4 units = 225
- `datacenter_grouping_merges_worlds_and_recomputes_quantiles` — p50=300 across both worlds, not derivable from either world's own median
- `hq_filter_narrows_the_result`
- `raw_sales_respects_the_requested_window` — the `[from, to)` regression guard

## The test suite could not pass as written ❌

The first run failed **5/5**, and every number came back at *exactly* 5x expected:

| assertion | expected | got |
|---|---|---|
| `world1.sales` | 3 | 15 |
| `world1.gil` | 900 | 4500 |
| `rows[0].sales` | 1 | 5 |
| `dc.sales` | 4 | 20 |

All five tests seed the same `FIXTURE_ITEM` (999000001), and cargo runs tests in a binary concurrently by default — so each test observed the union of all five seeds. `mutations_sync = 1` correctly makes each `ALTER TABLE ... DELETE` synchronous, but that only orders a test against *itself*, not against the four others racing it. The doc comment's own invocation reproduces this.

The fixture id comment anticipated collision with backfilled prod data, but not collision between the tests themselves.

**Fix:** partitioned the fixture into `FIXTURE_ITEM_OHLC`/`_VWAP`/`_DATACENTER`/`_HQ`/`_RAW_WINDOW` (999000001–005) and threaded the id through `seed`. That keeps the tests parallel rather than pinning `--test-threads=1`, which would have hidden the sharp edge instead of removing it.

**After the fix: 5 passed, 0 failed in default parallel mode** (and still 5/5 serially). `cargo fmt --all -- --check` clean; `cargo clippy -p ultros-clickhouse --all-targets -- -D warnings` clean.

## Not covered by this run

The endpoint's HTTP layer, caching/`Cache-Control` behavior, the `MAX_BUCKETS` ladder, and the `Node::Path` batched rendering are all untouched by these tests — this run validates the aggregate SQL only. Worth noting the deliberate no-`FINAL` tradeoff is invisible to a 4-row fixture; it needs prod-scale data to characterize.

Also flagging: this branch and #989 both rewrite `item_view.rs`, so whichever lands second needs a rebase.

akarras:
**Rebased onto main and re-verified — merging.**

#989 and #991 landed first, so I merged current main into this branch (`0b4a83b7`). Conflict surface was smaller than expected: **only `.gitignore`** actually conflicted — both sides had added a `.superpowers/` rule with different comment wording; kept main's. Everything else auto-merged, including all 7 locale files and **`item_view.rs`**, which #989 had rewritten.

A clean textual auto-merge of two independent rewrites of the same file proves nothing on its own, so I re-verified rather than assuming:

- `cargo clippy -p ultros-app --all-targets -- -D warnings` — clean (that's where the auto-merged `item_view.rs` lives)
- **CI "Lint, checks, and tests" — pass (8m41s)** on the merge commit; that job is workspace-wide fmt + `clippy --all-targets -D warnings`, so it covers the server crate and the new endpoint
- **ClickHouse integration tests re-run on the merged base — 5/5 pass** in default parallel mode

The Docker build passed on this branch's pre-merge commit (1h19m) and is re-running now; the merge only pulls in already-green main plus a one-line `.gitignore` resolution.

PR #991, opened 2026-07-29
Flip Finder: spreadsheet shell, derived metrics, saved views

akarras:
Rebuilds `/flip-finder/:world`. Spec: `docs/superpowers/specs/2026-07-27-flip-finder-redesign-design.md`. Plan: `docs/superpowers/plans/2026-07-27-flip-finder-redesign.md`.

## Why

Measured on live prod at 1280×720, the first data row sat at **y=827** — the entire viewport was chrome. Two structural causes: every filter rendered twice (a toolbar input, then a chip echoing the same value), and explanatory copy stood in for UI.

The default sort made it worse. Sorting by ROI put laundering artifacts on top:

```
Leather Choker   buy 2 gil   profit 213,749,998   ROI 2147483647%
Copper Ring HQ   buy 1 gil   profit  18,999,999   ROI 1900000000%
```

`2147483647` is `i32::MAX` — `((profit as f32 / cheapest as f32) * 100.0) as i32` saturating.

## What changed

- **Window-scroll virtualization.** `VirtualScroller` gains a `ScrollSource` mode so the list virtualizes against window scroll instead of a fixed-height inner container, keeping native mobile scrolling. Container mode is the default; the other 7 call sites are untouched.
- **Metrics derived from data present on every row.** ClickHouse enrichment covers ~7% of traded items (measured — see below), so Velocity, Drift and Confidence derive from the 6-sale `RecentSales` buffer, with ClickHouse preferred where present. Trend / Sales-per-day / 30d-volume become opt-in.
- **ROI demoted and fixed.** Computed in f64 and clamped rather than saturating. Still available as an opt-in column.
- **Sticky control bar**, ~76px, replacing ~474px of static chrome. Filter chips are the *only* representation of a filter — click the value to edit inline, `x` to clear, unset filters live behind `+ Filter`.
- **Sortable headers that tell the truth.** One `SortHeader` component replaces three near-identical blocks: click an inactive column to sort descending, click the active one to flip. The arrow follows the direction actually applied, and `?dir=` is now reachable from the UI.
- **Saved views** in localStorage. Every filter is already a `query_signal`, so a view is just `{name, query, world?}`. World pinning is opt-in per view. The six presets become built-in views, all moved to `last-sold=1d`.
- **Copy deleted** — tool header, calc explainer, assumption badges, features grid, tips list, plus 30 orphaned locale keys across all 7 locales.

## 🔴 Sitewide change a reviewer must look at

**`body { overflow-x: hidden }` was removed** (`style/tailwind.css`). That declaration made `body` a scroll container — a non-`visible` value on one axis forces the other to compute to `auto` — so **every `position: sticky` in the application resolved against a scrollport that never scrolls, and nothing has ever stuck.** `html { overflow-x: hidden }` is retained and the UA propagates it to the viewport, so horizontal page scroll is still suppressed.

This turns on stickiness in five places that were authored expecting it. Blast radius audited: of 8 tables under `.main-content`, 7 sit inside `overflow-x-auto` wrappers; only `/retainers` newly activates. Two mobile bugs it exposed are fixed in `c9d3c041` — the item-explorer toolbar's `top-[72px]` was clearing chrome that doesn't exist below `lg`, and the home sidebar pinned transparent over the stacked column.

**`/retainers` has not been checked logged in.** It renders one table per retainer and is the only page where `.main-content th` newly activates. A logged-in scroll is the one check that couldn't be run here.

## Merged with main (157 commits)

Main moved substantially under this branch, including work on the same file. Notable resolutions:

- **#978's `query_defaults` module kept in full**, and it closes most of what was still open here. `seed_query_default` fires from the route component rather than inside the `Suspense` closure — so a live market refetch can't silently reinstate a filter the user just cleared — and `filter_query_signal` navigates with replace/no-scroll so a keystroke isn't a history entry. `seed_query_default("next-sale", DEFAULT_MAX_SALE_TIME)` survives untouched. Its UI half (lifting Max Sale Time into the primary toolbar) is superseded: this branch deletes both toolbars and `next-sale` already has a chip.
- **#978 also settles a design question this branch had left open.** `next-sale` filters on `avg_sale_duration`, the derived buffer metric, not ClickHouse — so the default velocity filter doesn't depend on async enrichment and can't produce row churn as enrichment lands.
- **Main's realtime work adopted into the new chrome.** `<RealtimeStatus>` moved from the results-summary panel this branch removes into sticky bar row 1.
- **`SalesCadenceBadge` adopted, with its fallback unified.** Main's badge and this branch's Velocity column had independently grown the same ClickHouse-then-derived fallback. The Sales/day cell now falls back to `velocity_per_day` — the value the Velocity column displays — so the two columns can't contradict each other.
- **`analyzer_placeholder_7d_12h` restored to all 7 locales.** It was deleted as orphaned, which was true at the time; #978 then gave it a consumer in `vendor_resale.rs`.

## Known follow-ups (not blocking)

- The spec's `?vel=` velocity floor exists, is chipped and tested, but is not itself seeded as a default. #978's `next-sale=1d` covers the same ground more strictly, so the default view is no longer topped by laundering rows — but if `?vel=` is wanted as a default too, `seed_query_default` is the mechanism.
- Post-tax state has no on-screen indicator when active (the default), though it changes every profit number shown.
- The `+ Filter` and Columns popovers don't close on outside-click or Escape.
- `analyzer.rs` is large; the 14 near-identical filter-chip blocks are the obvious extraction.

Full triage of ~40 accumulated review findings is in the branch's development notes.

## Verification

```
cargo test -p ultros-app --lib   → 277 passed; 0 failed
cargo fmt --all -- --check       → exit 0
cargo clippy -p ultros-app --all-targets -- -D warnings → exit 0
```

Zero failures. The two long-standing failures this branch previously carried (`item_explorer::test_job_filtering`, `retainers::test_sort_order`) are fixed on main and now pass.

Hydration was verified on a real SSR hard load of `/flip-finder/Gilgamesh` (not a client-side nav): 76px bar, header emitting `sticky … top: 76px`, client takeover, no `tachys` panic. Column reachability and chip commit behavior verified in Chrome at 1280×720.

## Measurements behind the design

ClickHouse coverage, stratified sample of 150 items drawn from `/api/v1/recentSales/Gilgamesh` (24,428 actively-trading items), queried against `/api/v1/resale_quality/Gilgamesh` at `window_days=30`: **11 rows, 7.3%**, flat across every price band. Concrete case — item 44507 has 25 sales in 30 days per `extended_history` and `variants: []` in `item_stats`. `deep_scan_batch` has no filter that would hide them and `refresh_window` has no LIMIT, so the gap is upstream in the CH `sales` ingest. Tracked separately; nothing here depends on it.

Built-in views under the tighter 1d window, 23,174 rows on Gilgamesh: Realistic 1718→324, Big ticket 371→77, Volume 2025→701, 300% 1030→143, 500% 60→9, 100k profit 714→91. Nothing collapses; "500% return" is the thinnest and the one to re-check on a low-liquidity world.


akarras:
## 🔴 SECURITY — live API token committed; rotate it now

This branch added `.mcp.json` containing a working credential:

```json
"GLITCHTIP_TOKEN": "ff6ef133…"   (64-hex, glitchtip.ultros.app, org "ultros")
```

`akarras/ultros` is **public**, and I confirmed through the GitHub contents API that this blob was publicly readable on the pushed branch — so the token is already exposed, independent of whether this PR ever merges.

**`.mcp.json` is already in `.gitignore` (line 4, on `main` too), so it had to be force-added past that rule** — this is accidental, not intentional. It also embeds a machine-specific absolute path (`C:/Users/chw11/code/mcp-glitchtip/build/index.js`) that would not work for anyone else.

**Removed in b486f01b**, so a squash-merge can no longer carry it to `main`.

⚠️ **Removal is not remediation.** The blob stays reachable in this branch's history and GitHub serves blobs by SHA. **The token must be rotated in GlitchTip.** I did not attempt that — rotating credentials is yours to do.

---

### Minor, while I'm here

`ultros-frontend/ultros-app/Cargo.toml` re-adds `"DomRect"` to `web-sys` features; it is already present a few lines above. Harmless (cargo dedupes) but redundant.

### Verified so far

- **Locale parity clean** — 1500 keys across all 7 files, 18 added / 29 removed, and the new keys carry real translations (only `analyzer_add_filter` matches EN in German, where "Filter" is genuinely the same word — not a stub).
- **The `body { overflow-x: hidden }` analysis is correct**, and I confirmed the failure mode empirically on #989's branch: with that declaration present, `getComputedStyle(body).overflowY` computes to `auto`, and a sticky bar scrolled from `top: 474` to `-726` instead of pinning at 0. #989's `SectionNav` is inert for exactly this reason, so **this PR is a prerequisite for that one's jump nav to work at all.**
- **ROI fix is correct** — f64 arithmetic clamped to ±`ROI_DISPLAY_CEILING`, replacing the `as i32` saturation that produced `2147483647`.
- **`ScrollSource::Window` uses `SSR_FALLBACK_ROWS = 20` behind an Effect-gated `hydrated` flag** — the established anti-mismatch pattern in this repo, correctly applied.

Browser verification of the sticky blast radius (incl. `/retainers`) is running now; I'll follow up.

akarras:
**Browser-verified against a live build of this branch** (served page reports build `b486f01b`; served CSS contains `sticky-bar` ×7 and `filter-chip` ×12 and no `body{…overflow-x:hidden}`, so these measurements are genuinely this branch's code).

## The sitewide CSS change does what it claims ✅

```
getComputedStyle(body) -> overflowX: "visible", overflowY: "visible"   // no longer a scroll container
```

Sticky control bar, measured by scrolling in steps:

| scrollY | bar top |
|---|---|
| 0 | 150 |
| 100 | 50 |
| **200** | **0** ← pins |
| 300 | 0 |
| 381 | 0 |

Bar height is exactly **76px**, and the second sticky resolves at `top: 76px` — the table header stacks directly below the bar rather than under it, as designed.

**The contrast is the proof.** I ran the identical probe on #989's branch, which keeps `body { overflow-x: hidden }`: there `overflowY` computes to `auto`, and its `SectionNav` bar went from `top: 474` to `-726` under scroll instead of pinning. Same mechanism, opposite outcome — your analysis is exactly right, and **this PR is a prerequisite for #989's jump nav to function at all.**

## 🎯 /retainers — the check you flagged as unrun

Exercised via `/retainers/listings/1125` (a **public** route, so no Discord login needed) with **60 real listings** and 2702px of scroll:

- 5 sticky `th` cells, all `top: 0px`
- header pinned at **0** after an 800px scroll (from 204), nothing occluded

So the one page where `.main-content th` newly activates behaves correctly.

## Also verified

- **ROI saturation gone** — no `2147483647` anywhere in the rendered output; the f64-and-clamp fix is correct.
- **Chips are the only filter surface** ("Next Sale ≤ 1d"), with Views / Save view / Columns / Clear all / + Filter present, and the new Item / Profit / Profit-Day / Velocity / Drift columns rendering.
- **Hydration clean** on `/flip-finder/:world` (`boot-progress` absent).
- **Locale parity** — 1500 keys × 7 files, 18 added / 29 removed, real translations throughout (only `analyzer_add_filter` matches EN in German, where "Filter" is the same word).

## Two known-environmental notes, not regressions

1. `/retainers/listings/:id` leaves `boot-progress` present under this **debug** wasm build (`cargo leptos serve` without `--release`). That's the documented debug-only `leptos_meta` → tachys path on this exact route, which release tolerates.
2. A `RefCell already borrowed` cascade appears here **and** on #989's branch, on different routes. Same class on two independent branches ⇒ environmental/debug-build, not introduced by either PR.

Separately, the empty Flip Finder I first hit was the local ingestion firehose saturating the sqlx pool (`Connection pool timed out` flooding the server log) plus a genuinely thin dev dataset — after a restart the page returned 4 real rows. Not a PR issue.

## Outstanding

- 🔴 **Rotate the GlitchTip token** (see previous comment). Removed in b486f01b, but it was published on a public branch and removal is not remediation.
- Duplicate `"DomRect"` in `ultros-app/Cargo.toml` web-sys features — harmless, worth tidying.

**Recommended merge order: #991 → #989 → #990**, since #991 unblocks #989's sticky nav and #989/#990 both rewrite `item_view.rs`.

PR #993, opened 2026-07-29
feat(search): center the search bar and hint at job-set search

akarras:
Two fixes to the topbar search, both reported from using the site.

## 1. The search bar wasn't centered

`.top-bar` was a flex row with `margin-inline: auto` on `.top-bar-search` **and** `margin-left: auto` on `.top-bar-actions`. Three auto margins split the free space three ways, so the search box could never land in the middle — it sat left of center by roughly half the width difference between the actions cluster and the (desktop-hidden) hamburger.

Now a 3-column grid with equal `1fr` outer tracks:

```css
grid-template-columns: 1fr minmax(0, 480px) 1fr;
```

Columns are assigned explicitly (`grid-column: 1/2/3`) because the hamburger is `display: none` on desktop — a hidden element isn't a grid item, so auto-placement would slide the search box into column 1.

Also dropped the SearchBox root's `md:w-full sm:w-[424px]` in favor of `w-full`; that fixed 424px applied between 640px and 767px and fought the parent's sizing.

**Verified** against a standalone repro of the app-shell + topbar rules: at 1280px the search box's center is within a pixel of the bar's center. At 375px the outer tracks collapse to their content width and the search takes the remainder, so the mobile layout stays hamburger-left / search-middle / actions-right.

## 2. The hint didn't mention job sets

The index covers job gear sets (`/items/jobset/<job>`), but nothing surfaced that, so "SAM" never occurred to anyone as a query.

- Placeholder now reads **"Search items, jobs, currencies… (⌘K / CTRL K)"**.
- Focusing an empty box opens a hint panel with the three things the index covers, including clickable `SAM` / `WHM` / `BLM` chips that run the search. Those match the backend's `job equipment` documents, indexed as `"<name> (<abbreviation>)"` — so clicking `SAM` lands on Samurai's gear.
- New keys `search_hint_title` / `_jobs` / `_items` / `_tools`, translated in all seven locales.

## Drive-bys in the same component

- Clear button and hint chips `prevent_default()` on mousedown so the input keeps focus. Previously clicking clear kicked off the 250ms focusout timer, which collapsed the panel a beat after the click.
- `aria-expanded` now reflects the results listbox only, not the hint panel — the hint panel isn't a combobox popup.
- Removed a `Show` block rendering `"No results found for \"…\""` from *inside* a wrapper that is itself hidden whenever the result list is empty. It could never render, and it was a hardcoded English string. Worth flagging because `.Jules/palette.md` has a note prescribing exactly that block — the live empty state above it (`search_no_results`) is untouched and still does the job.

## Testing

- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p ultros-app --all-targets` — clean, zero warnings.
- Full-workspace clippy **not** run: the submodules aren't initialized in this worktree, so I symlinked them from the main clone to compile the frontend crate, then removed the symlinks. No crate outside `ultros-app` and `style/tailwind.css` is touched.
- Layout verified in-browser via the repro described above; the running app was not built.


akarras:
Verified — merging.

**First, a note for anyone reading the diff view:** this branch is based on `698f0ed8`, before today's five merges, so a plain `git diff origin/main` makes it *look* like it reverts #991 (removing `body`'s overflow comment and the whole `sticky-bar` utility). It doesn't. Diffed against the actual merge-base, the only CSS change is the `.top-bar` flex→grid conversion, and `git merge-tree` against current main is clean — git's 3-way merge is unaffected by the stale base.

**Centering.** The reasoning holds: `grid-template-columns: 1fr minmax(0, 480px) 1fr` centers the middle track regardless of how the outer clusters differ in width, which three competing `auto` margins genuinely cannot do. And the explicit `grid-column` assignments are load-bearing, not defensive — a `display: none` element is not a grid item, so with auto-placement the search box really would land in column 1 on desktop.

**Dead code removal — confirmed, not just asserted.** The `Show` block you removed sat inside a wrapper carrying `class:hidden=move || search_results.with(|v| v.is_empty())` (line 398 on main), while the `Show` itself fired on `search_results...is_empty()` (line 515) — the *same* predicate. It could only ever render inside an element hidden by exactly the condition that made it render. Never visible. The live empty state at line 389 sits outside that wrapper and is untouched, so behavior is unchanged.

Good catch that `.Jules/palette.md` prescribes exactly that block — the note is what produced dead code, and it's worth the palette entry being corrected rather than the block being re-added later.

**Locale parity:** 1515 keys across all 7 files; the 4 new `search_hint_*` keys are present everywhere with real translations (0 identical-to-EN).

Lint/clippy/fmt green.

PR #995, opened 2026-07-29
fix(jobset): gear-set card totals stuck on "—" after a direct load

akarras:
Fixes the report that `https://ultros.app/items/jobset/SAM?world=Gilgamesh` shows no prices in the gear-set cards on a direct load — the table below them works, and changing the world makes the card prices appear.

## Root cause

The gear-set cards render their NQ/HQ totals **outside any `<Suspense>`**. That's the whole difference:

1. SSR renders the body with the cheapest-listings resource still pending, so it ships `—` and `class="hidden"`.
2. The client has the resolved resource in the hydration payload on its *first* render — but tachys **adopts** the SSR text node without writing to it (`tachys/src/view/strings.rs`: `if !FROM_SERVER { set_text }`), while storing the *client-computed* string as that node's state.
3. So the DOM keeps `—` even though the reactive graph believes it already rendered the total. `rebuild` only patches the node when the value *changes* — which is exactly why switching worlds (a refetch, hence a different total) made the prices finally appear.

`GilOrDash` (#696) keeps the element shape stable, so this never panicked like the rest of the hydration cluster — it just silently dropped the values.

Evidence beyond reading the code:

- Direct load of `/items/jobset/SAM?world=Gilgamesh` → all 13 cards `—`. A client-side nav to the same component → real totals. So the memo, the context and the resource were all fine; only the hydration path lost the value.
- `/items/jobset/PLD/set/740` works on a direct load, and its SSR markup shows why: those totals sit inside a `<Suspense>`, so SSR emitted the fallback (`<!--s-47-o-->…`) and hydration *replaced* it rather than adopting it.

## Fix

The same `Effect`-driven `hydrated` gate already used by `<CheapestPrice>` (#740), `<RelativeTime>` (#730) and the item-explorer price sort. Effects run client-only, after the initial view is rendered, so SSR and the first CSR render both emit `—`; a frame later the flag flips, the memo recomputes to a genuinely different value, and tachys patches the DOM. Cost is one frame of `—`, same as the skeleton in the table below.

## Notes for review

- The sortable list below the cards was never affected — `<CheapestPrice>` already has this gate.
- I deliberately left the set-detail page's two `set_total` columns alone. The comment there claims they don't need the gate; I checked rather than assumed, and it's correct — they're inside a `<Suspense>`.

## Tests

- Two unit tests pin the gate: `card_totals` reports `(None, None)` pre-hydration even holding a fully populated price map.
- Unit tests structurally cannot see this seam, so `integration/jobset-card-hydration.cjs` diffs a direct (SSR + hydrate) load against a client-side navigation to the same route. **Run against production, where the bug is still live, it fails on all 13 SAM cards** with the exact symptom. It's data-independent — an empty dev DB makes it pass rather than fail spuriously.

```
BASE_URL=http://127.0.0.1:8080 npm --prefix integration run test:jobset-card-hydration
```

## Verification

- `./check_ci.sh` → exit 0 (fmt + clippy clean).
- `cargo test -p ultros-app --lib job_set_card` → 8 passed.
- **Not** verified end-to-end against a built binary — that needs `cargo leptos build` plus a DB with market data, which I didn't have. Worth running the probe against a real server before merging.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,21 @@ To exercise login flow locally:
 LEPTOS_FEATURES=test-auth ./scripts/run_e2e.sh
 ```
 
+### Targeted probes
+
+The runner only asserts on titles and body substrings, which can't see values that
+hydration silently drops. [integration/jobset-card-hydration.cjs](integration/jobset-card-hydration.cjs)
+covers that seam for the gear-set cards on `/items/jobset/<JOB>`: it reads the NQ/HQ
+totals after a direct (SSR + hydrate) load and again after a client-side navigation to
+the same route, and fails if they disagree. It needs market data to be meaningful but
+passes rather than failing when there is none, so it's safe against an empty dev DB.
+
+```bash
+BASE_URL=http://127.0.0.1:8080 npm --prefix integration run test:jobset-card-hydration
+```
+
+`JOBSET` (default `SAM`) and `WORLD` (default `Gilgamesh`) pick the route.
+
 ### Caveats
 
 - Requires a populated `.env` (DATABASE_URL, DISCORD_*, KEY) — or those vars exported directly.
--- a/integration/jobset-card-hydration.cjs
+++ b/integration/jobset-card-hydration.cjs
@@ -1,0 +1,153 @@
+// Regression probe for the gear-set cards on `/items/jobset/<JOB>` losing
+// their NQ/HQ totals on a direct (SSR + hydrate) load.
+//
+// The totals render outside any <Suspense>, so SSR ships the "—" placeholder
+// with the cheapest-listings resource still pending, while the client has the
+// resolved resource on its first render. tachys *adopts* the SSR text node
+// during hydration without writing to it, so the DOM stayed on "—" even though
+// the reactive graph held the real total — and only a later *change* to the
+// value (switching worlds, hence a refetch) ever patched it in.
+//
+// Unit tests can't see this: it only exists in the SSR-payload/hydration seam.
+// So we compare the same component reached two ways:
+//
+//   direct load  -> hydrated from SSR markup   (the broken path)
+//   client nav   -> built fresh by the router  (the known-good path)
+//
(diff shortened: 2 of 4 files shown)


akarras:
**Ran the probe you asked for — red on prod, green on this branch. Merging.**

You flagged the end-to-end check as the gap. I built this branch (`cargo leptos serve`, live postgres + ClickHouse) and ran `jobset-card-hydration.cjs` against both arms:

**Production (`https://ultros.app`, bug live) — FAILS on all 13 SAM cards:**
```
[fail] 13 gear-set card(s) lost their totals on hydration:
  - "Diadochos": direct load NQ=— HQ=—, client-side nav NQ=30,235,933 HQ=30,014,245
  - "Occult":    direct load NQ=— HQ=—, client-side nav NQ=129,899,466 HQ=—
  - "Ceremonial": direct load NQ=— HQ=—, client-side nav NQ=432,772 HQ=236,903
  … 13/13
```

**This branch (local, `WORLD=Cerberus`) — PASSES:**
```
[info] 13 card(s) on direct load, 13 after client-side nav, 13 with price data
[done] gear-set card totals survive hydration
```

The `13 with price data` line matters: the probe is designed to pass vacuously against an empty DB, so that count is what makes this a *meaningful* pass rather than a no-op. Confirmed the server was serving this branch by content — the page chrome renders build `4bf89c98`, not a stale binary.

**On the analysis:** the root cause reads correctly. `GilOrDash` keeping the element shape stable is exactly why this silently lost values instead of joining the tachys panic cluster, and the `Effect`-driven `hydrated` gate is the same pattern already load-bearing in `<CheapestPrice>` (#740) and `<RelativeTime>` (#730) — so this is consistent with the established fix for this class, not a new mechanism.

Also appreciated that you checked the set-detail page's `set_total` columns rather than trusting the comment claiming they didn't need the gate — they are inside a `<Suspense>`, so that's right.

Unit tests: `cargo test -p ultros-app --lib job_set_card` per your run; lint/clippy/fmt green in CI.

PR #996, opened 2026-07-29
update game data to 7.55

akarras:
Bumps `xiv-gen/ffxiv-datamining` from `1b5c1af` (7.51) to `c142b12` (7.55), picking up:

| commit | date | |
|---|---|---|
| `c142b12` | 2026-07-28 | 7.55 |
| `709431b` | 2026-07-22 | schema updates |
| `81fc999` | 2026-06-23 | 7.51h2 |
| `62245d8` | 2026-06-12 | 7.51h1 |

The nested `csv/cn`, `csv/ko`, `csv/tc` submodules move with it.

**Effect:** 413 newly-named items, 49 of them marketable (Monarch Whetstone, the Tule/Torna/Carwen gear sets, the Crescent housing items, etc.). These are the items currently rendering without names.

## Icons — not fixed here, and not fixable yet

`universalis-assets` is **already at its upstream HEAD** (`1586efa`, "Upload new assets for 7.51", 2026-06-04) and `master` is its only branch, so there is nothing to bump. Upstream has not published 7.55 icons.

Confirmed by measurement: all **49** new marketable items lack an `icon2x/<item_id>.png` (icons are keyed by *item id*, not the `Icon` column — see `get_item_image` in `ultros-xiv-icons/src/lib.rs`). Across all 16,843 marketable items in 7.55, 52 are missing an icon; 49 of those are the new ones.

So this PR fixes the **names**; icons stay blank until Universalis-FFXIV publishes a 7.55 asset drop. Worth re-checking that repo in a week or so.

`ultros/static/classjob-icons` is also already at HEAD — no change.

## Verification

No compile was run (the worktree's submodules are uninitialized; CI does the real build). Instead the CSV-schema risk — the only thing a data bump can realistically break — was ruled out statically:

- Extracted all **135** `column = "..."` mappings across the **31** sheets `read_data()` actually reads, expanding `count = N` arrays and marking `default_if_missing` fields optional.
- Re-implemented `read_csv_vec`'s header normalization exactly: SaintCoinach `key`-row detection, `{`/`}` stripping (`Level{Item}` → `LevelItem`), `<%>`→`Percent`, `ItemIngredient`→`Ingredient`, and the `ItemReceive`/`CountReceive`/`ItemCost`/`CountCost` multi-index remaps.
- Confirmed every required column resolves in **all 7 locales** at the new pin: 0 missing.
- **Calibration:** the same check against the old pin also scores 0, so the harness is measuring a real signal rather than passing vacuously.

Only structural change anywhere is `tc` gaining trailing **unnamed** columns in ClassJob/ClassJobCategory/BaseParam/ENpcBase — these cannot match a modeled name. `tc`'s pre-existing header-vs-data raggedness (e.g. ENpcBase 110 header / 116 data) is unchanged, just +1 on both sides.

`cargo fmt --all -- --check` passes. Clippy was not run locally (needs a full workspace compile with recursive submodules); CI covers it.


akarras:
Verified, and by a stronger route than the PR claims.

The PR rules out the schema risk statically (135 column mappings × 31 sheets × 7 locales, 0 missing) because it couldn't compile locally. That analysis matches my own notes for this exact bump (`1b5c1af` → `c142b12`: zero header changes in en/de/fr/ja; tc gained trailing *unnamed* columns, which are unmatched by name and therefore harmless; tc's `ENpcBase` is natively ragged and pre-existing).

But it doesn't need to rest on static analysis: **`.github/workflows/rust.yml` checks out `submodules: recursive` and then runs `cargo clippy --all-targets -- -D warnings`.** Clippy runs build scripts, so `xiv-gen-db`'s build script actually parses these CSVs and generates the rkyv DB during CI. A missing or renamed required column would panic the build script and fail the job.

**"Lint, checks, and tests" is green on this commit**, so the new game data is *execution-verified* to parse, not just statically checked.

The icons framing is right and worth keeping honest: names and icons are separate repos, `universalis-assets` is already at its upstream HEAD (`1586efa`, 7.51 drop), and icons are keyed by **item id** not the `Icon` column — so the 49 new marketable items stay iconless until Universalis publishes a 7.55 asset drop. This PR fixes names only, as stated.

PR #998, opened 2026-07-29
fix(scrip-sources): deterministic row ranking + working dedup (and: the page renders zero rows in prod)

akarras:
## Summary

Two bugs on `/scrip-sources`. **One is fixed here and fully verified. The second — the page renders *zero rows* in production — is diagnosed but deliberately not fixed, because the correct fix needs a game-data judgement call I can't verify autonomously.** Details in the second section; please read it, it's the bigger of the two.

## 1. Fixed: non-deterministic row ranking + ineffective dedup

`ScripSourceTable` collects rows by iterating `data.collectables_shop_items`, a `std::collections::HashMap`. `RandomState` randomizes that iteration per process, and the SSR server and the hydrating wasm client each build their own copy of the game data — so rows reach the ranking step in a **different order on the two sides**.

The ranking left ties unresolved:

```rust
SortMode::ScripAmount => results.sort_unstable_by_key(|d| Reverse(d.scrip_amount)),
SortMode::Cost        => results.sort_unstable_by_key(|d| d.cost),
```

`high_reward` is a small integer shared by hundreds of items, so ties are the common case, not the edge case. Unstable sort + tied keys = tied rows stay in *input* order, which differs per process. The list is then cut to 100, so the two sides don't merely order rows differently — they keep a **different set of rows**. That order reaches the DOM via the `VirtualScroller`'s `<For>` key, i.e. the hydration-mismatch class fixed for the item page in #960.

**Fix:** tie-break every sort mode on the stable `item_id`. `CostPerScrip` also moves from `partial_cmp().unwrap()` to `total_cmp`, dropping a latent NaN panic.

**Also fixed:** the dedup never worked. `dedup_by_key` removes only *consecutive* duplicates, but the list is sorted by a cost/scrip metric, not by item id — so an item stocked by several shops at different reward tiers rendered twice. Replaced with a first-wins `retain`, keeping the best-ranked row per item. (The existing comment already documented the intent: *"since item may appear in multiple shop lists"*.)

Three of the five new tests fail on the previous code:

| test | before | after |
|---|---|---|
| `ranking_is_independent_of_input_order` | `[1,2,3]` vs `[3,2,1]` | equal |
| `truncation_keeps_the_same_rows_regardless_of_input_order` | `[1,2,3,4,5]` vs `[10,9,8,7,6]` | equal |
| `repeated_items_collapse_even_when_not_adjacent` | `[1,2,1]` | `[1,2]` |

## 2. NOT fixed — `/scrip-sources` renders zero rows in production

While verifying the above I found the page is **completely dead**, and has to be:

```rust
41784 => ScripType::OrangeCrafters,
25199 => ScripType::WhiteCrafters,   // ...etc — these are scrip *item ids*
_     => ScripType::Other(id),
```

...is called as `ScripType::from_id(reward.currency)`. But `CollectablesShopRewardScrip.Currency` is **a small enum index, not an item id**. Every value in the current pinned game data:

```
$ awk -F, 'NR>1 {print $2}' csv/en/CollectablesShopRewardScrip.csv | sort -n | uniq -c
      1 0
    537 2
    119 4
     31 6
     13 7
```

So `from_id` returns `Other(_)` for **every** row, and both filter branches then drop it — the `else` branch does `if matches!(scrip_type, ScripType::Other(_)) { continue; }`, and with a filter active no `Other` row can equal a named variant either. **No row can ever pass, under any filter.**

Confirmed live on `https://ultros.app/scrip-sources?world=Gilgamesh` (release `698f0ed`): column headers render, `role="row-group"` count is **0**. Not an API problem — `/api/v1/cheapest/Gilgamesh` returns 200 with ~984KB of listings.

**Why I didn't fix it:** the fix is a `Currency` → scrip-type table, and I can't establish the colour tiers to a standard I'd ship user-facing labels at. What the data *does* support, unambiguously:

| Currency | rows | crafted / gathered | job level range |
|---|---|---|---|
| 2 | 1149 | 1067 / 82 | 50–98 |
| 4 | 250 | 59 / 191 | 50–98 |
| 6 | 93 | 87 / 6 | 80–100 |
| 7 | 18 | 0 / 18 | 100 |

So 2/6 are Crafters' and 4/7 are Gatherers' — that part is solid. For the tier my best guess is `2→Purple Crafters, 4→Purple Gatherers, 6→Orange Crafters, 7→Orange Gatherers` (four live scrip types post-7.0 with White removed; currency 7 being all-level-100-gathered fits Orange Gatherers' exactly). But currency 6 includes level-80 Shadowbringers items like *Lignum Vitae Lumber*, which argues against a clean Orange reading — so I'd be guessing, and a wrong guess mislabels every row. **Your call / a 2-minute in-game check settles it.**

Worth noting the structural lesson regardless of the mapping: an unrecognized `Currency` currently causes the row to be *silently dropped*, so one new expansion adding `Currency = 8` blanks the whole page again. Whatever mapping lands, unknown values should probably stay visible rather than filtered out.

## GlitchTip triage: blocked this run

No issues could be triaged — every `glitchtip_issues` call returned `Authentication failed`. Not an outage: `glitchtip.ultros.app` and `ultros.app` both return 200. `.mcp.json` resolves the token from `${GLITCHTIP_TOKEN}`, which isn't set in this environment — looks like fallout from the #991 token leak being sanitized to an env-var reference without the var being provisioned for the scheduled task. **The backlog was untouched this run; no issues were ignored or resolved.** Worth re-provisioning that env var before the next run.

Both bugs here were found by code inspection instead, using the durable rule from #960/#6831: *any HashMap iteration whose order reaches the DOM is an SSR/CSR bug in this repo.*

## Verification

- `cargo test -p ultros-app --lib` — **300 passed, 0 failed** (5 new)
- `cargo clippy -p ultros-app --all-targets -- -D warnings` — clean (exit 0)

Workspace-wide clippy was not run (a stale `cargo` process held the shared target lock, so this built against an isolated target dir); the change is confined to one file in `ultros-app`.

### DIFF
--- a/ultros-frontend/ultros-app/src/routes/scrip_sources.rs
+++ b/ultros-frontend/ultros-app/src/routes/scrip_sources.rs
@@ -21,7 +21,7 @@ use leptos_router::{
     NavigateOptions,
     hooks::{query_signal, use_navigate, use_query_map},
 };
-use std::{cmp::Reverse, sync::Arc};
+use std::{cmp::Reverse, collections::HashSet, sync::Arc};
 use ultros_api_types::cheapest_listings::{CheapestListings, CheapestListingsMap};
 use xiv_gen::{CollectablesShopRewardScripId, ItemId, Recipe};
 
@@ -104,6 +104,49 @@ impl std::fmt::Display for SortMode {
         };
         f.write_str(val)
     }
+}
+
+/// Maximum rows rendered by the table.
+const ROW_LIMIT: usize = 100;
+
+/// Rank the collected rows, collapse repeated items, and cap the list.
+///
+/// The ranking has to be a *total* order. Rows are collected by iterating
+/// `collectables_shop_items`, a `std::collections::HashMap`, so they arrive
+/// here in an order that `RandomState` randomizes per process. The SSR server
+/// and the hydrating wasm client each build their own copy of the game data,
+/// so ranking that leaves ties unresolved puts different rows in different
+/// places — and, at the `limit` boundary, drops a different *set* of rows
+/// entirely — on the two sides. That is the hydration-mismatch class fixed for
+/// the item page in #960. Tie-breaking on the stable item id pins one order.
+fn rank_scrip_sources(
+    mut results: Vec<ScripSourceData>,
+    sort_mode: SortMode,
+    limit: usize,
+) -> Vec<ScripSourceData> {
+    match sort_mode {
+        // `total_cmp` rather than `partial_cmp().unwrap()`: the unwrap was a
+        // latent panic if a cost ever produced a NaN ratio.
+        SortMode::CostPerScrip => results.sort_unstable_by(|a, b| {
+            a.cost_per_scrip
+                .total_cmp(&b.cost_per_scrip)
+                .then_with(|| a.item_id.0.cmp(&b.item_id.0))
+        }),
+        SortMode::ScripAmount => {
+            results.sort_unstable_by_key(|d| (Reverse(d.scrip_amount), d.item_id.0))
+        }
+        SortMode::Cost => results.sort_unstable_by_key(|d| (d.cost, d.item_id.0)),
+    }
+
+    // An item stocked by several collectables shops yields one row per shop.
+    // After a metric sort those rows are not adjacent, so the previous
+    // `dedup_by_key` — which only removes *consecutive* duplicates — left them
+    // on screen. Keep the first, i.e. best-ranked, row for each item.
+    let mut seen = HashSet::with_capacity(results.len());
+    results.retain(|r| seen.insert(r.item_id));
+
+    results.truncate(limit);
+    results
 }
 
 #[component]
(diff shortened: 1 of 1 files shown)


akarras:
**Both halves independently verified. Merging the fix; the zero-rows finding needs your call.**

## The fix — verified by execution

CI doesn't run `cargo test` (commented out in `rust.yml`), so green CI only proves this compiles. Ran the tests locally:

```
running 5 tests
test routes::scrip_sources::tests::dedup_keeps_the_best_ranked_row_for_an_item ... ok
test routes::scrip_sources::tests::repeated_items_collapse_even_when_not_adjacent ... ok
test routes::scrip_sources::tests::sort_modes_still_rank_by_their_metric ... ok
test routes::scrip_sources::tests::truncation_keeps_the_same_rows_regardless_of_input_order ... ok
test routes::scrip_sources::tests::ranking_is_independent_of_input_order ... ok
test result: ok. 5 passed; 0 failed; 295 filtered out
```

The diagnosis is right, and it's the same class as #960: a `std::collections::HashMap` whose `RandomState` iteration order reaches the DOM through the `<For>` key. Tie-breaking on `item_id` is the correct fix because it makes the order *total*, not merely stable.

Worth noting this closes two hazards I flagged when reviewing **#992** (which this supersedes): that its stable→unstable sort left ties breaking arbitrarily at the truncation boundary, and that the `partial_cmp().unwrap()` float comparator was a latent NaN panic. `total_cmp` handles the second properly.

The dedup change is also correct on Rust semantics alone — `dedup_by_key` removes only *consecutive* duplicates, and the list is sorted by a cost metric rather than by item id, so multi-shop items were never adjacent and never collapsed. The `HashSet` + `retain` first-wins form matches the existing comment's stated intent.

Confirmed this branch has #992 in its history and merges cleanly against current main.

## The zero-rows finding — confirmed, and I stopped where you did

Verified independently rather than taking it on trust:

```
$ awk -F, 'NR>3 {print }' csv/en/CollectablesShopRewardScrip.csv | sort -n | uniq -c
      1 0
    537 2
    119 4
     31 6
     13 7
```

...against a `from_id` matching `41784`/`41785`/`25199`/`33913`/`25200`/`33914`. Every row resolves to `Other(_)`, and both filter branches drop it. And live: `https://ultros.app/scrip-sources?world=Gilgamesh` returns 200 with 2.9MB and **zero** `role="row"`, `role="row-group"` and `<tbody>` elements. The page is dead exactly as described.

I looked for the mapping — `CurrencyScripConvert.csv` is the only obvious candidate and it's 4 rows of unnamed columns, so it isn't it. **I deliberately did not guess** which of `{2,4,6,7}` is White/Purple/Orange: shipping a wrong mapping would render visibly incorrect scrip types, which is worse than an obviously-empty page. You were right to split it out, and it needs a game-data judgement to close.

PR #999, opened 2026-07-29
🛡️ Sentry: [test coverage improvement] extract and test ConfidenceBadge logic

akarras:
🛡️ Sentry: [test coverage improvement] extract and test ConfidenceBadge logic

💡 What: Extracted the conditional formatting and styling logic for `ConfidenceBadge` in `tool_help.rs` into a pure function `get_tool_confidence_level` and an enum `ToolConfidenceLevel`.
🎯 Why: Component logic embedded directly in Leptos macros makes it untestable due to its tight coupling with the reactive context. Extracting it enables comprehensive unit testing.
📊 Impact: Adds full coverage for the threshold logic (`total_sales >= 20` and `daily_sales >= 1.0`) and styling outputs (`text-emerald-300`, `text-amber-300`, `text-red-300`) of the `ConfidenceBadge`.
🔬 Verification: Run `cargo test --package ultros-app --lib components::tool_help::tests`

---
*PR created automatically by Jules for task [14595038427896188832](https://jules.google.com/task/14595038427896188832) started by @akarras*

PR #1000, opened 2026-07-30
fix(scrip-sources): read Currency as an enum index so the page renders rows

akarras:
## The bug

`/scrip-sources` renders **zero data rows** in production, and has since it shipped. Live on `/pkg/c4df0cb/` (tip of `main`) as of this PR:

```
GET https://ultros.app/scrip-sources?world=Gilgamesh
  http=200  bytes=2948196
  role="columnheader" count: 5
  role="row-group"    count: 0
```

The headers render, so `ScripSourceTable` mounted and the cheapest-listings resource resolved fine — prices were available. Every row was lost in the scrip-type filter.

## Root cause

`ScripType::from_id` matched scrip **item ids** (41784 Orange Crafters', 33913 Purple Crafters', …), but it is called with `reward.currency` — and `CollectablesShopRewardScrip.Currency` is a small **enum index**. Every row in the pinned 7.55 data carries `0`, `2`, `4`, `6` or `7`, so `from_id` returned `Other(_)` for all of them, and the unfiltered branch dropped `Other` outright:

```rust
} else {
    // Default to showing Crafters scrips if no filter
    if matches!(scrip_type, ScripType::Other(_)) {
        continue;
    }
}
```

No row could pass under any filter. This is the same identifier-space mistake as #942, which keyed a game-data lookup on a localized `.name`; here it's an enum index read as an item id.

## Deriving the mapping from the game data

The mapping is derived, not guessed. `CollectablesShopRewardScrip.Currency` is reused by shops that pay out *items* rather than scrips, so the first step is to keep only the `CollectablesShop` rows with `RewardType = 1` (the `Lv. XX-YY` turn-in groups) and drop the material-exchange shops. Joining `CollectablesShopItem` across that filter:

| Currency | rows | turn-ins | → |
|---|---|---|---|
| 2 | 1089 | crafted, lv 50–99 | Purple Crafters' |
| 4 | 163 | gathered/fished, lv 50–98 | Purple Gatherers' |
| 6 | 93 | crafted, lv 78–80 **and lv 100** | Orange Crafters' |
| 7 | 18 | gathered/fished, lv 100 | Orange Gatherers' |

`2`/`4` are the purple (levelling) crafter/gatherer pair; `6`/`7` the orange (level 100) pair.

What pins currency `6` to Orange Crafters' rather than the retired white scrip: its level-100 rows are **exactly one item per crafting job**, all eight DoH jobs — the "Rarefied" max-level craft set.

```
Rarefied Claro Walnut Fishing Rod    -> Carpenter
Rarefied Ra'Kaznar Round Knife       -> Blacksmith
Rarefied Ra'Kaznar Ring              -> Armorer
Rarefied Black Star Earrings         -> Goldsmith
Rarefied Gargantuaskin Hat           -> Leatherworker
Rarefied Thunderyards Silk Culottes  -> Weaver
Rarefied Claro Walnut Flat Brush     -> Alchemist
Rarefied Tacos de Carne Asada        -> Culinarian
```

Currency `6`'s lv 78–80 rows are the Shadowbringers tier that collapsed into the same high-tier crafter slot when white scrips were removed in 7.0. No collectable awards a white scrip in the current data, which is why nothing maps to `WhiteCrafters`/`WhiteGatherers` any more.

## The fix

1. **`from_id` → `from_currency`**, matching the enum indices, with the derivation recorded in a doc comment so the next data bump can re-check it.
2. **Unrecognised currencies stay visible.** A stale mapping should cost a few oddly-labelled rows, not the entire page — one new expansion adding `Currency = 8` would blank it again. An unrecognised `?scrip=` value is likewise treated as "no filter" rather than emptying the table.
3. The six-branch `if s_filter == "..." && scrip_type != ...` chain collapses into `passes_scrip_filter`, which is what makes the bug unit-testable at all. Net −24 lines of logic.

## Verification

Genuine red → green, run locally:

- **RED** (mapping still on item ids, filter still dropping `Other`) — `cargo test -p ultros-app --lib scrip` exited **101**, 4 failed / 7 passed:
  - `currency 2 is unmapped, so its rows never render`
  - `currency_indices_map_to_the_right_scrip: left: Other(2), right: PurpleCrafters`
  - `item id 41784 was treated as a currency index`
  - `unrecognised currency dropped from the unfiltered list`
- **GREEN** — `cargo test -p ultros-app --lib` exited **0**: **307 passed, 0 failed** (full lib suite; was 300 before these 7 tests).
- `cargo fmt --all -- --check` → exit 0.
- `cargo clippy -p ultros-app --all-targets -- -D warnings` → exit 0. (This pulled in `ultros-db` / `leptos_axum`, so the `ssr` feature path type-checked too.)

Against the real pinned data, 1031 currency-2 rows and 87 currency-6 rows have recipes and will now reach the table (capped at `ROW_LIMIT = 100`).

No new user-facing strings — every `scrip_sources_*` key this touches already exists in all seven locales.

## Known gaps, deliberately not fixed here

- **Gatherer scrip rows still can't appear.** The cost model does `else { continue; }` for items without a recipe, and currencies 4 and 7 are entirely gathered/fished (0 of 181 rows have a recipe). Pricing a gathered turn-in means using the item's own market price rather than summed ingredients — a feature, not part of this fix. Until then the two Gatherers options in the filter are always empty.
- **The White options in the filter are dead** for the reason above: nothing in the current game data awards a white scrip. Left in place rather than removed as UI churn inside a bug fix.
- **"All Scrips" ranks purple and orange rows together** by gil-per-scrip, which is apples-to-oranges across two currencies. The scrip type is shown per row and the filter separates them, but a default that mixes tiers is worth revisiting.

### DIFF
--- a/ultros-frontend/ultros-app/src/routes/scrip_sources.rs
+++ b/ultros-frontend/ultros-app/src/routes/scrip_sources.rs
@@ -53,15 +53,49 @@ enum ScripType {
 }
 
 impl ScripType {
-    fn from_id(id: u32) -> Self {
-        match id {
-            41784 => ScripType::OrangeCrafters,
-            41785 => ScripType::OrangeGatherers,
-            25199 => ScripType::WhiteCrafters,
-            33913 => ScripType::PurpleCrafters,
-            25200 => ScripType::WhiteGatherers,
-            33914 => ScripType::PurpleGatherers,
-            _ => ScripType::Other(id),
+    /// Map a `CollectablesShopRewardScrip.Currency` value to the scrip it pays.
+    ///
+    /// `Currency` is a small **enum index**, not an item id — every row in the
+    /// 7.55 data carries `0`, `2`, `4`, `6` or `7`. Matching it against scrip
+    /// item ids is what left this page blank, so the mapping below is derived
+    /// from the game data instead. Joining `CollectablesShopItem` to the
+    /// `RewardType = 1` (scrip-paying) shops, ignoring the material-exchange
+    /// shops that reuse this column, gives:
+    ///
+    /// | Currency | rows | turn-ins |
+    /// |---|---|---|
+    /// | 2 | 1089 | crafted, lv 50-99 |
+    /// | 4 |  163 | gathered/fished, lv 50-98 |
+    /// | 6 |   93 | crafted, lv 78-80 and lv 100 |
+    /// | 7 |   18 | gathered/fished, lv 100 |
+    ///
+    /// So `2`/`4` are the purple (levelling) crafter/gatherer pair and `6`/`7`
+    /// the orange (level 100) pair. Currency `6`'s level-100 rows are exactly
+    /// one item per crafting job — the eight "Rarefied" max-level crafts — which
+    /// is what pins it to Orange Crafters' rather than the retired white scrip;
+    /// its lv 78-80 rows are the Shadowbringers tier that collapsed into the
+    /// same high-tier crafter slot when white scrips were removed in 7.0.
+    fn from_currency(currency: u32) -> Self {
+        match currency {
+            2 => ScripType::PurpleCrafters,
+            4 => ScripType::PurpleGatherers,
+            6 => ScripType::OrangeCrafters,
+            7 => ScripType::OrangeGatherers,
+            other => ScripType::Other(other),
+        }
+    }
+
+    /// The `?scrip=` query value that selects this type, as emitted by the
+    /// toolbar `<select>`.
+    fn from_filter_key(key: &str) -> Option<Self> {
+        match key {
+            "OrangeCrafters" => Some(ScripType::OrangeCrafters),
+            "OrangeGatherers" => Some(ScripType::OrangeGatherers),
+            "WhiteCrafters" => Some(ScripType::WhiteCrafters),
+            "PurpleCrafters" => Some(ScripType::PurpleCrafters),
+            "WhiteGatherers" => Some(ScripType::WhiteGatherers),
+            "PurpleGatherers" => Some(ScripType::PurpleGatherers),
+            _ => None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


akarras:
Merged. Verified in a real browser against a local SSR build of this branch: `/scrip-sources?world=Gilgamesh` now renders a populated table (headers + 19 row-groups, hydrated cleanly) where prod on `main` renders **0** rows. The currency-index fix is correct and this is a large net win.

One correction to the PR's "Known gaps" section, found while verifying — **the Gatherers filters are not empty; they are populated entirely by mislabelled rows.**

`ScripType::from_currency` is right, but the *runtime* loop in `ScripSourceTable` iterates every `data.collectables_shop_items` entry and never applies the `CollectablesShop.RewardType == 1` filter that the derivation in the doc comment used. `Currency` is reused by the material-exchange shops, so those rows are also rendered and labelled with a scrip they don't award.

Joining the pinned 7.55 data (`CollectablesShop` → `CollectablesShopItem` → `CollectablesShopRewardScrip`, counting only rows that have a recipe, i.e. the ones the cost model can actually render):

| currency | RewardType | renderable rows | what they are |
|---|---|---|---|
| 2 | 1 (scrip shop) | 1031 | genuine Purple Crafters' ✅ |
| 2 | 2 (material exchange) | 36 | mislabelled ❌ |
| 4 | 1 (scrip shop) | **0** | gatherer turn-ins, no recipes — the stated gap ✅ |
| 4 | 2 (material exchange) | **59** | mislabelled ❌ |
| 6 | 1 | 87 | genuine Orange Crafters' ✅ |
| 7 | 1 | 0 | orange gatherer, no recipes ✅ |

So `?scrip=PurpleGatherers` returns ~59 rows that are **100% material-exchange items**, not gatherer turn-ins. Confirmed the exact rows in the browser: `Resplendent Alchemist's Component B`, `Resplendent Carpenter's Component C` (shop: *Resplendent Materials Exchange*) and `Oddly Specific Cedar Plank` (shop: *Oddly Specific Materials Exchange (Crafting)*) — all `Currency=4, RewardType=2`. `?scrip=PurpleCrafters` is ~1067 rows of which 36 (3.4%) are similarly spurious.

Not a blocker and not a regression — the page was 100% dead before, and ~92% of what it now shows is correct — but the doc comment's claim that the Gatherers options "are always empty" is inverted in practice.

**Why I didn't just fix it here:** the proper fix is to restrict to `RewardType == 1`, and `xiv-gen` does not load `CollectablesShop.csv` at all (no `collectables_shops` field on `Data`). That needs a new sheet in `xiv-gen/src/lib.rs` + `csv_to_rkyv.rs`, which changes the rkyv payload shipped to every client — too wide to graft onto a bug fix at review time. Filed as follow-up.

Verification run on the merge of this branch with #999/#1013/#1001/#1015/#1014: `./check_ci.sh` exit 0, `cargo test -p ultros-app --lib` 330 passed / 0 failed.

PR #1013, opened 2026-07-30
fix(clickhouse): thread real Postgres sale ids into the dual-write

akarras:
## Problem

The live sale-ingest path wrote every ClickHouse row with `pg_id = 0`, which breaks
`sales` dedup in both directions.

`update_sales` / `insert_unrecorded_sales` (`ultros-db/src/sales.rs`) built its return
values from a hand-rolled `Model { id: 0, .. }` and then inserted with
`.exec_without_returning()`, so the real Postgres ids never came back. Those return
values are exactly what both live producers (`ultros/src/main.rs` websocket handler and
`ultros/src/item_update_service.rs` catch-up) publish on the sale event bus, and what the
analyzer's dual-write feeds to `SaleRow::from_api_sale` — which sets `pg_id: s.id`.

`sales` is `ReplacingMergeTree(inserted_at) ORDER BY (item_id, hq, world_id, sold_date, pg_id)`,
so `pg_id` is the only discriminator between two sales of the same item/hq/world in the
same second. Consequences:

- **(a) Live path loses sales.** Two distinct same-second sales of the same
  `(item, hq, world)` share the whole sort key when `pg_id` is constant `0`, and collapse
  to one row on merge.
- **(b) Backfill double-counts.** `backfill.rs` goes through `SaleRow::from_db_model`,
  which carries the *real* `sale_history.id`. Same logical sale, different sort key →
  ReplacingMergeTree never merges the two, so every sale the live path already wrote gets
  a second copy.

Every rollup reads `FROM sales FINAL` (`item_stats_window`, `sales_hourly`,
`world_kpi_5min`, and `item_quality_score` derived from the first), so all of them inherit
both errors.

## Changes

1. **`ultros-db/src/sales.rs`** — `insert_unrecorded_sales` now uses
   `InsertMany::exec_with_returning` (SeaORM 2.0; Postgres supports `INSERT ... RETURNING`)
   and builds the returned `SaleHistory` values from the rows Postgres actually inserted.
   Buyers are re-attached by `buying_character_id` via a new `attach_buyers` helper rather
   than zipped positionally, because Postgres does not promise RETURNING row order matches
   the VALUES order.
2. **Analyzer path confirmed end-to-end.** Both producers of `SaleEventData`
   (`main.rs:157`, `item_update_service.rs:216`) forward the `update_sales` return value
   unchanged, and `analyzer_service.rs:629` calls `SaleRow::from_api_sale` on each element,
   so real ids now reach `pg_id`. No changes needed there.
3. **Comment fixes.** `schema.rs`, `backfill.rs`, and the `SaleRow::pg_id` doc all claimed
   the idempotency key included `buying_character_id`, which is not in the `ORDER BY`. They
   now describe the actual key and spell out why a placeholder `pg_id` is corrupting.
4. **Tests** — see below.

No schema changes (ClickHouse is Tier 1).

## Tests

`cargo test -p ultros-db -p ultros-clickhouse --lib`

- `ultros-db`: `attach_buyers_preserves_the_postgres_ids` (direct regression guard for the
  old `id: 0`), `attach_buyers_keeps_same_second_sales_distinct`,
  `attach_buyers_matches_by_id_not_position`, `attach_buyers_drops_rows_with_no_known_buyer`.
- `ultros-clickhouse`: `freshly_recorded_sale_carries_a_nonzero_pg_id` — runs a recorded
  `sale_history::Model` through the real dual-write conversion chain
  (`SaleHistoryReturn` → `SaleHistory` → `SaleRow::from_api_sale`) and asserts a nonzero,
  correct `pg_id`; plus `live_and_backfill_paths_agree_on_pg_id`, which asserts
  `from_api_sale` and `from_db_model` produce an identical sort key for the same PG row.

**Not covered:** no Postgres was available in this environment, so `exec_with_returning`
itself was not exercised against a live database. Worth one manual smoke run against dev
before the remediation below (`clickhouse_parity_check` is the fastest check — see step 5).

## Remediation plan for existing data — DO NOT RUN AGAINST PROD WITHOUT THE OWNER

**This fix must land and be deployed before the planned full one-shot backfill.** The prod
backfill has so far covered only ~7% of items; running the complete one against the current
data would double-count every sale the live path has already written this month.

Existing prod `sales` contains two kinds of damage:

- `pg_id = 0` rows written by the live path — some fraction of them silently merged, so the
  rows are both mislabelled *and* undercounted. They cannot be repaired in place; they have
  to be deleted and re-derived from Postgres.
- For any `(world, month)` the backfill already covered, a duplicate real-`pg_id` copy of
  the same sales. Deleting the `pg_id = 0` rows resolves this: the surviving real-id rows
  are the correct ones.

### Steps

1. **Snapshot the blast radius** (read-only; run first, keep the output):
   ```sql
   SELECT toYYYYMM(sold_date) AS ym, world_id, count() AS bad_rows
   FROM sales WHERE pg_id = 0
   GROUP BY ym, world_id ORDER BY ym, world_id;
   ```
   The `(world_id, ym)` tuples this returns are exactly the chunks that need re-backfilling
   in step 3.

2. **Delete the placeholder rows.** `ALTER TABLE ... DELETE` is a mutation — schedule it
   off-peak and let it finish before step 3:
   ```sql
   ALTER TABLE sales DELETE WHERE pg_id = 0;
   ```
   Watch it with `SELECT * FROM system.mutations WHERE table = 'sales' AND NOT is_done`.
   (`DELETE FROM sales WHERE pg_id = 0` — lightweight delete — is an option if the volume
   is large, but the mutation is simpler to reason about and the row count here should be
   modest.)

3. **Clear `_backfill_state` for the affected chunks** so `backfill_sales` will rewrite them
   instead of skipping. `_backfill_state` is `ReplacingMergeTree(completed_at)`, so a delete
   is required — inserting a "not done" marker is not expressive in that schema:
   ```sql
   ALTER TABLE _backfill_state DELETE
   WHERE (world_id, year_month) IN (
     -- the tuples from step 1
     (40, 202607), ...
   );
   ```
   If the affected set turns out to be "everything", `TRUNCATE TABLE _backfill_state` is
   equivalent and cheaper. Verify with
   `SELECT count() FROM _backfill_state FINAL WHERE ...` before moving on.

4. **Re-run the backfill** for the cleared range. This is the same run the owner already
   planned, now safe:
   ```bash
   cargo run --release --bin clickhouse_backfill 2022
   ```
   Chunks outside the cleared set are skipped, so the incremental cost is bounded by the
   damage found in step 1.

5. **Verify parity** before touching rollups:
   ```bash
   cargo run --release --bin clickhouse_parity_check
   ```
   Exits nonzero on mismatch. Any remaining `(world, ym)` mismatch means a chunk was missed
   in step 3.

6. **Rebuild the rollups.** They all read `FROM sales FINAL`, so they self-heal once `sales`
   is correct — but only over their trailing windows (`world_kpi_5min` 50h, `sales_hourly`
   30h, `item_stats_window` 1/7/30/90d). Historical buckets outside those windows are
   immutable and keep the bad numbers until explicitly recomputed. Either accept that for
   pre-window history, or truncate and rebuild:
   ```sql
   TRUNCATE TABLE item_stats_window;
   TRUNCATE TABLE item_quality_score;
   TRUNCATE TABLE sales_hourly;
   TRUNCATE TABLE world_kpi_5min;
   ```
   then a `rollups::refresh_all` (a web-server restart does the seed run) — noting that this
   only repopulates the trailing windows, so `sales_hourly` / `world_kpi_5min` history older
   than 30h/50h would need a widened one-off backfill query if the owner wants it restored.
   Recommendation: **don't truncate `sales_hourly` / `world_kpi_5min`** unless the owner
   wants to write that one-off; the bad history is bounded and the trailing windows correct
   themselves within a refresh cycle. `item_stats_window` and `item_quality_score` are fully
   recomputed from `sales` on every refresh, so those two are safe to truncate.

### Ordering constraint

Steps 2–4 must happen **after** this PR is deployed. If the live path is still writing
`pg_id = 0` while the backfill runs, the cleanup re-creates the exact problem it fixed.

### DIFF
--- a/ultros-clickhouse/src/backfill.rs
+++ b/ultros-clickhouse/src/backfill.rs
@@ -5,9 +5,12 @@
 //! in the `_backfill_state` table and skipped.
 //!
 //! Idempotent: the `sales` table is a `ReplacingMergeTree` keyed on
-//! `(item_id, hq, world_id, sold_date, buying_character_id)`, so re-streaming
-//! the same rows just merges into no-ops. That means dual-writes overlapping
-//! the backfill window are safe too.
+//! `(item_id, hq, world_id, sold_date, pg_id)`, so re-streaming the same rows
+//! just merges into no-ops. Dual-writes overlapping the backfill window are
+//! safe *because* both paths key on the same Postgres `sale_history.id` — see
+//! [`SaleRow::from_db_model`] (this path) and [`SaleRow::from_api_sale`] (the
+//! live path). If either ever writes a placeholder id, the two paths stop
+//! agreeing on row identity and the overlap double-counts instead of merging.
 
 use std::time::Instant;
 
--- a/ultros-clickhouse/src/rows.rs
+++ b/ultros-clickhouse/src/rows.rs
@@ -21,10 +21,13 @@ use serde::{Deserialize, Serialize};
 #[derive(Row, Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct SaleRow {
     /// Carries the Postgres `sale_history.id` so each PG row maps to a
-    /// distinct CH row. Without this, two sales sharing
-    /// `(item_id, hq, world_id, sold_date, buying_character_id)` collapse
-    /// under the `ReplacingMergeTree` dedup — and Universalis-side replays
-    /// produce a non-trivial number of those.
+    /// distinct CH row. It is the last component of the `sales` ORDER BY key,
+    /// so without a real value two sales sharing
+    /// `(item_id, hq, world_id, sold_date)` collapse under the
+    /// `ReplacingMergeTree` dedup — and Universalis-side replays produce a
+    /// non-trivial number of those. Every producer must supply the real id:
+    /// a placeholder makes the live path and the backfill disagree on row
+    /// identity, so the same sale is stored twice and never merges.
     pub pg_id: i32,
     #[serde(with = "clickhouse::serde::chrono::datetime")]
     pub sold_date: chrono::DateTime<chrono::Utc>,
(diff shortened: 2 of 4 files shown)


akarras:
## Review: LGTM — correct fix, verified end-to-end. Merge when ready, but read the remediation notes before running the cleanup.

Verified (against sea-orm 2.0.0 as pinned in Cargo.lock): `InsertMany::exec_with_returning` exists, emits `RETURNING` on Postgres, and returns exactly the inserted rows — no conflict clause means no partial-insert path. Positional-mapping fragility is genuinely gone: every field of the returned `SaleHistory` comes from the RETURNING row, and buyer names join on `buying_character_id` (a returned column), so ordering can't mis-map. Traced `update_sales` → `SaleEventData` (main.rs:157, item_update_service.rs:216) → `SaleRow::from_api_sale` → `pg_id`: no remaining zeroing point. Comment fixes in schema.rs/backfill.rs/rows.rs are accurate against the real DDL.

**Bonus fix worth adding to the PR body**: `ultros-api-types/src/lib.rs:101` `upsert_sales` retains on `sale.id != incoming.id` — with all live ids = 0, every incoming sale evicted the previous one, so the live sale feed could only ever hold ONE websocket-delivered sale. This PR fixes that as a side effect.

### ⚠️ Before running the remediation (the important part)

1. **The pg_id=0 collapse was accidentally hiding Universalis replay duplicates.** The pre-insert dedup window in `insert_unrecorded_sales` is only `limit = sales.len()` recent rows (and ignores `price_per_item`), so replayed batches can produce duplicate Postgres rows. Those used to merge away in CH (identical key); with real ids they become permanent double-counts — and `clickhouse_parity_check` will happily report parity on them (both sides count the same duplicated PG rows). Before the cleanup run: measure `SELECT count() - uniqExact((sold_item_id, hq, world_id, sold_date, buying_character_id, quantity)) FROM sale_history` to quantify it, and consider widening the dedup window / a unique natural-key index as a follow-up.
2. **Restrict the `ALTER TABLE sales DELETE WHERE pg_id = 0` by the affected `toYYYYMM` partitions** from step 1 — the table is partitioned by month and this materially shortens the mutation.
3. **Hard-gate the cleanup on one real smoke test**: one live sale observed with nonzero `pg_id` in CH after deploy. If `exec_with_returning` ever errored at runtime (e.g. `support_returning()` false → `DbErr::BackendNotSupported`), the cleanup would delete old rows while ingest writes nothing.
4. Expect a visible dip in analyzer history between the delete and backfill completion — worth timing accordingly.

### Test notes (non-blocking)
The new tests assert the conversion chain, which was never broken — the bug lived in the two deleted lines (`Model { id: 0 }` + `exec_without_returning`), and nothing exercises `insert_unrecorded_sales` itself. sea-orm's `mock` feature (not currently enabled) would allow asserting the statement is a RETURNING insert — cheap follow-up. Also remember CI has `cargo test` commented out, so these tests only run locally.

Minor: `rows.rs:170` has a redundant `recorded.clone()`; `attach_buyers_matches_by_id_not_position` oversells its name (it tests buyer lookup, not sale-id mapping — which is now safe by construction).

PR #1020, opened 2026-07-30
feat(ingest): instrument and harden the five silent market-data failure modes

akarras:
Market-data ingest could fail in at least five ways that were invisible from outside the process: the app stays up, requests succeed, and the numbers simply stop moving. Serving stale data is not an error, so nothing in the request path could notice — and two of these modes stopped ingestion *permanently* while the process looked perfectly healthy.

No happy-path behavior changes.

## 1. Broadcast lag was swallowed

Both analyzer loops used `if let Ok(..) = recv()`, which treats `RecvError::Lagged(n)` exactly like a value that happened not to match. A burst larger than the ring vanished with no log, no metric, and no way to tell from outside — Postgres stayed correct, so the in-RAM caches and the ClickHouse dual-write just quietly drifted.

They now go through `handle_bus_recv` ([event.rs](../blob/claude/ultros-ingest-failures-7d60b2/ultros/src/event.rs)), which matches `Lagged` explicitly, increments `ultros_analyzer_bus_lagged_total{bus}` by the dropped count, warns, and continues.

The rings were also sized below one real burst. `UpdateService::check_items` walks item ids in chunks of 100 through `buffer_unordered(50)`, and each item emits two listing events plus one sale event — ~200 listing and ~100 sale events per chunk, against rings of 100 and 40. Any full sweep was *guaranteed* to lag.

- **listings: 100 → 1024** (~5 chunks of headroom; slots hold `Arc<ListingEventData>` carrying a handful of listings each, so single-digit MiB when full)
- **history: 40 → 512** (~5 chunks; `SaleEventData` has no retainer payload so slots are cheaper)

Both numbers are justified in doc comments at the constants.

## 2. No staleness gauge

`listing_last_updated` already stamps an ingest time on every update, so `now - max(date_time)` per world is a ready-made "seconds since we last heard from this world" signal. New [`ingest_health.rs`](../blob/claude/ultros-ingest-failures-7d60b2/ultros/src/ingest_health.rs) publishes it every 60s as `ultros_world_ingest_staleness_seconds{world}`, plus `ultros_worlds_never_ingested` for worlds that would otherwise have no series to alert on at all.

This is the metric to page on; everything else explains why it moved.

## 3. Websocket had no pong deadline

A half-open TCP connection yields no frames, never errors, and leaves `is_terminated()` false — so `websocket.next()` blocks forever, the existing reconnect path is never reached, and the only fix was a process restart. Pongs were logged and discarded.

The worker now tracks the last inbound frame (any frame — Pong, data, Ping, Close) and tears the socket down after **150s** of silence, reconnecting through the existing path. That's 2.5× the 60s ping interval, so one slow round-trip is tolerated and two consecutive missed pongs trip it. Override with `UNIVERSALIS_WEBSOCKET_LIVENESS_TIMEOUT_SECS`. Counted by `universalis_websocket_liveness_timeouts_total`.

This adds the `metrics` facade crate to `universalis` — it no-ops without a recorder, matching how `ultros-db` already reports from library code.

## 4. Snapshot age

Restoring a snapshot skips the Postgres reload entirely (`run_worker`), so whatever prices it holds become "live" the moment `initiated` flips — with no visible symptom. A week-old snapshot served week-old prices as live.

Snapshots older than **3 hours** are now rejected and fall through to the database load. Snapshots are written every 15 minutes and on shutdown, so any larger gap means the process was down and the market has moved on; 3h keeps a fast restart cheap (deploy, crash loop, OOM restart) while capping how wrong the served data can be. Names that don't parse are also rejected — "we don't know how old this is" is a reason to reload, not a reason to serve.

Age comes off the filename rather than mtime, so a file copy can't make a stale snapshot look fresh. Counted by `ultros_analyzer_snapshot_rejected_total{reason}`; the age booted from is reported as `ultros_analyzer_snapshot_age_seconds`.

## 5. Panics in hot loops

All converted to log-and-skip with `ultros_analyzer_skipped_events_total{op,reason}`.

The `.expect("Unknown world")` in `add_sale` was the worst of them: it ran inside the history loop, so one unknown world panicked the spawned task and **permanently stopped live sale ingestion and the ClickHouse dual-write** while axum kept serving from a frozen cache. Dropping one event costs one sale; panicking costs every sale from then on.

Sites converted: `add_sale`, `add_listings` (region + world), `remove_listings` (region + world), `CheapestListings::remove_listing`. In `add_listings` both required entries are now resolved before either is written, so a missing world can't leave the region map holding a price the world map never learned about. In `remove_listings` a missing region no longer skips the datacenter and world removals — leaving a sold listing in place is exactly what makes prices read as stale.

**Beyond the listed sites:** the four unwraps in the `run_worker` database reload (`~568-598`) fail the same way for the same reason and abort the worker *before* `initiated` is ever set, so the analyzer would both serve nothing and ingest nothing, forever. Converted those too.

### Root cause, out of scope

`WorldCache` is built exactly once in `main.rs` and never refreshes, and on a cold database that build races the task populating the world table (per the existing comment at the call site). A world it misses is absent from the analyzer's maps *for the lifetime of the process*, and every event for it is dropped.

**A refresh path is deliberately not in this PR.** Until one exists the workaround is a restart once the world table is populated — but the condition is now visible as `ultros_analyzer_skipped_events_total{reason="unknown_world"}` climbing steadily, instead of a panic that kills ingestion outright. Documented in `docs/ingest-observability.md`.

## Testing

`./check_ci.sh` passes (fmt + clippy clean). Tests run locally since CI doesn't run them:

- **ultros**: 163 passed, 1 failed
- **universalis + ultros-db**: 38 passed, 0 failed

New tests: bus lag handling over a real broadcast channel (burst larger than the ring reports lag once, then resumes and keeps delivering), snapshot filename age parsing including unparseable names, and snapshot age rejection wired through `try_restore_from_snapshot` with an injected `now`.

### Two pre-existing failures, both confirmed against an unmodified `main`

I stashed this branch, rebuilt, and ran both against `c4df0cb7` before touching either.

1. **`analyzer_service::tests::test_persistence` deadlocked** — read guards on `new_recent_sale_history` were held across a later restore that write-locks the same map, so everything past the first restore assertion was unreachable. **Fixed here** (scoped the guards), because my new snapshot-age assertions live in that test and couldn't otherwise run. It now passes in ~5s.

2. **`web::api::push::tests::label_for_safari_on_ios` fails** (`got Browser (Safari on macOS)`, `push.rs:155`). Unrelated to ingest — **left alone**, flagging it here so it isn't mistaken for fallout from this change.

## Docs

New `docs/ingest-observability.md`: metric reference, suggested alert expressions, the `WorldCache` gap, and the tunable env vars and constants.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10302,6 +10302,7 @@ dependencies = [
  "chrono",
  "clap",
  "futures",
+ "metrics",
  "reqwest 0.11.27",
  "serde",
  "serde_json",
--- a/docs/ingest-observability.md
+++ b/docs/ingest-observability.md
@@ -1,0 +1,68 @@
+# Ingest observability
+
+Market-data ingest fails silently by default. The app keeps serving, requests
+keep returning 200, and the numbers simply stop moving — serving stale data is
+not an error, so nothing in the request path can notice. These metrics exist to
+turn each of those silent modes into something alertable. They are exported on
+the Prometheus endpoint at `:9091/metrics` (see `ultros/src/web_metrics.rs`).
+
+## Metrics
+
+| Metric | Type | Labels | What it means |
+| --- | --- | --- | --- |
+| `ultros_world_ingest_staleness_seconds` | gauge | `world` | Seconds since the newest `listing_last_updated` row for that world. The single best "is ingest alive" signal. |
+| `ultros_worlds_never_ingested` | gauge | — | Worlds in `WorldCache` with no `listing_last_updated` rows at all. |
+| `ultros_analyzer_bus_lagged_total` | counter | `bus` | Events the analyzer's broadcast receiver was too slow to read, and the channel discarded. Nonzero means the in-RAM caches and ClickHouse have drifted from Postgres. |
+| `ultros_analyzer_skipped_events_total` | counter | `op`, `reason` | Listing/sale events dropped because a world, datacenter or region was missing from the analyzer's maps. |
+| `ultros_analyzer_snapshot_rejected_total` | counter | `reason` | Startup snapshots refused (`too_old`, `unparseable_name`), causing a fall back to the Postgres reload. |
+| `ultros_analyzer_snapshot_age_seconds` | gauge | — | Age of the snapshot this process booted from. Set once at startup. |
+| `universalis_websocket_liveness_timeouts_total` | counter | — | Websocket connections torn down for delivering no frames within the liveness deadline. |
+
+Pre-existing and still useful alongside these:
+`ultros_websocket_rx{WorldId}`, `ultros_catchup_items_recovered{world}`,
+`ultros_catchup_window_saturated{world}`.
+
+## Suggested alerts
+
+`ultros_world_ingest_staleness_seconds` is the one to page on. Everything else
+explains *why* it went up.
+
+- **A world went silent** — `max_over_time(ultros_world_ingest_staleness_seconds[15m]) > 3600`.
+  Quiet worlds do go an hour without a listing change; a whole datacenter going
+  quiet at once does not.
+- **Every world went silent** — the same expression firing across most series at
+  once points at the websocket or the process, not the market. Correlate with
+  `universalis_websocket_liveness_timeouts_total`.
+- **Silent data loss** — `increase(ultros_analyzer_bus_lagged_total[1h]) > 0` or
+  `increase(ultros_analyzer_skipped_events_total[1h]) > 0`. Neither should ever
+  be nonzero in steady state. Sustained `bus_lagged` means the ring sizes in
+  `ultros/src/event.rs` need revisiting; sustained `skipped_events` with
+  `reason="unknown_world"` means `WorldCache` is missing a world (see below).
+- **Booted on stale data** — `increase(ultros_analyzer_snapshot_rejected_total{reason="too_old"}[1h]) > 0`
+  is informational: the guard did its job, but the process was down long enough
+  to matter.
+
+## Known gap: `WorldCache` never refreshes
+
+`WorldCache` is built exactly once, in `main.rs`, and on a cold database that
+build races the task that populates the world table. A world it misses is absent
+from the analyzer's maps for the lifetime of the process, and every event for
+that world is dropped — visible as
+`ultros_analyzer_skipped_events_total{reason="unknown_world"}` climbing steadily.
+
+Until a refresh path exists, the fix is a restart once the world table is
+populated. Before this change these lookups panicked instead, which killed live
+sale ingestion and the ClickHouse dual-write outright while the process kept
+serving.
+
+## Tuning
+
+| Env var | Default | Effect |
+| --- | --- | --- |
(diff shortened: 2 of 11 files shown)


akarras:
## Review + fixes pushed (7be93ce8)

Deep review verdict on the original head: design sound, all five sub-items genuinely implemented (verified: no missed `.expect()` sites — the sweep actually caught 9 hot-path sites, one more than the issue listed; Lagged handling correct with the receiver still positioned at the oldest retained message; reconnect preserves world subscriptions via `resend_subscriptions`; snapshot timestamp correctly taken from the filename rather than fragile mtime). Three MAJORs needed fixing, now addressed:

1. **Snapshot metrics fired before the recorder existed.** `ultros_analyzer_snapshot_rejected_total` / `_age_seconds` were emitted during analyzer startup, but the Prometheus recorder was installed inside `start_metrics_server` → `start_web`, spawned after `start_analyzer` — so the samples hit the NoopRecorder and the exact alert docs/ingest-observability.md recommends could never fire (rejection is an immediate return, so the loss was guaranteed, not a race). The recorder is now installed in main right after tracing init, before any service spawns, with the handle threaded into `start_web` for the /metrics route.
2. **The liveness path awaited `ws.close(None)` on the socket it had just declared half-open** — async-tungstenite's close() blocks on flush, which on a vanished peer sits in kernel retransmit for ~13-30 min (tcp_retries2), stalling the single-loop worker and reintroducing the exact silent stall being fixed. Now `drop(ws)` with a comment; the close() calls on send-error paths were deliberately kept (peer proven writable-broken there — different case).
3. **`handle_bus_recv` collapsed Lagged and Closed into one `None`** with a "keep looping on None" doc — on Closed that's an immediate-return hot spin emitting one warn per iteration into the tracing→GlitchTip pipeline. Now a three-state `BusRecv { Msg, Lagged, Closed }`; both analyzer loops break on Closed (cancellation arms kept), plus a new drain-then-closed-forever test.

Also fixed: `MissedTickBehavior::Delay` on the staleness interval (Burst turns one slow query into a storm), `metrics` made optional behind the `websocket` feature in universalis, metric renamed to `ultros_websocket_liveness_timeouts_total` (repo convention), negative/future-dated snapshot ages now rejected with reason `future_dated` + tests, and the docs now say plainly that bus-lag coverage is **analyzer-only** (the five alert/websocket bus consumers still swallow Lagged — deliberate scope, good follow-up candidate).

One pre-existing red fixed in passing: `cargo test -p universalis` failed before these changes — the `update_subscription` doctest dials the live Universalis websocket and panics on the rustls CryptoProvider; now `no_run` with a note.

Left as-is per review judgment: the bus capacity bumps (100→1024 / 40→512). They're reasoned in the comments, but they affect five other consumers and partly pre-empt the new lag metric — your call whether to keep them or ship the metric first and size from data.

Gates: fmt-check 0; `clippy -p ultros -p universalis -p ultros-db --all-targets -- -D warnings` 0; `cargo test -p universalis` 18+1 pass. (The `ultros` test binary doesn't link on Windows/MSVC — clippy `--all-targets` is its test-compile gate, per repo convention.)

From the review, worth knowing: this PR also fixes a real deadlock in the pre-existing `test_persistence` (a read guard held across a write-lock acquisition on a current_thread runtime) — Part 2 of that test had never actually executed before this branch.

PR #1021, opened 2026-07-30
Add BreadcrumbList JSON-LD to item pages

### LINKED ISSUE #1004: Add BreadcrumbList JSON-LD to item pages (evaluate Product markup cautiously)
_From the July 2026 improvement audit (task 12). Coordination: touches `item_view.rs` — may conflict with the canonical fix (task 2) and freshness-badge fix (task 10) if in flight._

Ultros item pages (`https://ultros.app/item/{world-or-region}/{id}`, ~20k items, the site's main SEO surface) emit zero structured data. Add JSON-LD.

Context: the ONLY JSON-LD in the codebase is the home page's WebSite/SoftwareApplication graph (`routes/home_page.rs` ~33-146). NOTE: there is a known bug where that script serves as empty `{}` in production — if that fix hasn't landed yet, coordinate: whatever mechanism you use here must be verified against SERVED SSR HTML via curl, not the hydrated DOM.

1. **BreadcrumbList (safe, do it)**: item pages already render the hierarchy (item → category link, `item_view.rs` ~1784-1791). Emit BreadcrumbList JSON-LD: Home → Item Explorer → {category} → {item}. Names/urls must match the visible links (category URLs need the same encoding the visible link uses). This must be rendered server-side with the item's data — the item name and category come from the in-process `xiv_gen_db` so they're available during SSR.
2. **Product/AggregateOffer (evaluate FIRST, then decide)**: the data exists (lowest price, listing count) but Google's Product rich-result policies target real purchasable products with real currency; gil is a fictional currency and "GIL" is not a valid ISO 4217 code. Marking prices up as offers risks spam/manual-action problems rather than rich results. Research current Google guidance on virtual items before implementing; the likely right answer is BreadcrumbList only, plus maybe a plain Thing/CreativeWork description — write your conclusion in the PR either way. Do NOT ship Product markup with `priceCurrency: "GIL"` without documenting why it's safe.
3. **Localize correctly**: item names in SSR come from the server's game-data language (known repo gotcha: the server renders English regardless of locale cookie). Use whatever the SSR page actually renders so the markup matches visible content — matching beats localization here.
4. The i18n rule applies only to user-visible strings; JSON-LD literals like "Item Explorer" should match the site's rendered breadcrumb text.

Verify: curl a local item page, extract the JSON-LD, run it through a schema validator. Confirm the script body survives into served HTML (see the empty-`{}` bug above). `./check_ci.sh` before committing.

akarras:
This change adds server-side rendered BreadcrumbList JSON-LD structured data to FFXIV item pages on ultros.app, improving search engine discovery and optimization (SEO).

Evaluation of Product/AggregateOffer markup:
Google requires real ISO 4217 currencies for Product rich results. Because FFXIV gil is a virtual currency and 'GIL' is not a valid ISO 4217 code, adding Product markup with fictional currencies risks structured data manual actions or spam flags. Thus, we purposely only added the safe BreadcrumbList JSON-LD and documented this architectural decision in comments.

Implementation details:
1. Defined a pure, unit-tested function 'build_breadcrumb_json_ld' that outputs BreadcrumbList JSON-LD as a serialized string.
2. Injected it via an unescaped HTML '<script type="application/ld+json">' inside the 'ItemView' component using 'inner_html' to avoid the empty-{} leptos_meta::Script SSR bug.
3. Matches URL hierarchy and text of visible links exactly (e.g. category encoding 's.name.replace("/", "%2F")' is preserved, and missing categories are cleanly omitted from the breadcrumb list).
4. Added comprehensive unit tests covering both category-included and category-omitted cases.

Fixes #1004

---
*PR created automatically by Jules for task [4252118697297993143](https://jules.google.com/task/4252118697297993143) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/routes/item_view.rs
+++ b/ultros-frontend/ultros-app/src/routes/item_view.rs
@@ -1671,6 +1671,65 @@ fn DiscordCommandChip(
     }
 }
 
+/// Builds the item page's `BreadcrumbList` JSON-LD.
+///
+/// `category` is `(display_name, search_category_id)`. The id — not the
+/// category's localized name — is what the URL is keyed on: #1001 moved
+/// `/items/category/:category` to an id precisely because the name differs per
+/// locale, so a name-keyed URL here would hand Google a link that doesn't
+/// resolve. This must keep matching the visible category link in the view below.
+fn build_breadcrumb_json_ld(
+    item_name: &str,
+    world_val: &str,
+    item_id_val: i32,
+    category: Option<(&str, i32)>,
+) -> String {
+    let mut items = vec![
+        serde_json::json!({
+            "@type": "ListItem",
+            "position": 1,
+            "name": "Home",
+            "item": "https://ultros.app/"
+        }),
+        serde_json::json!({
+            "@type": "ListItem",
+            "position": 2,
+            "name": "Item Explorer",
+            "item": "https://ultros.app/items"
+        }),
+    ];
+
+    if let Some((c_name, category_id)) = category {
+        items.push(serde_json::json!({
+            "@type": "ListItem",
+            "position": 3,
+            "name": c_name,
+            "item": format!("https://ultros.app/items/category/{category_id}")
+        }));
+        items.push(serde_json::json!({
+            "@type": "ListItem",
+            "position": 4,
+            "name": item_name,
+            "item": format!("https://ultros.app/item/{world_val}/{item_id_val}")
+        }));
+    } else {
+        items.push(serde_json::json!({
+            "@type": "ListItem",
+            "position": 3,
+            "name": item_name,
+            "item": format!("https://ultros.app/item/{world_val}/{item_id_val}")
+        }));
+    }
+
+    let json_value = serde_json::json!({
+        "@context": "https://schema.org",
+        "@type": "BreadcrumbList",
+        "itemListElement": items
+    });
+
+    serde_json::to_string(&json_value).unwrap_or_default()
+}
+
 /// Gates the item page on the `:id` route param actually naming a real item.
 /// A param that fails to parse, or parses to an id with no matching item,
 /// previously fell through to `unwrap_or_default()` and silently rendered an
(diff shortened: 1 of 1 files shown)


akarras:
## Review: request changes — the rendering mechanism is right, the URLs in the breadcrumbs are wrong.

What's right (and important): using a plain `<script type="application/ld+json" inner_html=json_ld />` bypasses the leptos_meta head-registration path where the home page's empty-`{}` bug lives — verified against tachys 0.2.18 source that inner_html closures are invoked during SSR and skipped on hydrate (no wipe, no cursor drift). serde_json means quotes/apostrophes/non-ASCII in item names are escaped correctly. The Product/Offer decision (not shipped, documented, gil ≠ ISO 4217) is exactly what the issue asked for. Schema shape is valid.

### Blocking

1. **The final breadcrumb node contradicts the page's own canonical and is non-deterministic on the URL Google actually crawls.** The last ListItem emits `https://ultros.app/item/{world}/{id}` (item_view.rs:1704), but the canonical (and the sitemap's submitted form) is worldless `https://ultros.app/item/{id}`. On `/item/{id}` — the crawled route — `world()` falls back through the cookie → GuessedRegion → IP geolocation, so the same page asserts a breadcrumb terminus of North-America or Europe depending on which crawl IP fetched it, while declaring a different URL canonical in the same document. Fix: use the canonical worldless URL for the final node, or omit `item` on the last ListItem (Google explicitly permits that).

2. **Category URL is only half-encoded.** `s_name.replace("/", "%2F")` (item_view.rs:1695) handles slashes but not spaces/apostrophes, and essentially every real ItemSearchCategory name has spaces ("Gladiator's Arm"). A literal space makes the JSON-LD `item` an invalid URI — and unlike an href, nothing normalizes it. Use the repo's existing pattern: `utf8_percent_encode(..., COMPONENT_UNRESERVED)` from routes/item_view_scope.rs:29 (the consumer already percent-decodes, item_explorer.rs:178). The test at item_view.rs:2140 currently enshrines the broken output (`...Two-Handed Sword` with a literal space) — update it.

3. **Unknown/malformed item ids ship invalid breadcrumbs.** `/item/abc` → item_id 0 → `"name": ""` (empty name is invalid per Google) with `item: .../item/North-America/0`, plus a category node linking `/items/category/` (empty segment, matches no route). Guard: render no JSON-LD when `item_name()` is empty / the id isn't in tracked_data().

### Coordination

4. **Hard merge conflict with PR #1015** (verified via merge-tree): this PR inserts its script right after the `<Link rel="canonical" prop:href=...>` line that #1015 deletes and replaces with `<MetaCanonical>`. When rebasing, keep #1015's MetaCanonical — restoring the old Link will also trip the new check_ci.sh `prop:` guard #1015 adds.

### Minor

- Breadcrumb `name` uses ItemUiCategory while the URL targets an ItemSearchCategory page whose title is the search-category name — the markup asserts a page name that isn't the linked page's title. Inherited from the visible link, but now it's a machine-readable claim; consider using `s.name` for the node name.
- Harden inner_html against `</script>` breakout: serde_json doesn't escape `<` — add `.replace('<', "\u003c")` on the serialized string (one line, free).
- World segment in the item URL isn't percent-encoded (item_view_scope.rs's `item_href` exists for this).
- `serde_json::to_string(...).unwrap_or_default()` on failure emits an empty script body — fail loudly (expect) or skip the tag entirely.
- The issue's verification step (curl a local item page, confirm the script body survives into served HTML + run a schema validator) wasn't performed — please include the curl output in the PR after the fixes.

Run `./check_ci.sh` before pushing.

akarras:
Heads up — this now **conflicts with `main`** and needs a rebase.

This PR touches `ultros-frontend/ultros-app/src/routes/item_view.rs`, which two commits merged today also changed:

- **#1015** replaced `<Link rel="canonical" prop:href=…>` with `<MetaCanonical href=…>` (and dropped the `leptos_meta::Link` import).
- **#1001** re-keyed the category link from a localized name to `s.key_id.0`.

Verified locally: `git merge-tree --write-tree origin/main pr-1021` → **CONFLICTS**.

Relevant to a BreadcrumbList JSON-LD change specifically: `/items/category/:category` URLs are now keyed on the **numeric `ItemSearchCategory` id**, not the localized name (`/items/category/17`, not `/items/category/Shields`). The old name form still resolves via a fallback, but anything this PR emits into structured data should use the id form — the name form only resolves for English visitors, since SSR renders game data in English regardless of locale. `sitemap.rs` and the search index were moved to ids in the same change.

Also note `check_ci.sh` now fails on `prop:(href|alt|src|content)=` in `ultros-app/src/`.

PR #1022, opened 2026-07-30
fix(listings): fix unsorted diff and mismatched merge key in listing ingestion

akarras:
## Summary

- `remove_listings` fed `PartialDiffIterator` two **unsorted** inputs (DB rows with no `ORDER BY`, and the raw websocket `Vec<ListingView>`), even though the iterator explicitly assumes sorted sets. Only positionally-lucky rows got matched and deleted; the rest survived as phantom listings feeding `cheapest_listings()`, the Flip Finder buy price, and item pages.
- `update_listings` sorted its incoming/DB lists by `(hq, quantity, price, retainer_name)`, but the merge loop compared `(price, quantity, retainer_name, hq)` — a different field order with different `Option` unwrap semantics. Unchanged rows could get misclassified as add+remove, causing spurious churn that drove false undercut alerts, the live ticker, and analyzer cache invalidation.
- `remove_listings` never called `set_last_updated`, so removals didn't refresh the catch-up marker (`update_listings` already did, same subsystem PR #986 fixed).

## Fix

Both diff paths are now extracted into pure, unit-tested functions:
- `listings_to_remove` — sorts both the DB rows and the incoming removal list by one canonical key (world, retainer, price, quantity, hq — matching `ListingData`'s `PartialOrd<ListingView>` impl) before diffing. Duplicate/identical listings are legal, and sorting both sides makes the match multiset-correct rather than positional.
- `diff_update_listings` — one canonical key (`hq, quantity, price, retainer` with consistent `unwrap_or` semantics) used for both sorts and the merge-loop comparator, so all three agree on what "equal" means.
- `remove_listings` now calls `self.set_last_updated(world_id, item_id)`.

## Tests

5 new unit tests in `ultros-db/src/listings.rs` (`listings::diff_tests`), using a small deterministic xorshift PRNG to shuffle inputs (reproducible, no new dependency):

- `remove_listings_deletes_exact_shuffled_subset` — 25 shuffled listings, random subset removed, asserts the exact subset (and nothing else) is deleted.
- `remove_listings_handles_duplicate_identical_listings_by_multiplicity` — 3 identical DB rows + 1 unrelated, removes exactly 2 of the identical ones.
- `update_listings_reports_no_churn_for_unchanged_listings` — shuffled unchanged listings produce zero adds/removes.
- `update_listings_unwraps_missing_quantity_consistently_between_sort_and_merge` — adversarial case where raw `Option` sort order disagrees with resolved (`unwrap_or`) order; would misclassify unchanged rows as churn under the old code.
- `update_listings_adds_and_removes_only_the_changed_rows` — genuine add + genuine remove among shuffled unchanged rows, asserts only the changed rows are flagged.

Each of the two bug-catching tests was verified by temporarily reintroducing the original buggy logic and confirming the test failed for the expected reason, then restoring the fix and confirming all 5 pass (TDD red/green).

### Test output (local)

```
$ cargo test -p ultros-db --lib listings
running 5 tests
test listings::diff_tests::update_listings_unwraps_missing_quantity_consistently_between_sort_and_merge ... ok
test listings::diff_tests::update_listings_reports_no_churn_for_unchanged_listings ... ok
test listings::diff_tests::update_listings_adds_and_removes_only_the_changed_rows ... ok
test listings::diff_tests::remove_listings_handles_duplicate_identical_listings_by_multiplicity ... ok
test listings::diff_tests::remove_listings_deletes_exact_shuffled_subset ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 26 filtered out; finished in 0.00s
```

### `check_ci.sh` (local)

```
$ cargo fmt --all -- --check
(clean)

$ cargo clippy --all-targets -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 18m 05s
(no warnings)
```

CI does not run `cargo test`, so the output above is from a local run against this branch.

### DIFF
--- a/ultros-db/src/listings.rs
+++ b/ultros-db/src/listings.rs
@@ -69,6 +69,168 @@ impl PartialOrd<ListingView> for ListingData {
     }
 }
 
+/// Sort key for `active_listing::Model`/`retainer::Model` pairs matching the field
+/// order of `ListingData`'s `PartialOrd<ListingView>` impl above: world, retainer
+/// name, price, quantity, hq.
+fn remove_diff_key_model<'a>(
+    listing: &active_listing::Model,
+    retainer_name: &'a str,
+) -> (u16, &'a str, i32, i32, bool) {
+    (
+        listing.world_id as u16,
+        retainer_name,
+        listing.price_per_unit,
+        listing.quantity,
+        listing.hq,
+    )
+}
+
+/// Same key, computed from the incoming websocket view, using the same
+/// `unwrap_or_default` semantics as the `PartialOrd<ListingView>` impl.
+fn remove_diff_key_view(listing: &ListingView) -> (u16, &str, i32, i32, bool) {
+    (
+        listing.world_id.unwrap_or_default(),
+        listing.retainer_name.as_str(),
+        listing.price_per_unit.unwrap_or_default() as i32,
+        listing.quantity.unwrap_or_default() as i32,
+        listing.hq,
+    )
+}
+
+/// Diffs the DB's current listings against the incoming "listings that no longer
+/// exist" view from the websocket, returning the DB rows to delete.
+///
+/// `PartialDiffIterator` assumes both inputs are already sorted by the same key;
+/// duplicate/identical listings are legal (a retainer can have several identical
+/// listings), so both sides are sorted here by the exact comparator key before
+/// diffing, which also makes the match multiset-correct (each DB row pairs with
+/// exactly one incoming row with a matching key, not just a positionally lucky one).
+fn listings_to_remove(
+    db_listings: Vec<(active_listing::Model, retainer::Model)>,
+    mut remove_listings: Vec<ListingView>,
+) -> Vec<active_listing::Model> {
+    let mut db_listings = db_listings;
+    db_listings.sort_by(|(a, ar), (b, br)| {
+        remove_diff_key_model(a, &ar.name).cmp(&remove_diff_key_model(b, &br.name))
+    });
+    remove_listings.sort_by(|a, b| remove_diff_key_view(a).cmp(&remove_diff_key_view(b)));
+
+    PartialDiffIterator::new(
+        db_listings.into_iter().map(|(l, r)| ListingData(l, r)),
+        remove_listings.into_iter(),
+    )
+    .filter_map(|listing| match listing {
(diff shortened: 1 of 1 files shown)


akarras:
## Review + fixes pushed (5d7665ce)

Deep review verdict on the original head: core diff/merge fixes were **correct** (comparator totality, multiset handling, and event semantics all verified — correctly removing phantoms strictly improves the cheapest cache and undercut accuracy), but four findings needed fixing. All are now addressed in 5d7665ce:

1. **`set_last_updated` was stamped unconditionally, including zero-row removes** — re-creating the #986 catch-up-marker failure mode: most `ListingsRemove` events find nothing to delete (the paired full-board update already removed the rows), and each no-op stamped "ingested" while a concurrently-failed `update_listings` left the gap invisible to catch-up forever. Now gated on `!items.is_empty()`.
2. **The remove key couldn't match rows the insert path wrote**: `remove_diff_key_view` used `unwrap_or_default()` (price→0) while `create_listing` stores `unwrap_or(total)` — a None-price listing was stored as price=total and removed with key price=0 → never matched → permanent phantom (the exact Bug-1 symptom surviving the fix). Now uses the same fallbacks as the insert path.
3. **Comparator drift is now impossible**: the hand-written five-arm `PartialOrd<ListingView>`/`PartialEq` impls are one-liners delegating to the same key functions the sorts use — the drift class that caused both original bugs can't recur silently.
4. **Tests now actually discriminate**: two of the three original update tests passed on the OLD code. Added `update_listings_keeps_hq_row_when_nq_row_disappears` (old code: added=1/removed=2 churn; new: removed=[NQ], added=[]) and `remove_listings_matches_rows_stored_from_none_price_views` (old code removes nothing), and strengthened the shuffled-subset test to vary retainer/hq/duplicate prices instead of price only.

Gates run locally (CI has `cargo test` disabled): `cargo test -p ultros-db` → 26 passed / 0 failed; fmt-check and `clippy -p ultros-db --all-targets -D warnings` clean.

Two operational notes for after merge:
- Correct phantom removal means more `remove_listing` calls hitting the "was cheapest" re-query branch in analyzer_service — expect a real step-up in DB round-trips post-deploy; that's the cost of correctness, just don't read it as a regression.
- Each `ListingsRemove` that deletes rows now also upserts the marker row concurrently with the paired `ListingsAdd`'s upsert on the same `(item, world)` key — row-level contention only, but it's new write traffic on that table.

PR #1023, opened 2026-07-30
feat(build): worktree fallback for submodule data dirs

akarras:
## Summary

`xiv-gen-db` and `ultros-xiv-icons` no longer require submodules to be initialized in linked git worktrees. Both build paths resolve their data dir through a fallback chain:

1. `FFXIV_DATAMINING_DIR` / `UNIVERSALIS_ASSETS_DIR` env override (`cargo:rerun-if-env-changed` wired)
2. the local submodule, when populated — including the nested cn/ko/tc csv submodules (probe: `csv/{en,cn,tc}/Item.csv` + `csv/ko/csv/Item.csv`; for icons an `icon2x/` with ≥10k files — full set ~17.2k — so an interrupted fetch falls back instead of shipping a truncated tarball)
3. the **main worktree's copy**, discovered via `git worktree list --porcelain`, with a `cargo:warning` when the fallback engages and another when the worktree's pinned submodule SHA drifts from what main has checked out — on **both** paths

CI and Docker (`actions/checkout` with `submodules: recursive`) hit case 2 and are unchanged.

Also:
- **both** consuming build scripts register the **resolved** data dir for `rerun-if-changed` (un-canonicalized — Windows `\?\` paths confuse cargo): `xiv-gen-db` registers `<resolved>/csv` so a datamining bump still re-runs the rkyv generation now that the CSVs may live outside the package (and the pin-drift warning stays alive); the icons side registers the resolved `icon2x` — a fixed `./universalis-assets/icon2x` doesn't exist under the fallback, and cargo re-runs a build script on *every* build when a registered path is missing (would have re-run the ~5-minute resize each build); fixes the pre-existing `rerun-if-change` typo
- the worktree-discovery/pin-drift mechanics live once in `xiv-gen/src/worktree_fallback.rs`, `include!`d by both `csv_to_rkyv.rs` and the icons build.rs, so the two paths can't drift; resolution is a `Result`-returning, warning-collecting resolver (`csv_to_rkyv::resolved_datamining_dir()`), with panic + `cargo:warning` emission at the build-script boundary
- the feature-gated `csv_to_rkyv` tests get a real gate: `check_ci.sh` now lints `-p xiv-gen --features csv_to_rkyv`, and AGENTS.md documents the local `cargo test -p xiv-gen --features csv_to_rkyv` run (CI's test step is disabled)
- CLAUDE.md worktree instructions updated; the `--reference` init dance is now main-checkout-only guidance
- adds the phase-2 design spec (`docs/superpowers/specs/2026-07-30-game-data-packs-design.md`): LFS data packs with measured wins (icons 130MB → ~30MB at lossy q85, rkyv client payload −34% via zstd-19, pinned-SHA sparse fetch validated at seconds vs a 1.6GB clone). Follow-up PRs implement it.

## Verification

- `cargo check -p xiv-gen-db` in a worktree with **uninitialized** submodules: fell back to the main checkout's copy with the expected warning, and the build-script output registers `cargo:rerun-if-changed=<main>/xiv-gen/ffxiv-datamining/csv` (un-canonicalized)
- after initializing submodules (populated path): `cargo clippy -p ultros-xiv-icons --all-targets` re-ran the icon build against the local submodule, registering the local resolved `icon2x` + the shared helper file
- `cargo test -p xiv-gen --features csv_to_rkyv` — 3 passed (worktree parsing + drift-message tests, shared by both build paths)
- `cargo clippy -p xiv-gen -p xiv-gen-db --features xiv-gen/csv_to_rkyv --all-targets -- -D warnings` + `cargo fmt --all -- --check` — clean

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,12 @@ This repository enforces strict CI checks. Before committing any code, you **mus
 
 Failure to follow these steps will result in CI failures.
 
+Note on feature-gated code: `xiv-gen`'s `csv_to_rkyv` module is behind the
+non-default `csv_to_rkyv` feature, so plain `cargo test -p xiv-gen` compiles
+its tests out (and CI's test step is disabled entirely). When touching that
+module, run `cargo test -p xiv-gen --features csv_to_rkyv` locally —
+`check_ci.sh` lints it, but nothing else executes its tests.
+
 ## Git hooks (optional but recommended)
 
 Tracked hooks live under `scripts/hooks/`. One-time install:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,6 +9,10 @@ Run `./check_ci.sh` from the repo root. It runs `cargo fmt --all -- --check` and
 
 ## When the submodule isn't initialized
 
+**Worktrees usually need no submodule setup.** The `xiv-gen-db` and `ultros-xiv-icons` build scripts resolve their data dirs with a fallback chain: `FFXIV_DATAMINING_DIR` / `UNIVERSALIS_ASSETS_DIR` env override → the local submodule if populated → **the main worktree's copy** (discovered via `git worktree list`). As long as the main checkout has its submodules populated, builds in a linked worktree just work; a `cargo:warning` tells you when the fallback (or a pin drift between the worktree's recorded submodule SHA and what main has checked out) is in play. `ultros/static/classjob-icons` is runtime-only static content — not needed to build.
+
+The rest of this section is about populating the **main checkout** (or a standalone clone).
+
 `./check_ci.sh` runs clippy which compiles the whole workspace, and the `xiv-gen-db` build script reads from `xiv-gen/ffxiv-datamining/` — a git submodule. The csv data for `cn`, `ko`, `tc` lives in *nested* submodules of `ffxiv-datamining` (separate xivapi-adjacent repos), so a non-recursive init only gets you en/ja/de/fr and the build still panics on `cn/Item.csv`.
 
 ### Use `--reference`, not `--init --recursive`
--- a/check_ci.sh
+++ b/check_ci.sh
@@ -6,3 +6,10 @@ cargo fmt --all -- --check
 
 # Run cargo clippy check
 cargo clippy --all-targets -- -D warnings
+
+# xiv-gen's csv_to_rkyv module (and its tests) only compile under this
+# non-default feature, so the workspace clippy above never sees it. Lint it
+# explicitly; the matching test gate is:
+#   cargo test -p xiv-gen --features csv_to_rkyv
+# (CI's test step is disabled, so run that locally when touching the module.)
+cargo clippy -p xiv-gen --features csv_to_rkyv --all-targets -- -D warnings
--- a/docs/superpowers/specs/2026-07-30-game-data-packs-design.md
+++ b/docs/superpowers/specs/2026-07-30-game-data-packs-design.md
@@ -42,10 +42,21 @@ resolve their data dirs through a fallback chain:
 1. `FFXIV_DATAMINING_DIR` / `UNIVERSALIS_ASSETS_DIR` env override
    (`cargo:rerun-if-env-changed` wired in the consuming build scripts);
 2. the local submodule, when populated (probe: `csv/{en,cn,tc}/Item.csv` +
-   `csv/ko/csv/Item.csv`; for icons a non-empty `icon2x/`);
+   `csv/ko/csv/Item.csv`; for icons an `icon2x/` holding ≥10k files — the full
+   set is ~17.2k, so an interrupted fetch falls back instead of shipping a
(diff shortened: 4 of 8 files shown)


akarras:
## Review: request changes — the known rerun-if-changed trap is fixed on the icons path and left open on the CSV path (the higher-traffic one). Fixes incoming on this branch.

### Critical

1. **`xiv-gen-db` never registers the resolved CSV dir for rerun-if-changed** — build.rs emits only `rerun-if-env-changed=FFXIV_DATAMINING_DIR` + `rerun-if-changed=build.rs`. And this is a *regression*, not just a gap: before this PR the CSVs lived inside the `xiv-gen` path-dependency, so a submodule bump dirtied the package and re-ran the build script. With the fallback reading the main clone (outside every package's file list), nothing dirties — bump datamining to 7.6 and every linked worktree keeps emitting 7.55 rkyv forever, silently. It's also currently unfixable from the build script: `datamining_dir()` is private, so the script can't learn the path it read. Fix: expose the resolved dir and emit `cargo:rerun-if-changed=<resolved>/csv` from xiv-gen-db/build.rs — the icons side already does exactly this (ultros-xiv-icons/build.rs:198-201), so the asymmetry reads as an oversight. Bonus effect: once the script stops re-running, the pin-drift warning stops being emitted too, so fix 1 also keeps fix 2 alive.

### Major

2. **Pin-drift detection exists only for datamining** — the PR body and the Phase-1 spec claim it for both paths, but `warn_on_pin_drift` has no equivalent in the icons build; a worktree on a pre-bump branch silently builds images.tar.zst from main's newer icons. Add it (or amend body+spec to say datamining-only).
3. **The new tests can't run under any command anyone uses**: they live behind the non-default `csv_to_rkyv` feature, which only a build-dependency enables (decoupled under resolver v2), CI's test step is commented out, and check_ci runs clippy WITHOUT --all-features — so `./check_ci.sh` green doesn't even lint them. Needs `cargo test -p xiv-gen --features csv_to_rkyv` documented/wired, and the PR-body verification claim corrected.
4. `parse_main_worktree` is duplicated between csv_to_rkyv.rs and the icons build.rs (the icon copy inlined, untestable), held in sync by a comment — and they've already diverged on day one (finding 2).

### Minor

- `assets_populated` accepts a single file in icon2x/ — an interrupted fetch passes the probe and ships a silently incomplete tarball; use a count threshold (~17k) or named probe files, matching the CSV side's rigor.
- `datamining_populated` probes 1 sheet per language but the build reads 31 — a partially-missing checkout dies later with a raw open error instead of the actionable panic.
- `datamining_dir()` prints cargo: directives and panics inside a pub library fn — fine today (only the build script calls it), but a Result-returning variant with the panic at the build-script boundary is cleaner, and finding 1 needs the accessor anyway.

### Verified clean

Main-clone discovery via `git worktree list --porcelain` is correct (bare mains fall through cleanly); CI/Docker safe (all four workflows use recursive submodules so the fallback is inert; where git is absent the panic still names the real problem); ko's nested `csv/ko/csv/` layout handled correctly; Windows paths sound (the \?\ UNC form never reaches the rerun-if-changed line); no scope creep.

akarras:
## Fixes pushed (35d8c42f)

All review findings addressed:

- **rerun-if-changed regression (critical)**: `xiv-gen-db/build.rs` now consumes a new public `resolved_datamining_dir()` (Result + warnings-vec — the library fn no longer prints cargo directives or panics; that happens at the build-script boundary) and registers `cargo:rerun-if-changed=<resolved>/csv` with the un-canonicalized path, mirroring the icons side's Windows-safe approach. **Live-verified in the actual fallback scenario**: built `xiv-gen-db` in a worktree *before* initializing submodules and confirmed the build-script output contains the resolved-path rerun line (no \?\ prefix). A datamining bump in the main clone now rebuilds worktree rkyv again.
- **Icons pin-drift warning**: the fallback branch now emits the same drift warning as the CSV side, via a shared helper.
- **Dedup**: `main_worktree`/`parse_main_worktree`/`pin_drift_warning` now live in one shared `xiv-gen/src/worktree_fallback.rs`, `include!`d by both consumers (std-only, fully qualified), with tests — the two copies can no longer drift.
- **Tests actually run now**: check_ci.sh gained `cargo clippy -p xiv-gen --features csv_to_rkyv --all-targets -- -D warnings`, and AGENTS.md documents the local `cargo test -p xiv-gen --features csv_to_rkyv` requirement (3/3 pass).
- Minors: `assets_populated` requires ≥10,000 icon files (full set is 17,208 as of 7.55 — provenance commented) so interrupted fetches fall back instead of shipping a partial tarball; the late CSV-open panic now names the resolved dir and the env override; spec + PR body updated to match reality.

Gates: fmt clean; clippy clean for xiv-gen (+feature), xiv-gen-db, and ultros-xiv-icons (build script exercised on the populated path too); fallback `cargo check -p xiv-gen-db` green with the warning surfaced.

PR #1024, opened 2026-07-30
fix(analyzer): make best_deals profit/ROI match the Flip Finder

akarras:
The home page's Top Opportunities card and the Flip Finder it links to quoted **different gil for the same flip**. Three separate causes, all in the `/api/v1/best_deals` path.

## What was wrong

**1. The backend never applied the 5% marketboard tax.** The frontend applies it by default (`include_tax` starts on), so every backend profit was overstated by 5% of the sale price.

**2. ROI divided by the buy price with no zero-guard.** A 0-gil listing put `inf`/`NaN` into an `f32`; serde_json serializes non-finite floats as `null`, which then fails to deserialize into the client's `f32`.

**3. The sale-price estimates disagreed.** The backend undercut the world floor by 1 and bumped an empty board by 1.2×; the frontend used a plain `median.min(floor)`.

## What changed

`flip_profit_and_roi` in `analyzer_service.rs` now owns the profit/ROI math: it applies `POST_TAX_MULTIPLIER`, returns `None` for a nonpositive cost (the call site drops those rows with `?`), and clamps ROI to ±100000 to match the frontend's `ROI_DISPLAY_CEILING` so both surfaces agree at the tail.

On the valuation split, I picked alignment over documenting-as-different. `calculate_valuation` had **no callers besides `get_best_resale`** — no undercut alerts depend on the `floor − 1` semantics — so aligning was safe. It's renamed `estimate_sale_price` and now computes `median.min(floor)`, identical to the frontend, with a doc comment at the definition site explaining that it is *not* a "what should I list at" number and that a lister-facing valuation should be a separate function rather than a skew of this one.

Retainer-city tax variation remains unmodeled on both sides, as before.

## One change beyond the immediate scope

Applying the tax broke something adjacent that had to be repaired in the same commit. The Top Opportunities card back-solved its "Buy X · Sell Y" line as `buy = profit × 100 / ROI`, `sell = buy + profit`. Once `profit` went post-tax, `buy + profit` became *what you keep*, not *what you list at* — the card would have quoted a sell price 5% too low. The back-solve was also already wrong whenever ROI got clamped.

So `buy_price` and `est_sale_price` now ship on the wire and the card reads them directly, with the old derivation kept only as a fallback for a server that predates the fields. **If you'd rather not grow the DTO**, the alternative is relabeling the card's "Sell" to mean net proceeds — happy to switch.

## Testing

- 4 new backend tests: tax application, zero-guard, clamp/finiteness, and valuation alignment with the frontend formula.
- 3 new frontend tests for `derive_buy_sell`, including the clamped-ROI case that previously produced a nonsense buy price.
- Full `ultros-app` suite: 304 passed, 0 failed.
- `./check_ci.sh` exits 0 — clean `fmt` and `clippy`.

Two **pre-existing** failures in `cargo test -p ultros --bin ultros`, neither introduced here:

- `analyzer_service::tests::test_persistence` hangs indefinitely. Confirmed by stashing this branch's changes and reproducing on the unmodified tree. It `set_current_dir`s process-globally and constructs a ClickHouse client.
- `web::api::push::tests::label_for_safari_on_ios` fails — a pure-function test on a hardcoded UA string where the iPhone UA's `like Mac OS X` is detected as macOS.

Both look worth a separate fix.


akarras:
## Review: LGTM with one doc fix before merge

Verified in full: the tax math matches the frontend bit-for-bit (same f32 multiply, same truncating `as i32`, same order); the zero/negative-cost guard makes ROI finite by construction (bounded quotient + clamp — no NaN can propagate); `calculate_valuation` had exactly ONE non-test caller on main (`get_best_resale`) so the undercut/alerting path is untouched — the PR body's claim checks out; no retainer-city tax added; and the DTO growth (`buy_price`/`est_sale_price`) is genuinely forced by the tax change (the card back-solved sell = buy + profit, which post-tax would have told users to list 5% below the price the profit assumed) and is backwards-compatible.

### Fix before merge (doc accuracy, not code)

The comments at `best_deals.rs:27-33`, `analyzer_service.rs:331`, and `:1663` claim the card and the Flip Finder now "agree on a given flip." The **formula** is identical; the **median feeding it** is not:
- backend: `select_nth_unstable(len/2)` — upper-middle, no even-n averaging, raw window, no sniper clamp;
- frontend: sniper-clamped set, even-n average of the two middle values.

Two sales at 100 and 200 gil → backend est 200/profit 90/ROI 90%; frontend est 150/profit 42/ROI 42%. Same flip, different numbers — and the divergence is largest on exactly the troll/laundered rows (frontend drops troll rows and ignores troll floors; backend's `estimate_sale_price(1_000_000, Some(1))` canonicalizes returning 1). The code is still a clear improvement — just reword the comments to "same *shape*; the medians are computed differently and the surfaces can still disagree on low-sample or troll-adjacent rows" so a future maintainer doesn't trust an equivalence that doesn't hold.

### Non-blocking notes

- `POST_TAX_MULTIPLIER` and `ROI_CEILING` mirror bare literals in the frontend (`0.95` at analyzer.rs:953, `ROI_DISPLAY_CEILING` at analysis.rs:301) with doc-comment-only linkage — this PR exists because two surfaces drifted; those constants belong in `ultros-api-types` where one edit moves both.
- Residual display mismatch: backend ships f32 ROI rendered with `{:.0}` (rounds) while the frontend truncates via `as i32` — 90.7% shows as 91% on the card, 90% in the table.
- Undisclosed but correct behavior change: Discord `/ffxiv analyze profit` is the second `get_best_resale` consumer — its embed now shows post-tax numbers and its `minimum_profit` filters post-tax. Worth a line in the PR body so it isn't filed as a regression later.
- `estimate_sale_price` lost the old `max(1, ...)` floor — `(100, Some(0)) == 0` is now asserted as intended. Both real callers filter on min_profit so it's theoretical, but the card's `profit > 0` is the only guard left for a generic API consumer.
- Test gap flagged for a follow-up: nothing asserts `get_best_resale` drops the nonpositive-cost row at the `?` boundary, and the repo's own CH-surface design doc asks for a serde round-trip test against the server DTO fixture — this PR hand-syncs two DTOs across crates and was the moment to add it.

(Windows note: the ultros test binary doesn't link on MSVC, so the four backend tests need CI or a Linux run; clippy `--all-targets` is the local compile gate.)

PR #1033, opened 2026-07-30
feat(data): replace game-data submodules with git-LFS packs

akarras:
## What / why

The three game-data submodules (~2.1GB of working tree, chronically failing recursive inits, and the whole reason worktrees needed setup) are replaced by git-LFS-tracked prebuilt packs in `data/` (~26MB total):

- `data/xiv-db/{en,ja,de,fr,cn,ko,tc}.rkyv` — the CSV→rkyv pipeline output (en: 4.52MiB, flate2 container unchanged in this PR)
- `data/icons/images.tar.zst` — **21.4MB, down from 130MB**: icons now encode as lossy WebP q85 (the old `image`-crate encoder was lossless-only; q85 measured at ~20% of the bytes with no visible quality loss at the served 60/40/25px sizes)
- `data/manifest.toml` — the five upstream repos + pinned SHAs (single source of truth, replaces submodule gitlinks)

`xiv-gen-db` and `ultros-xiv-icons` now `include_bytes!` the packs directly — **no build-time CSV parse or 17k-image resize ever again** (minutes off cold builds), and worktrees/CI need zero data setup. A new `game-data-pack` crate regenerates packs via pinned-SHA partial clone + sparse checkout (seconds and ~11MB of git transfer instead of a 1.6GB clone; validated against GitHub). `update_game_data.yml` now regenerates packs nightly and auto-PRs the diff with a generation summary (item deltas + which new items lack icons upstream). classjob-icons' served fonts are vendored (1.7MB); all three submodules and `.gitmodules` are gone.

## LFS evidence

`git lfs ls-files` lists exactly the 8 pack files; git history verified pointer-only at the object level (`git cat-file -s` = 132-133 bytes per pack blob). `.gitattributes` landed before any pack bytes. Build guards fail actionably (`Run: git lfs install && git lfs pull`) when packs are pointer stubs; the wasm-client build (no `embed` feature) never requires packs.

Pack regeneration is byte-deterministic on a fixed toolchain (sorted tar entries, zeroed headers; rkyv 0.7 archived-map output verified insertion-order- and process-independent), so the nightly `git diff --quiet data/` gate only fires on real upstream changes.

## Migration for existing clones

```
git lfs install && git lfs pull
```

Stale submodule directories can simply be deleted. Fresh clones need git-lfs (README updated).

## Notes

- **Cargo.lock**: adding the generator crate re-resolved serenity's open-ended `reqwest >= 0.11.22` constraint, deduping reqwest 0.11.27 → 0.12.28 (the Discord bot's HTTP stack) — a quick Discord smoke test after deploy is recommended.
- **LFS bandwidth**: CI checkouts now pull ~26MB from LFS each run; if quota errors appear, add an `actions/cache` on `.git/lfs` keyed by `git lfs ls-files` hash.
- **First nightly run**: trigger `update_game_data.yml` once via `workflow_dispatch` to confirm the peter-evans LFS write-back end to end.
- **Relationship to #1023**: independent branches; whichever merges second resolves small conflicts in `ultros-xiv-icons/build.rs`, `CLAUDE.md`, and `xiv-gen/src/csv_to_rkyv.rs` — this branch's `build.rs` wins, both keep `read_data_from`, and #1023's fallback chain becomes deletable dead code once no build script reads a submodule.
- **Relationship to the zstd-container PR** (up next): whichever merges second regenerates the rkyv packs in the new container format.
- Full verification trail (per-task reviews + final whole-branch review): `.superpowers/sdd/progress.md` on this branch.


akarras:
Reviewed. The engineering here is good and CI is fully green (lint **and** the 1h13m Docker build). I'm **not merging it**, because it turns on Git LFS for a repo that has never used it and the recurring cost lands outside anything CI can show you. One question from you unblocks it.

## The arithmetic

Measured from the pointer files on this branch:

| Object | Size |
|---|---|
| `data/icons/images.tar.zst` | 21.4 MB |
| `data/xiv-db/*.rkyv` × 7 | 30.0 MB |
| **Per LFS checkout** | **51.4 MB** |

Four workflows check out with `lfs: true` (`rust.yml`, `docker-publish.yml`, `update_game_data.yml`, `glitchtip-upload-test.yml`). `rust.yml` and `docker-publish.yml` both run on every PR push, so:

- **~103 MB of LFS bandwidth per PR push.** GitHub's free allowance is 1 GB/month — roughly **10 pushes**. Today alone this repo took ~14.
- **`update_game_data.yml` runs `0 0 * * *` and opens a PR with regenerated packs.** Pushing that branch uploads ~51 MB of *new* LFS objects, and LFS objects are not garbage-collected when a branch is deleted or a PR is closed. That's **~1.5 GB/month of permanent storage growth** against a 1 GB allowance — whether or not the nightly PR is ever merged.

For context on how tight the budget already is here: Actions cache is at **7.08 GB of the 10 GB quota**, and the cache-eviction note in `rust.yml` documents a previous incident where quota pressure caused `failed to solve: not_found` on the Docker export.

I can't see the billing endpoint from this token, so if you already have LFS data packs on this account, this is a non-issue and I'll merge on your word.

## The one-way door

Once packs are pushed and referenced from merged history, backing LFS out means rewriting history or leaving the objects billable forever. That asymmetry is why I'd rather ask than merge and find out.

## Options, cheapest first

1. **Confirm you have (or will buy) LFS data packs.** $5/mo per 50 GB bandwidth + 50 GB storage covers this comfortably. Say the word and I'll merge.
2. **Drop `lfs: true` from `glitchtip-upload-test.yml`** — that job doesn't compile `xiv-gen-db` with `embed`, so it shouldn't need the packs. Free saving either way; worth doing regardless.
3. **Have `update_game_data.yml` force-push a single fixed branch** instead of opening a fresh PR branch per night. Doesn't fix storage accretion (old OIDs persist) but stops multiplying it.
4. **Host the packs outside git** — a release asset or object store fetched by the build script, with the SHA pinned in `data/manifest.toml`. Keeps every build-time win in this PR, no LFS billing. Larger change.

## Also needs doing before merge

The branch is now `CONFLICTING` against main in `CLAUDE.md`, `xiv-gen-db/build.rs`, `xiv-gen/src/csv_to_rkyv.rs` and `ultros-frontend/ultros-xiv-icons/build.rs` — #1023 (worktree fallback for submodule data dirs) landed after this branch was cut. Resolution is "take this branch's side" and delete `xiv-gen/src/worktree_fallback.rs`: with no submodules there's nothing to fall back to, so #1023 becomes dead code. Worth calling that out in the PR body so it doesn't read as an accidental revert.

## What I checked and liked

- `lfs_guard::assert_not_lfs_pointer` gated on `CARGO_FEATURE_EMBED` is the right call — the wasm client never reads the packs, so a pointer-stub checkout shouldn't fail its build.
- `cargo:rerun-if-changed` is declared per pack, so an LFS pull that swaps a stub for real content re-runs the guard. That's the trap from the phase-1 design notes, handled.

PR #1035, opened 2026-07-31
fix(best_deals): align backend profit/ROI with the Flip Finder's estimate

akarras:
## Summary

The `/api/v1/best_deals` backend (home page "Top Opportunities" card) and the Flip Finder table computed flip profit/ROI differently, so the same flip showed different numbers in each place. Two fixes:

- **Tax + zero-guard** (`ffc71efe`): `best_deals` computed profit/ROI with no market tax, while the Flip Finder applies a 5% tax by default. A `cheapest_price <= 0` could also produce inf/NaN ROI in the serialized response. Added `calculate_profit_and_roi()`, which applies the 5% tax (matching the frontend default) and returns ROI `0.0` when cost is non-positive — mirroring the frontend's own guard in `analysis.rs::return_on_investment`.
- **Valuation formula alignment** (`ea7712bc`): `calculate_valuation` undercut the current floor by 1 gil and bumped empty-market estimates by 1.2x, while the frontend's `estimated_sale_price` is just the median capped by the world floor, guarded against "troll" listings (implausible prices >50x the median). Dropped the undercut/bump and ported the frontend's `is_troll_listing` guard into `get_best_resale` so both surfaces drop/ignore the same rows. Confirmed via grep that `calculate_valuation` has no other callers (e.g. undercut alerts don't use it), so this was safe to change.

Retainer-city tax variation is intentionally not modeled anywhere — out of scope.

## Test plan

- [x] Added unit tests for the tax + zero-guard behavior (`test_calculate_profit_and_roi_applies_market_tax`, `test_calculate_profit_and_roi_guards_nonpositive_cost`)
- [x] Rewrote `test_calculate_valuation` for the new median/floor semantics and added `test_is_troll_listing`
- [x] All tests written test-first (RED confirmed before implementation, GREEN after)
- [x] `./check_ci.sh` (fmt + clippy `-D warnings`) passes clean
- [ ] Not verified: actual home-card vs. Flip-Finder rendering in a running app — no dev environment was set up for this session

### DIFF
--- a/ultros/src/analyzer_service.rs
+++ b/ultros/src/analyzer_service.rs
@@ -326,10 +326,28 @@ struct AnalyzerState {
     cheapest_items: BTreeMap<AnySelector, CheapestListings>,
 }
 
+/// Ratio above the median at which a listing is presumed to be a "troll"
+/// price meant to game rankings rather than a real sale opportunity.
+/// Mirrors `TROLL_MULTIPLE` in
+/// ultros-frontend/ultros-app/src/routes/analyzer.rs so both surfaces treat
+/// the same rows as implausible.
+const TROLL_MULTIPLE: i64 = 50;
+
+fn is_troll_listing(price: i32, median: i32) -> bool {
+    median > 0 && (price as i64) > (median as i64).saturating_mul(TROLL_MULTIPLE)
+}
+
+/// Estimated sale price for a flip: the median recent sale price, capped by
+/// the current world floor when one exists. Mirrors the frontend Flip
+/// Finder's `estimated_sale_price` computation
+/// (ultros-frontend/ultros-app/src/routes/analyzer.rs) so the home-page
+/// "Top Opportunities" card and the Flip Finder table agree on the same
+/// flip. Callers should filter troll listings (see `is_troll_listing`) out
+/// of `current_listing_price` before calling this.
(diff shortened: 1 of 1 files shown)


akarras:
Superseded by #1024, which merged at 07:02 and implements the same fix (5% market tax + zero-cost guard + `calculate_valuation` → `estimate_sale_price` alignment with the Flip Finder's `median.min(floor)`).

Closing rather than merging because this version regresses the zero-cost guard relative to what shipped:

- **#1024** returns `None` for `cost <= 0` and the call site drops the row with `?`. A 0-gil listing is corrupt data or a trade-channel handoff — there is no cost basis to quote.
- **This PR** keeps the row with `profit = 950, roi = 0.0`. That publishes a fabricated "950 gil profit at 0% ROI" entry on the Top Opportunities card built from data we know is bad.

It would also no longer compile against main: `ResaleStats` now carries `buy_price`/`est_sale_price` on the wire (#1024), which `top_opportunity.rs` reads directly instead of back-solving `buy = profit * 100 / ROI` — a derivation that broke once profit went post-tax.

**One idea here is worth keeping**: the backend `is_troll_listing` filter. The frontend excludes troll listings before computing the world floor and the backend does not, so the two can still disagree on `est_sale_price` for items with an absurd outlier listing. That's a real remaining gap and worth its own small PR against current main.

PR #1037, opened 2026-07-31
⚡ Bolt: Optimize string substring match for case insensitive strings

akarras:
⚡ Bolt: Avoid heap allocations for case-insensitive string matching

💡 What: Created `string_utils::contains_ignore_ascii_case` and `starts_with_ignore_ascii_case` to replace `.to_lowercase().contains(...)` loops in hot render paths.
🎯 Why: `.to_lowercase()` allocates a new `String` on the heap. When performed in a loop (like searching through all FFXIV items in a dropdown), this results in thousands of unnecessary allocations on every keystroke/render, putting pressure on memory and garbage collection.
📊 Impact: Reduces heap allocations significantly during item search / filtering.
🔬 Measurement: Verify search functionality in Lists/Alerts drop-down or by tracking JS heap usage in browser devtools while searching.

---
*PR created automatically by Jules for task [9829487518192420136](https://jules.google.com/task/9829487518192420136) started by @akarras*


akarras:
Closing — the optimization is worth doing, but this implementation silently breaks item search for every non-English locale.

## The regression

Both helpers compare bytes with `eq_ignore_ascii_case`, which folds only `A-Z`/`a-z`. Every call site still builds the needle with the **Unicode-aware** `s.to_lowercase()`, but now passes the **raw, un-lowercased** name as the haystack. So an uppercase non-ASCII character in the name can never match its lowercase form in the query.

Reproduced with the exact helper from this PR, called exactly as the call sites call it:

```
name               query        old    new
Épée de Bronze     épée         true   false   <-- regression
ÄTHERISCHES ÖL     ätherisch    true   false   <-- regression
Éclat de Terre     éclat        true   false   <-- regression
Bronze Sword       bronze       true   true
BRONZE SWORD       bronze       true   true
```

This is not hypothetical — in `ffxiv-datamining`, `csv/fr/Item.csv` has 46,115 lines containing uppercase accented characters and `csv/de/Item.csv` has 35,673. It hits `SearchBox`, `CreateAlertDrawer`, `AddRecipeToCurrentListModal`, `ListView`, and the Flip Finder's `?name=` filter.

Worth noting `add_recipe_to_current_list.rs` also changes *what* is searched: it dropped the precomputed `lower` field from the tuple and now matches against `i.name`. If that cached string was doing anything other than lowercasing the same name, this is a second behavior change.

## Two other things

- `test_clippy2.sh` is committed at the repo root. That's a scratch file.
- The `.jules/bolt.md` entry is duplicated: a `2024-06-25` version whose backticks got eaten ("allocate intermediate  instances with  or ") followed by a `2026-06-25` "(update)" that says the same thing correctly. Should be one entry, and neither date is today.

## If you want to redo this

Keep the allocation out of the loop *without* changing the matching semantics — lowercase the haystack once per item rather than per keystroke, or fold the needle and haystack the same way. `str::to_lowercase` on the query is already once-per-query, so the win is entirely in the per-row path; a `contains_ignore_case` that handles non-ASCII (or a precomputed lowercased name cached alongside the item, which `add_recipe_to_current_list.rs` was already doing) gets the same benefit with no behavior change.

Also worth a real measurement before landing: `str::contains` uses an optimized two-way search, while `windows(n).any(..)` is naive O(n·m). For short queries against short names the allocation may not dominate. The PR's stated measurement is "verify search functionality", which isn't one.

PR #1039, opened 2026-07-31
feat(item-explorer): let gear-set icons fill their tile

akarras:
## What

The gear-set cards on the item explorer rendered each piece as a 25px `IconSize::Small` icon centered in a much larger square tile. Most of each tile was padding and card chrome wrapped around a tiny image.

Two changes:

**`ItemIcon` gains an optional `fill` prop.** When set, the wrapper sizes to `100%` of its parent instead of `icon_size`'s fixed `25px`/`40px`/`60px` box, and the `` stretches with `w-full h-full object-contain`. `icon_size` still selects which source variant is fetched, so resolution and rendered size are chosen independently.

**`JobSetCard` uses it.** Pieces now request the 60px `Large` source and fill their square. The tile drops `bg-white/5`, its border, and `p-0.5` in favor of `rounded-md overflow-hidden` with a hairline ring that brightens to brand on hover, plus a slight scale. At the ~50px the 6/5-column grid produces, the Large source downscales cleanly rather than upscaling.

## Why

Reported visually: the icons were small enough that the surrounding card read as unnecessary decoration rather than as a frame for the art.

## Notes for review

- Both `fill` modes emit the same element and attribute shape — only the values differ — so the SSR and CSR view trees still agree for tachys hydration. This matters here; `job_set_card.rs` has prior hydration-mismatch history (see the `GilOrDash` module docs in that file).
- Scope is the set-piece grid only. The material rows on the set detail page still use `Small`, since those are genuine inline icon-beside-text rows where a fixed small box is correct.
- `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` both pass. I did **not** run the app to eyeball the result — the icon art has its own baked-in background, so with the card chrome gone the tiles read as a tighter mosaic of colored squares. If that ends up too dense, `gap-1.5` on the grid is the knob.

### DIFF
--- a/ultros-frontend/ultros-app/src/components/item_icon.rs
+++ b/ultros-frontend/ultros-app/src/components/item_icon.rs
@@ -8,6 +8,13 @@ pub fn ItemIcon(
     #[prop(into)] item_id: Signal<i32>,
     icon_size: IconSize,
     #[prop(optional)] loading: &'static str,
+    /// Stretch the icon to fill its parent instead of pinning it to
+    /// `icon_size`'s fixed pixel box. `icon_size` still selects which
+    /// source variant is fetched, so pass the smallest one that stays
+    /// sharp at the rendered size. Used by tile layouts (gear set
+    /// cards) where the container, not the icon, decides the size.
+    #[prop(optional)]
+    fill: bool,
 ) -> impl IntoView {
     // ⚡ Bolt Optimization: Replace Memo::new with Signal::derive
     // `valid_search_category` and `item_name` are simple map lookups.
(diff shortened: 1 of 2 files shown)


akarras:
Merged after browser verification on a live local build (`/items/jobset/DNC`, desktop 1440px).

The gear-set tiles now render the icon art edge-to-edge — the 6/5-column grid produces ~50px tiles and the `IconSize::Large` (60px) source downscales cleanly, so the result reads as a mosaic of the art rather than tiny icons floating in card chrome. The density concern flagged in the PR notes is not a problem at `gap-1.5`; the tiles stay legible and distinct.

Worth recording for the next person: the `bg-white/5` this PR removed **was already dead**. `style/tailwind.css:821` has a global `a:not(.nav-link):not(.btn)…{ background-color: transparent }` at specificity (0,5,1), which beat the utility on any `<a>` ([[anchor specificity trap]] territory). Same for `rounded` vs the rule's `@apply … rounded-md`. So the visual delta here is genuinely the border and the icon sizing, not a background swap — which matches what the PR set out to do. The new `ring-1`/`hover:ring-brand-500/50` is box-shadow, so it is unaffected by that rule.

Both `fill` modes emit the same element and attribute shape, so the SSR/CSR view trees still agree; hydration completed clean (`boot-progress` absent, no console errors) on the jobset page.

PR #987, opened 2026-07-27
chore(deps): weekly cargo update refresh + js-yaml security patch

akarras:
Automated weekly `big-dependency-updates` run (2026-07-27).

## What changed
- **In-range `cargo update` lockfile refresh** — 24 net version changes, no manifest edits (all via existing caret reqs). Notable: `leptos_router 0.8.15`, `tokio 1.53.1`, `tokio-util 0.7.19`, `tokio-stream 0.1.19`, `libc 0.2.189`, `either 1.17`, `zerocopy 0.8.55`, `libsqlite3-sys 0.30→0.37` (single copy, transitive via sqlx). `concurrent-queue` dropped.
- **lodestone held at `192faccd`** — `cargo update` advances the `async` branch to a reqwest-0.12 commit that breaks the `ultros` crate's 0.11 client; held back as every prior run. reqwest 0.11.27 / 0.12.28 / 0.13.4 split preserved.
- **integration/ e2e harness:** `js-yaml 4.2.0 → 4.3.0` (package-lock only, dev-scope, in-range) fixing **GHSA-52cp-r559-cp3m** (high — YAML merge-key chains → quadratic CPU). `npm audit` → **0 vulnerabilities**.

## Big-dependency ceilings — all UNCHANGED (verified live against crates.io this run)
| Dep | Latest stable | Held at | Gate |
|-----|---------------|---------|------|
| leptos | 0.8.20 (0.9.0-**beta** only) | 0.8.x | `leptos_i18n 0.6.2` pins `leptos ^0.8` + `leptos-use ^0.18`; no 0.6.x successor / no 0.9 stable |
| leptos-use | 0.19.0 | 0.18.3 | duplicate-trap — `leptos_i18n` pins `^0.18` |
| tower-http | 0.7.0 | 0.6.11 | duplicate-trap — `leptos_axum 0.8.10` pins `^0.6` |
| itertools | 0.15.0 | 0.14.0 | duplicate-trap — `tantivy 0.26.1` pins `^0.14` |
| axum / tokio / sea-orm / sea-query | 0.8.9 / 1.53.1 / 2.0.0 / 1.0.1 | = latest | none |
| rkyv / reqwest(0.13) | 0.8.17 / 0.13.4 | 0.7.46 / 0.11 | deferred (wire-format risk / sentry-only + TLS-feature migration) |

**No newly-unblocked major bump this run.** wasm-bindgen family did not move (0.2.126) → no Dockerfile CLI bump needed.

## Security / Dependabot
[...]

PR #1040, opened 2026-07-31
fix(top-opportunities): stop ranking gil-transfer laundering, rebuild the card

akarras:
## The problem

The home-page Top Opportunities card ranked by raw absolute profit, which by construction selects for the most extreme gil-transfer laundering on the world. FFXIV caps direct trades at 1M gil, so a player moving currency between their own characters lists a worthless item at an enormous price and buys it from themselves — the market board is the only viable path. Every row in the reported screenshot was such a trade:

| item | buy → sell | ROI |
| --- | --- | --- |
| Archeo Kingdom Partisan (rendered as "Arc") | 3,608 → 252,000,016 | 6,984,380% |
| Hard Leather Pot Helm | 1 → 133,333,336 | 13,333,334,016% |
| Hempen Coif | 5 → 42,000,000 | 839,999,872% |
| Occult Bracelet of Blood | 10,000,001 → 39,999,966 | +300% |
| Golden Beaver | 89,999,980 → 111,599,989 | +24% |

The outlier wasn't leaking past the ranking — it *was* the sort key. Worse, `DEEP_SCAN_TOP_N` selected which rows got ClickHouse enrichment *by that same compromised metric*, so genuinely good deals never had their quality data fetched. The quality filter sat downstream of the thing it was meant to correct.

Note the last two rows: a ratio-based test can't catch them. Nobody has ever paid 90M for a Golden Beaver, but +24% looks perfectly reasonable.

## Backend — three eligibility layers

All derived from data present on **100% of rows**, and all applied inside the pass-1 `flat_map` so they run *before* the `DEEP_SCAN_TOP_N` truncation. Filtering afterwards would leave the deep-scan budget still spent on laundering — the original bug in a subtler form.

1. **Vendor anchor** — reject valuations above 100× xiv-gen `price_mid`, reusing the multiple already established by the frontend's `real_price`. Kills Hempen Coif and Hard Leather Pot Helm outright: vendor-sold starter gear where a 42M valuation is absurd against a ~50 gil vendor price. No velocity or ClickHouse data needed.
2. **Velocity floor** — from the 6-sale buffer, same formula and thresholds as the Flip Finder's default query. Kills Archeo Kingdom Partisan and Golden Beaver: neither is vendor-sold, neither has a real trade cadence.
3. **ROI ceiling** — covers the velocity floor's blind spot, laundering compressed into a short burst. Set at 5000% rather than something tighter because a legitimate cheap-item flip can genuinely clear 1000% (with `min_profit = 10_000`, a 715 → 10,715 gil flip is a 1400% return).

**ClickHouse stays refinement-only.** The Flip Finder redesign spec measured ~7% coverage of traded items (11 rows from a stratified 150-item sample, flat across price bands). An earlier draft of this used a VWAP sanity cap and sample-size gate as the primary defense; both were dropped because they'd silently no-op on ~93% of rows.

Also fixes the median tie-break to pick the **lower** middle on even-length input. The upper-middle pick resolved a two-sale laundering pair to the higher of the two — and the comment directly above it already asked for this ("If even, pick the lower one to be conservative?") before doing the opposite.

## Frontend — the card

- **Item name gets its own line.** Sharing a row with a 9-digit profit at `text-3xl` mono is what truncated "Archeo Kingdom Partisan" to "Arc". Structural fix, not a tweak.
- **Route line names the source world.** `ResaleStats.world_id` is the *source* world and both card components received it and rendered neither. The entire premise is a cross-world arbitrage and the card never named the other world.
- **ROI removed** from the featured slot and every compact row, following its demotion to an off-by-default column in #991. This is what frees the space the route line and cadence badge occupy.
- **Credibility signals** are the existing `SalesCadenceBadge` plus a recent-price range — both buffer-derived, so present on every card. Where ClickHouse has a row, the 30d VWAP upgrades that slot.
- **Error state split from empty.** `get_best_deals(...).await.ok()` collapsed a failed request into the same `None` as an empty result, rendering an outage as "the market is quiet."
- **Empty state** now states the card's own promise and links to an unfiltered Flip Finder, so the velocity floor reads as deliberate and adjustable.
- **`.card-link`** lets card-sized anchors opt out of the global `a:hover` underline, which was underlining all eleven text nodes at once.

The "view all" link carries the card's ranking through as `?sort=profit&vel=0.2`, so the Flip Finder opens ranked and filtered exactly as the card was.

## i18n

New keys in all seven locales with real translations. Also translates the `sales_cadence_*` labels, which shipped as English stubs in all six non-English locales — pre-existing, but this change puts those badges on the landing page, so it would have been a visible regression.

## Reviewer notes

**Rebased onto #1024.** That landed mid-work and made `profit` post-tax with `buy_price`/`est_sale_price` on the wire. My card was still back-solving buy/sell from `profit + ROI`, which post-tax gives what you *keep*, not what you *list at* — shipping unrebased would have had users listing 5% under market. Took their `derive_buy_sell` wholesale.

One deliberate asymmetry in the merge: the vendor anchor tests the **pre-tax** `est_sale_price` (a laundering listing claims the pre-tax price, so that's what should be caught) while the ROI ceiling tests the **post-tax** ROI (the figure both surfaces display).

**Thresholds were validated against Gilgamesh**, a high-population world. A low-population world may hit the empty state more than intended — the empty state is designed for that, but the numbers deserve a check against a small world.

**Known residual risk:** six laundering sales inside one hour yield a high derived velocity and pass the floor. The ROI ceiling covers the common shape (near-zero buy price) and the vendor anchor covers vendor-sold items, but a non-vendor item bought at a plausible price and "sold" at ~20× within one burst would still pass all three. No 100%-coverage signal we have today detects that.

**Verification status — please read.** `cargo fmt --all --check` passes. Before the rebase, all 14 new `resale_eligibility` tests passed along with the full backend suite. The post-rebase test run was still compiling when this PR was opened, and the merge resolution touched real logic, so **treat CI as the authority here, not my local run.** I'll follow up if it turns up anything.

Two pre-existing failures, unrelated and not fixed here: `web::api::push::tests::label_for_safari_on_ios` (a UA parser now reports macOS where the test expects iOS; `push.rs` unchanged since #645, likely fallout from the dep refresh in #987) and `analyzer_service::tests::test_persistence`, which hangs 60+ seconds on `set_current_dir` plus `ClickHouseClient::from_env()`.

**Not visually verified.** The card compiles, but I have not seen it render. The claims about the full item name at narrow width and the route line naming a different world need someone to run it against real data.

Design doc: `docs/superpowers/specs/2026-07-30-top-opportunities-redesign-design.md`
Plan: `docs/superpowers/plans/2026-07-30-top-opportunities-redesign.md`


akarras:
Reviewed and browser-verified against a live local build (Postgres + ClickHouse, `/` with `HOME_WORLD=Cerberus`).

**The core claim holds.** With the three gates applied, the card ranked five plausible trades rather than laundering — the largest was 1,750,000 → 3,849,999 on a 0.3/day glamour item, corroborated by a 30d VWAP of exactly 3,849,999. No 6,984,380% ROI rows, no vendor-priced starter gear at 42M. `view all` carried `?sort=profit&vel=0.2` through to the Flip Finder as designed, and the route line named the source world on every row.

Also checked:
- `.card-link` opt-out behaves in both themes — in light mode the anchor inherits body text colour (`rgb(11,11,12)`) and the card supplies its own tint, so nothing relies on the global `a:not(…)` rule any more.
- `RouteLine` is SSR-safe: `LocalWorldData` is provided server-side (`ultros/src/leptos.rs:104`, `web.rs:2128`) and world names are not localized, so the node structure agrees across SSR/CSR.
- `cargo fmt --all -- --check` clean on a tree with this branch merged alongside #1039/#1041/#1042 — no conflicts between any of them, and all 7 locale files stayed at an identical key count.

**Two cosmetic defects found and fixed in c850abb** (both only visible once the card actually renders, which is why the PR could not catch them):

1. The buy→sell pair rendered as **three stacked lines** on every row. `Gil` emits a block-level `flex` div, so two of them with a text arrow between never share a line regardless of width — `whitespace-nowrap` does nothing here, the wrapper itself has to be a flex row. Measured 57px → 20px on the featured card, no overflow at the ~390px column.
2. The anchor read `30d avg 3849999` — a bare integer sitting directly beside `3,849,999`. Now goes through `separate_with_commas` like every other gil figure on the card.

Re-verified after the fix; both corrected. Merging.

PR #1041, opened 2026-07-31
fix(world-picker): stop the collapsed value clipping the field, mark kinds with icons

akarras:
## What

The world picker's selected value rendered as a chip that overhung the bottom of its own input, and the "region" / "world" / "datacenter" badge sat inside the box as trailing muted text.

| Before | After |
|---|---|
| chip overhangs the field by 2px; badge text crowds the value | value sits on the input's own text baseline; kind shown as a leading icon |

## Why

`Select` rendered the collapsed value with the same closure it uses for dropdown rows, so the selected value inherited the row's `px-4 py-2`, `rounded-lg` and `hover:bg-…`. That box is 40px tall inside a 42px input and was positioned at `top-1 left-1` — overhanging the bottom border by exactly 2px, which the hover fill made visible as a chip clipping the field.

The collapsed value is now plain text laid out to the input's own metrics (`inset-0`, `px-3 py-2`, transparent border matching `.input`), so it lands in exactly the position the input's real text would.

## Icons

Rendering the value as plain text would have lost the ability to tell whether `North-America` is a region or a similarly-named world, so the kind is now a leading icon on both the dropdown rows and the collapsed field:

| | | |
|---|---|---|
| World | `BsGeoAltFill` | map pin |
| Datacenter | `FaServerSolid` | server rack |
| Region | `FaEarthAmericasSolid` | globe |

The metaphor zooms out as scope widens. `Select` gained an optional `selected_prefix` callback to carry it — existing callers are unaffected since it's `#[prop(optional)]`.

**The icon isn't the only carrier of the semantics.** It also renders `title` plus an `sr-only` label. Those were hardcoded English literals in the view before (against the no-hardcoded-strings rule in CLAUDE.md); they're now wired to i18n. `world` and `datacenter` keys already existed, and `region` is added to all seven locales with real translations.

`WorldOnlyPicker` deliberately gets no icons — everything in it is a world, so a marker would be noise.

## Defects fixed in the same component

- **Enter immediately after opening changed your selection.** The highlight started at 0 and was reset to 0 whenever the result list changed, so confirming the current value instead selected whichever entry sorted first. The highlight is now seeded from the current choice on open. The reset effect is also re-keyed from `final_result` to `current_input` — the former fires when the *async item list* arrives too, which would have yanked the highlight back to 0 underneath an already-open dropdown.
- **The keyboard highlight scrolled out of sight.** There was a `// Scroll into view logic could be added here if needed` and the rows already carried `scroll-mt-2` in anticipation. With ~90 worlds in a `max-h-96` panel, arrowing past the eighth row moved the highlight out of view permanently.

## Affordances

- A chevron on the trailing edge that rotates when open — nothing previously signalled this was a dropdown rather than a plain text field. Input and overlay both take `pr-9` so long names can't run underneath it.
- The current value is echoed as a placeholder while focused, so you can see what you're replacing as you type over it.

## Reviewer notes

- **Why not `scrollIntoView`:** the panel is `fixed` inside a portal, and the browser method drags the whole page around to satisfy it. The fix adjusts only the dropdown's own `scrollTop`. It uses `offsetTop`/`clientHeight` rather than `ScrollIntoViewOptions` specifically to avoid adding `web-sys` features.
- **Why the placeholder is conditional on focus:** set unconditionally it would double-render beneath the overlay text, since the overlay has no background.
- **Deliberately not addressed** (happy to take in a follow-up): an empty search silently falls back to showing the *entire* list, so typing `zzz` looks identical to typing nothing; the dropdown only ever opens downward and runs off-screen near the bottom of a short viewport; and the input has no accessible name, no `aria-controls`, and keeps `aria-activedescendant` set while closed.

## Testing

`./check_ci.sh` passes — fmt and workspace clippy, exit 0, no warnings.

**Not exercised in a running app.** No dev server was available and a full `cargo-leptos` build plus Postgres was disproportionate for a CSS-level change, so the geometry was verified against a markup-faithful reproduction instead: it confirmed the old overlay overhangs by exactly 2px and the new one renders at the same position as the input's own text. Worth a quick manual look at the picker before merging.

### DIFF
--- a/ultros-frontend/ultros-app/locales/cn.json
+++ b/ultros-frontend/ultros-app/locales/cn.json
@@ -42,6 +42,7 @@
     "currency_exchange": "通货兑换",
     "currency_exchange_desc": "消耗神典石，换取金币",
     "datacenter": "大区",
+    "region": "区域",
     "worlds": "服务器",
     "vendor_available": "NPC 可购买",
     "exchange_available": "可兑换",
--- a/ultros-frontend/ultros-app/locales/de.json
+++ b/ultros-frontend/ultros-app/locales/de.json
@@ -42,6 +42,7 @@
     "currency_exchange": "Währungstausch",
     "currency_exchange_desc": "Tomesteine ausgeben, Gil bekommen",
     "datacenter": "Rechenzentrum",
+    "region": "Region",
     "worlds": "Welten",
     "vendor_available": "Beim Händler erhältlich",
     "exchange_available": "Tausch möglich",
--- a/ultros-frontend/ultros-app/locales/en.json
+++ b/ultros-frontend/ultros-app/locales/en.json
@@ -42,6 +42,7 @@
     "currency_exchange": "Currency Exchange",
     "currency_exchange_desc": "Spend tomestones, get gil",
     "datacenter": "Datacenter",
+    "region": "Region",
     "worlds": "Worlds",
     "vendor_available": "Vendor Available",
     "exchange_available": "Exchange Available",
--- a/ultros-frontend/ultros-app/locales/fr.json
+++ b/ultros-frontend/ultros-app/locales/fr.json
@@ -42,6 +42,7 @@
     "currency_exchange": "Échange de devises",
     "currency_exchange_desc": "Dépensez vos mémoquartz, gagnez des gils",
     "datacenter": "Centre serveur",
+    "region": "Région",
     "worlds": "Mondes",
     "vendor_available": "Disponible chez un PNJ",
     "exchange_available": "Échange possible",
--- a/ultros-frontend/ultros-app/locales/ja.json
+++ b/ultros-frontend/ultros-app/locales/ja.json
@@ -42,6 +42,7 @@
     "currency_exchange": "通貨交換",
     "currency_exchange_desc": "トームストーンを使ってギルを稼ぐ",
     "datacenter": "データセンター",
+    "region": "地域",
     "worlds": "ワールド",
     "vendor_available": "NPCで入手可",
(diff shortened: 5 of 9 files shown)


akarras:
Merged after browser verification on a live local build.

**The geometry claim is exact.** Measured on the item-explorer world picker: input box and collapsed overlay both at `top: 137, height: 41` → **overhang 0**. The PR predicted the old overlay overhung by 2px and the new one lands on the input's own text position; that is what the numbers say.

**Both defects fixed in the same component check out:**

- *Enter-after-open no longer changes your selection.* On focus, `aria-activedescendant` was `select-item-2` — the currently-chosen `Cerberus` — not `select-item-0` (`Europe`, which sorts first). Pressing Enter immediately left the value at `Cerberus`. That was the reported bug.
- *Keyboard highlight stays in view.* With 153 rows in a 383px panel, arrowing down to `select-item-18` (`offsetTop 720`) moved `scrollTop` 201 → 381, putting the row inside `[381, 764]`. The `top + height - view_height` branch is doing exactly what it should, and the panel scrolls rather than the page.

Kind markers render on both surfaces — dropdown rows read `Region|Europe`, `Datacenter|Chaos`, `World|Cerberus` via the `sr-only` label, and the collapsed field carries the pin. `WorldOnlyPicker` on `/flip-finder/Cerberus` correctly shows no icon and also measures overhang 0.

No console errors. Note the `select-item-N` element ids are page-global, so two `Select`s on one page would collide in `get_element_by_id` — pre-existing (`aria-activedescendant` already used that scheme), not introduced here, and there is only one picker per page today.

PR #1042, opened 2026-07-31
feat(charts): chart modes & icon toolbar — candles, range, density (spec 2 of 4)

akarras:
## Summary

Implements spec 2 of the chart revamp ([design](docs/superpowers/specs/2026-07-26-chart-modes-toolbar-design.md), [plan](docs/superpowers/plans/2026-07-31-chart-modes-toolbar.md)). Spec 1 (server-side aggregation) shipped in #990; this renders the OHLC/quantile columns that payload already carries and adds the one mode that needs its own data.

- **Candles**: bodies batched into up/down `Node::Path`s plus one wick path (2,000 candles → 3 nodes). Buckets with fewer than 3 sales render wick-only ticks ("range known, direction unknown"); flat prices keep a 1.2px body floor. Colorblind-safe teal/burnt-orange pair with a luminance-delta regression test (not red/green).
- **Range**: low–high and p25–p75 ribbons (closed-polygon paths) under a p50 median line, up to two series.
- **Density**: new `PriceDensity` wire type, `price_density`/`price_min_max` ClickHouse queries (same `window_predicate`, no FINAL, no join), `GET /api/v1/price_density/{world}/{itemid}` sharing the spec-1 cache (`CacheKey` grew a `bins` field) and `Cache-Control` plumbing, and a layout batching cells into ≤8 nodes via an 8-step indigo ramp.
- **Toolbar**: the three stacked chip rows collapse into one `ChartToolbar` — icon-only mode group, group-by dropdown chip with region/datacenter/world icons, overlays popover with count badge. Slots reserved for spec 3's view toggle and world filter. Every icon button carries an aria-label.
- **Caption line** replaces `StatsStrip`: mode · grouping · n sales · avg · median.
- Modes with a series cap suppress extra series from drawing (metadata/legend intact) and surface a hint; the quantity toggle disables in density with a reason. Mode switches never touch the time window or grouping. Ten new i18n keys translated in all seven locales.

Deviation from spec text: ribbons use a closed-polygon `Node::Path` (`band_path_d`) rather than `Node::Area` — `Area` only fills to a flat baseline and cannot express a band between two curves; same node budget.

## Test plan

- [x] `cargo test -p ultros-charts` (73) — candle batching/sparse/floor/cap, range bands+order+cap, density steps/coverage, greyscale + ramp-order theme tests, all pre-existing regression tests
- [x] `cargo test -p ultros-api-types` (3), `-p ultros-clickhouse --lib` (24), `-p ultros price_series` (7, incl. `density_bin_width` and cache-key change)
- [x] `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` clean
- [ ] `price_density_smoke.rs` compiles and skip-gates; **not run against a live ClickHouse** (no docker on the dev machine) — run with the throwaway-docker one-liner in the file header
- [ ] Browser pass over the four modes on a datacenter-scope item page (dev box had no ClickHouse, so the data path couldn't be exercised locally)


akarras:
Merged after closing both unchecked boxes in the test plan against a live local build.

**Browser pass over the four modes** (`/item/Cerberus/5111`, Iron Ore, 49 sales, desktop 1440px):

| mode | result |
|---|---|
| Price line | line + market-avg overlay, caption `Price line · 49 sales · market avg 72 · median 72` |
| Candlesticks | **exactly 3 `<path>` nodes** — one wick path (`stroke`, `fill:none`) plus `#5eead4` up bodies and `#c2410c` down bodies. The batching and the colourblind-safe teal/burnt-orange pair are both doing what the PR says. |
| Price range | ribbons render |
| Sale density | 5 batched paths on the indigo ramp `#1e1b4b → #312e81 → #3730a3 → #4338ca → #c7d2fe`, dark-to-light in count order |

Mode switches updated the caption and left the time window and grouping alone. `aria-pressed` tracked correctly across all four. The Overlays popover reported `Quantity` as `disabled: true` in density mode and enabled elsewhere, as specified.

**The ClickHouse path was exercised for real**, which also covers the `price_density_smoke.rs` box indirectly:

```
GET /api/v1/price_density/Cerberus/5111?price_bins=32
→ 200 {price_bins: 32, price_lo: 1, bin_width: 4.3125, bucket_seconds: 2592000, cells: 22}
```

`density_bin_width` produced a sane width off a real `price_min_max`, and the payload round-tripped into the 8-step layout without the `expect("non-empty")` in `build_price_density_chart` ever being reachable — it sits behind the `density.is_empty()` early return.

Console was clean apart from the known-benign AdSense `TagError: already have ads`, and `boot-progress` was absent, so hydration completed.

Also ran locally (CI does not run tests — the step is commented out in `rust.yml`): `cargo test -p ultros-charts` **71 passed**, `-p ultros-api-types` **141 passed**, `-p ultros-clickhouse --lib` **24 passed**. Merged cleanly alongside #1039/#1041/#1040 with no conflicts and all 7 locale files at an identical key count.

PR #1043, opened 2026-07-31
chore(ci): drop the nightly game-data workflow

akarras:
Follow-up to #1033. Game-data updates are agent-driven rather than scheduled, so the nightly cron is dead weight.

## Why not a cron

A game-data bump can break its consumers — a renamed sheet or column shifts `xiv-gen`'s generated types — so the regeneration and the fallout have to be resolved in the same change. `update_game_data.yml` can do the regeneration and open a PR; it can't do the second half. Left running, it queues up PRs that don't build.

## It was already dormant

Worth stating plainly so this doesn't read as removing something live:

- The workflow file was **not on `main`** before #1033 re-added it.
- GitHub still has the workflow record as **`disabled_manually`** (`gh api repos/akarras/ultros/actions/workflows`).
- Its last runs were **March 2026** — failing every night from 03-04, last success 03-03.

## What changes

- Deletes `.github/workflows/update_game_data.yml`.
- Rewrites the CLAUDE.md "Auto-updates" bullet, which would otherwise be actively false — it currently claims the workflow "runs daily … and opens a PR if anything changed." Replaced with the manual path and the checks a bump needs to pass (`--latest`, then `cargo check -p xiv-gen-db --features embed` and `cargo test -p game-data-pack`).

No code change; `data/`, the generator, and the build scripts are untouched.

## One loose end for later

The pack generator is deterministic by design — sorted inputs, tar gnu headers zeroed for mtime/uid/gid, `BTreeSet`, no `SystemTime`/`Utc::now()` — and `icons.rs` documents why (readdir order "was fine for a build artifact but not for a file committed to LFS"). But **nothing tests that a rebuild is byte-identical**. That property is what made the old workflow's `git diff --quiet data/` guard meaningful, and it's what any future automation would depend on. Worth a test if this ever gets automated again.


akarras:
Merged. Verified the "already dormant" claim independently before merging: `gh api repos/akarras/ultros/actions/workflows` reports `.github/workflows/update_game_data.yml` as **`disabled_manually`**, while every other workflow is `active`. So merging #1033 was never going to wake a cron, and removing the file changes nothing operationally — it just stops the CLAUDE.md bullet from being a lie.

The reasoning is right, too: a cron can regenerate packs but cannot resolve the consumer breakage a renamed sheet or column causes in `xiv-gen`'s generated types, so it would queue PRs that do not build.

Agreed on the loose end — the byte-identical-rebuild property is currently load-bearing by construction only (sorted inputs, zeroed tar headers, `BTreeSet`, no `SystemTime`), with nothing asserting it. Worth a test in `game-data-pack/tests/` before any future automation leans on `git diff --quiet data/` again.

PR #1044, opened 2026-07-31
feat(welcome): add the color palette picker as an optional step

akarras:
## What

Adds a fourth, optional step to the welcome flow — **Step 4 — Color palette (optional)** — letting a new user pick a brand palette before they land on the app.

## Why

The welcome flow already collects home world, price zone, and language, but nothing about how the app looks. The palette picker was only reachable from the settings page, which a new user has no particular reason to visit.

## How

The palette radiogroup is extracted out of `ThemePicker` into a shared `PalettePicker` component rather than duplicated:

- `PalettePicker` takes a `show_label` prop (default `true`). The settings page renders it with the label exactly as before, so that page is visually unchanged.
- The welcome step passes `show_label=false`, since its own step heading already names it.
- One palette list, so the two surfaces can't drift apart as palettes are added.

The step is optional in the same sense as the price zone and language steps: the Continue CTA still gates only on home world, so skipping it costs nothing.

Selections write through the app-root `ThemeSettings` context, so they apply instantly and persist to localStorage plus a cookie — the choice survives the navigation to `/`.

## i18n

`welcome_step_palette_label` and `welcome_step_palette_help` added to all seven locales (en, fr, de, ja, cn, ko, tc) with real translations, not English stubs. A native speaker's eye on the non-English strings would be welcome.

## Reviewer note — pre-existing gap, not addressed here

`ThemePicker` only ever listed 12 of the 18 variants in `ThemePalette`. `Violet` (the **default**), `Teal`, `Emerald`, `Amber`, `Rose`, and `Sky` have CSS in `style/tailwind.css` but no button anywhere in the UI.

This PR inherits that list as-is rather than expanding scope, but it's more visible now: a fresh user arriving at the welcome page sees *no* button highlighted, because their current palette (`Violet`) isn't among the choices. Worth deciding separately whether to add the six missing buttons or drop the unused variants from the enum — happy to do either.

## Testing

`./check_ci.sh` passes clean — `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings`, exit 0, no warnings. Not exercised in a running browser.

### DIFF
--- a/ultros-frontend/ultros-app/locales/cn.json
+++ b/ultros-frontend/ultros-app/locales/cn.json
@@ -173,6 +173,8 @@
     "welcome_step_price_zone_help": "当您浏览物品时，这是我们将显示价格的服务器、数据中心或大区。大多数人会保留为其数据中心以获得最广泛的市场视图。",
     "welcome_step_language_label": "第 3 步 — 语言（可选）",
     "welcome_step_language_help": "选择您偏好的语言。您可以随时通过顶部导航栏进行切换。",
+    "welcome_step_palette_label": "第 4 步 — 配色方案（可选）",
+    "welcome_step_palette_help": "选择 Ultros 使用的强调色。仅影响外观，保存在您的浏览器中，可随时在设置页面更改。",
     "welcome_home_world_set_with_name": "主服务器已设置为 {{world}}。一切就绪。",
     "welcome_continue_cta": "前往 Ultros",
     "welcome_skip_for_now": "暂时跳过",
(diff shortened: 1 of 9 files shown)


akarras:
Reviewed. The extraction is the right shape — one palette list, `show_label` keeps the settings page byte-identical, and swapping `aria-labelledby="theme-palette-label"` for `aria-label` is a real improvement now that the component can appear twice on a page (the old form would have produced a duplicate id).

i18n checked directly rather than taken on trust: `welcome_step_palette_label` and `welcome_step_palette_help` are present in all seven locales with genuine translations (`第 4 步 — 配色方案（可选）`, `Étape 4 — Palette de couleurs (optionnel)`, `4단계 — 색상 팔레트 (선택)`, …), and every locale file ends at an identical key count.

**Fixed the gap you flagged rather than deferring it (6d4d401).** You were right that it becomes serious on this page specifically: `Violet` is `ThemePalette::default()`, so a first-time visitor — the entire audience for the welcome flow — met a radiogroup with nothing checked. All six missing variants (Violet, Teal, Emerald, Amber, Rose, Sky) already have `[data-palette="…"]` blocks in `style/tailwind.css`, so they were purely unreachable from the UI, not unimplemented. Added all six, plain-colour ones first so the default sits at the head of the row. That also fixes the same latent hole on the settings page.

I chose adding over dropping the enum variants because a persisted `theme_palette` cookie may already hold one of them — removing variants would be a migration, adding buttons is not.

Merging.

PR #1045, opened 2026-07-31
fix(side-nav): highlight the tool you're actually on

akarras:
The sidebar never highlighted the current page. Two independent bugs, both of which had to be fixed for anything to show.

## 1. The active state was never detected

`<A>` sets `aria-current="page"` only when the whole resolved href equals the current path, but the tool links are world-aware — `/flip-finder/{homeworld}`, `/scrip-sources?world=…`. So:

- viewing a world other than your homeworld missed,
- the `?world=` links could never match at all (the query string is part of the href but not of `location.pathname`),
- sub-routes like `/items/category/5` or `/list/42` missed.

Replaced them with a `SideNavItem` that owns a *section* — the first path segment — and sets `aria-current` from `use_location()`. That's what "which tool am I in" actually means here.

It renders a plain `<a>` rather than `<A>` on purpose: the router's global anchor-click handler keeps navigation client-side either way, and this leaves exactly one writer for `aria-current`. Spreading `attr:aria-current` onto `<A>` would have raced with the component's own reactive effect on the same attribute. The brand link at the top still uses `<A>`.

## 2. Even when detected, the CSS lost

`a:not(.nav-link):not(.btn):not(.btn-primary):not(.btn-secondary):not(.btn-ghost)` scores (0,5,1) and sets `background-color: transparent`, outweighing `.side-nav-item[aria-current="page"]` at (0,2,0). So `/alerts` — the one case where matching worked — still rendered flat.

`.side-nav-item` now opts out of those generic link rules the way `.nav-link` already did. Active styling is a brand tint, brand text/icon color, and an inset left accent bar so the current tool is still identifiable when the sidebar is collapsed to icons.

## Reviewer note

The opt-out has a side effect: inactive sidebar items now render in `--color-text` instead of `--brand-fg`, and no longer underline on hover. That's what the `.side-nav-item` rule always specified — it was just being overridden — and it gives the active item more contrast to stand out against, but it does slightly change the resting look of the sidebar. Easy to revert that bit alone if you'd rather keep the old tint.

## Testing

`./check_ci.sh` passes (fmt clean, clippy clean across the workspace). Verified by compiling, not visually — a visual check needs a full wasm build plus Postgres and ClickHouse.


akarras:
Reviewed. Both halves of the diagnosis check out, and the CSS half is more clearly correct than the PR claims.

**On the "side effect" you flagged** — inactive items dropping from `--brand-fg` to `--color-text` and losing hover underline — that is not a change of intent, it is the component finally getting what it always asked for. `.side-nav-item` has declared `color: var(--color-text)` and `text-decoration: none` since it was written (`style/tailwind.css:1811-1812`); the generic `a:not(…)` rule at (0,5,1) was simply outscoring both. So the opt-out restores the stated design rather than introducing a new one. I would keep it.

**On the detection half**: matching the first path segment is the right unit. `<A>`'s whole-href comparison could never have matched `/scrip-sources?world=…` (the query string is in the href but not in `location.pathname`), and world-bearing hrefs only matched when you were on your own home world. Using `use_location()` with one writer for `aria-current` also avoids racing `attr:aria-current` against the component's own effect, which is the right call.

**Merge conflict with #1040 resolved correctly.** Both branches add a `:not()` to the same four generic-anchor rules; I resolved it independently in a scratch worktree and got a byte-equivalent result to your `7f800d78` — `:not(.side-nav-item)` and `:not(.card-link)` both kept in all four. Also confirmed the merged branch diffs against `origin/main` in exactly two files, so nothing newer got reverted on the way through.

Required CI green on the merged head. Merging.

Not separately eyeballed in a browser — I lost the build budget to an unrelated toolchain fight — but the risk is confined to taste, since the failure mode of the one thing I could not check statically (plain `<a>` vs `<A>` for client-side nav) is a full page reload, not a broken link.

PR #1046, opened 2026-07-31
feat(charts): comparison grid — small multiples, shared crosshair, world filter (spec 3 of 4)

akarras:
## Summary

Implements spec 3 of the chart revamp ([design](docs/superpowers/specs/2026-07-26-chart-comparison-grid-design.md), [plan](docs/superpowers/plans/2026-07-31-chart-comparison-grid.md)), on top of #1042's modes and toolbar.

- **Grid view**: one small chart per series of the current grouping (Price, Candles, or Range marks — no axes/volume at cell size), filling the toolbar slot #1042 reserved (`LuChartNoAxesCombined` / `LuLayoutGrid`). Cells share a y-domain by default (per-cell scaling as the escape hatch), sort by name or by window change, and cap at 24 with a "+N more" affordance that opens the world filter.
- **Shared crosshair**: a `UnionIndex` (sorted distinct bucket timestamps, each series mapped by position, gaps as `None`) is built once per payload; pointer x resolves once against the shared per-cell x positions and one `hover_index` signal drives every cell's crosshair plus a single container-level tooltip listing every series' value. Exact by construction — one query, one bucket width, no snapping.
- **World filter**: searchable, datacenter-grouped multi-select chip (`TbFilterOutline`) with select-all/none per group and a `visible/total` count badge. It writes the **same `hidden_series` signal as the legend**, so the two can never disagree, and it's purely client-side visibility — never a refetch. The filter lists names at the *current grouping level* (worlds when grouped by world, DCs when grouped by datacenter), since hiding a name the legend doesn't know would silently do nothing.
- **Index to % change**: overlay toggle rebasing each series to 0% at its first plotted bucket; percent y-axis, market average/trendline/raw dots suppressed with a reason, caption reflects it.
- The mode-cap hint from #1042 now offers **Switch to grid** as its action. 16 new i18n keys in all seven locales.
- Also fixes `scene_view` to `impl IntoView + use<>` — its edition-2024 RPIT captured the `&Scene` borrow it doesn't actually hold.

**Deviations from the spec, flagged for review:**
- **Grid is disabled (with reason) in Density mode.** The spec offers grid in every mode on the premise that spec 1's per-series payload contains everything needed — that's false for density, whose payload is scope-wide. A per-world density grid would need N fetches, violating the spec's own "no new fetching" non-goal.
- **% change applies to overlay Price view only** (the spec specifies overlay-only; rebasing candle OHLC is deferred).

## Test plan

- [x] `cargo test -p ultros-charts` (88+12) — union-index exactness/subset/empty, grid cells/cap/domain/sort/marks/color-parity, % rebase + suppression + label format, plus every spec-1/2 regression test
- [x] `cargo fmt --all -- --check` and `cargo clippy --all-targets -j 16 -- -D warnings` clean after rebasing onto main (post-#1042/#1033)
- [ ] Browser pass (overlay↔grid state preservation, crosshair alignment, filter↔legend sync) — dev box has no ClickHouse; same limitation noted on #1042


akarras:
Code review done; holding the merge for the browser pass your own test plan leaves unchecked. Notes below so nothing is lost either way.

**`union_index.rs` is the right primitive and is implemented cleanly.** Sorted + deduped timestamps, one merge pass per series rather than a lookup per point, `Vec`-backed throughout so there is no HashMap iteration order reaching the DOM — which matters on this codebase specifically (see the `related_items.rs` SSR/CSR-order bug behind #960). The "exact by construction" claim holds: one query, one bucket width, positions matched on equality with no snapping. The one unstated precondition is that each series' buckets are ascending — you document it and the server guarantees it, but a violation would silently produce wrong `positions` rather than an error. Worth an `debug_assert` if it is cheap.

**`grid.rs` colour parity is done the subtle way and it is right.** `.enumerate()` before `.filter()` means the palette index comes from the full sorted list, so hiding a series does not reshuffle the remaining cells' colours — which is what keeps the grid and the legend agreeing. The `union.is_empty()` early return correctly guards both `.expect("non-empty")` calls below it. `sort_by(name)` before the optional change-sort keeps the default order deterministic across SSR/CSR.

**`scene_view(&Scene) -> impl IntoView + use<>`** is the correct fix — edition-2024 RPIT captures the `&Scene` lifetime by default, and the body does clone everything it needs.

**Deviations look right, not like corner-cutting.** Disabling grid in density mode because the payload is scope-wide rather than per-series is a real constraint, not a shortcut — a per-world density grid would need N fetches, which is exactly the non-goal the spec set. Same for deferring OHLC rebasing.

**i18n verified rather than assumed**: all 16 keys present in all seven locales with real translations, and every locale file lands on an identical key count after merging alongside #1044.

**What I could not check, and why it is worth checking**: crosshair alignment across cells, overlay↔grid state preservation, and filter↔legend sync are interaction properties that unit tests cannot establish. I lost this run's build budget to an unrelated Windows toolchain fight (vendored OpenSSL hitting MAX_PATH from a deep worktree, then a stalled rebuild). Everything ahead of you in the queue is merged and main is green, so this is next up.

akarras:
Browser pass done after all — the build finally came up. **All three unchecked items in your test plan verified**, on `/item/Chaos/5111` (Chaos DC, 8 worlds, 229 sales, world grouping).

**Grid renders the small multiples correctly.** 8 cells at 256×137, one per world, sorted alphabetically (Cerberus, Louisoix, Moogle, Omega / Phantom, Ragnarok, Sagittarius, Spriggan), each with its colour dot. Colour parity with the legend is exact — the `.enumerate()`-before-`.filter()` detail does what it is meant to. Toolbar shows the view toggle, the `Worlds 8/8` filter chip, `Sort by name`, and `Scale each cell`. Caption reads `Price line · Grid · grouped by World · 229 sales · market avg 106 · median 99`.

**Shared crosshair is exact, which is the headline claim.** A single `pointermove` at 60% across cell 0 put the crosshair line at **x = 220 in all eight cells** — one distinct value across the set, not eight near-misses. The `UnionIndex` approach delivers what it promises.

**Filter ↔ legend really are one signal.** Unchecking Omega in the world filter simultaneously: dropped the grid to 7 cells, moved the chip badge to `Worlds 7/8`, and restyled the legend entry to `opacity-40 line-through` (computed `opacity: 0.4`, `text-decoration-line: line-through`). Re-checking restored all three. They cannot disagree because they are the same `hidden_series`.

**Overlay ↔ grid preserves state.** With Ragnarok hidden: Grid → Overlay → Grid kept the filter at 7/8, the grouping, the timeline selection, and returned 7 cells. The caption correctly drops the `Grid` token in overlay view and recomputes `n` from visible series only (229 → 216), which is the right choice.

Hydration clean (`boot-progress` absent). Console had only the known-benign AdSense `TagError`, plus one unrelated pre-existing failure — see below.

**Unrelated pre-existing bug found while here, not yours:** `/api/v1/item_stats/{scope}/{item}` returns **400 for datacenter and region scopes** and 200 only for a world (`/item_stats/Chaos/5111` → 400, `/item_stats/Europe/5111` → 400, `/item_stats/Cerberus/5111` → 200). So every DC- or region-scoped item page fires a failing request. Filing separately.

Merging.

PR #1047, opened 2026-07-31
feat(charts): patch milestone bands and game-history calendar (spec 4 of 4)

akarras:
## Summary

Implements spec 4 of the chart revamp ([design](docs/superpowers/specs/2026-07-26-game-history-milestones-design.md), [plan](docs/superpowers/plans/2026-07-31-patch-milestones.md)) — the final phase. Independent of #1046 (both branch from main; small merge overlap in the toolbar/caption whichever lands second).

- **Seed calendar** in `ultros-api-types::game_history`: checked-in table of patch release dates (append-only historical facts — can become incomplete, never wrong). Global track seeded 2.0–7.3 majors plus Dawntrail point patches; **CN/KR ship empty** until sourced carefully, since an empty track draws nothing while a wrong date misattributes price moves. Region→track maps as data with a Global fallback for regions Universalis invents tomorrow.
- **Bands**: rendered behind everything (first scene nodes) — hue per expansion, alternating lightness between consecutive patches, a single edge line at expansion boundaries, version labels ("7.2") on bands ≥48px. Level-of-detail follows zoom: >2y expansions only, 6mo–2y majors, 30d–6mo point patches, <30d nothing. Density mode degrades to boundary lines only (a tint under a sequential ramp would misread as data).
- **Multi-track rule**: when Region-grouped series span different patch schedules (e.g. North-America + 中国), milestones turn off and the caption says why — picking a winner would silently mislabel half the chart.
- **`GET /api/v1/game-history`** (`?track=` filter, day-long `Cache-Control`) for external consumers; the WASM chart reads the seed const directly, no fetch.
- Toggle ("Patch bands", default on) in the overlays popover; two new i18n keys in all seven locales.

**Deviations from the spec (its premises changed under it):**
- **Expansion-name labels deferred.** The spec's "ExVersion costs nothing" predates #1033: the rkyv packs are now prebuilt LFS artifacts whose shape depends on enabled xiv-gen features, so `ex_version` needs a `game-data-pack` sheet addition + regenerated packs — a follow-up PR. Hue-coded eras + patch numbers carry the information meanwhile. Same deferral applies to the spec's `PatchMark` cross-check test.
- **Coverage-start milestone deferred.** "Earliest sale for the scope's worlds at all" is an unfiltered `min(sold_date)` off the ClickHouse primary-key prefix — the exact query shape this repo bans; needs a rollup design. "First recorded sale" already exists implicitly: the chart starts at the first data bucket.
- The seed's dates are hand-entered from well-known release history; **please eyeball them in review** (the spec itself asks for this). Appending 7.35+/7.4 and sourcing CN/KR are one-line follow-ups.

## Test plan

- [x] `cargo test -p ultros-api-types` (track mapping + fallback, seed sorted/unique/consistent, LOD tiers, label format, serde) and `-p ultros-charts` (75: bands tile without gaps/overlaps, draw behind all data, alternate lightness within an expansion and re-hue across, wide-band labels, density = lines + zero rects, empty-milestones regression)
- [x] `cargo fmt --all -- --check`, `cargo clippy --all-targets -j 16 -- -D warnings` clean
- [ ] Browser pass deferred — no ClickHouse on the dev box (same note as #1042/#1046)


akarras:
Reviewed the seed calendar you asked to have eyeballed. Two findings, one of which I think blocks the feature from doing anything useful today.

**1. The seed is ~12 months stale, which makes the bands a no-op on the default view.** The last row is `(730, (2025, 8, 5))` and today is 2026-07-31. On the chart window I get by default on a live build (roughly 2025-12 → 2026-06), `visible_patches` returns nothing inside the window, so the tiling falls back to the single leading band from the latest pre-window patch — one flat tint across the whole plot, no boundaries, no labels. The bands only start carrying information once you zoom back past 2025-08.

This is the "incomplete, never wrong" design working as intended, so it is not a correctness bug — but it does mean the feature ships invisible on the view most people see. Appending 7.31/7.35/7.4/7.5 (and 8.0 if it has landed) before merge would be the difference between shipping the feature and shipping the plumbing.

**2. `(250, (2015, 2, 24))` looks like 2.51's date, not 2.5's.** Patch 2.5 "Before the Fall" released **2015-01-20**; 2015-02-24 is 2.51, the Manderville Gold Saucer patch. Worth re-checking that row specifically. Every other row I could check reads correct — 2.0 2013-08-27, 3.0 2015-06-23, 4.0 2017-06-20, 5.0 2019-07-02, 6.0 2021-12-07, 6.55 2024-01-16, 7.0 2024-07-02, 7.2 2025-03-25 all match. 2.55 (2015-03-31) is absent, which is consistent with the "majors plus DT point patches" scope you describe.

Low blast radius either way — Universalis history does not reach 2015 — but a wrong date in a table whose whole premise is "can be incomplete, never wrong" is worth correcting at the source.

**On the rest**, the design decisions read well: region→track as data with a Global fallback, empty CN/KR rather than guessed dates, multi-track → off with a stated reason rather than picking a winner, and density degrading to boundary lines so a tint never competes with a sequential ramp. The `ex_version = version / 100 - 2` consistency assert and the sorted/unique seed test are the right guards for a hand-maintained table.

Not merging this run — I ran out of build budget for the browser pass, and #1046 (spec 3) is still open ahead of it. Holding for a follow-up once the seed is topped up.

akarras:
Seed completed through 7.55 with per-track sourcing (928be070):

- **Global** — verified against the [consolegameswiki patch archive](https://ffxiv.consolegameswiki.com/wiki/Patches): fixed 2.5 (2015-01-20 — the previously entered 2015-02-24 was 2.51's date), backfilled all x.x5 point patches, appended 7.35→7.55 (2026-07-28). One extraction error was caught by ordering sanity (2.35 is 2014-08-18, not April) — the seed's sorted-order test guards this class.
- **CN** — expansion launches 2.0→7.0 from the [zh-wikipedia version table](https://zh.wikipedia.org/zh-cn/%E6%9C%80%E7%BB%88%E5%B9%BB%E6%83%B3XIV%EF%BC%9A%E9%87%8D%E7%94%9F%E4%B9%8B%E5%A2%83) + official sdo.com version pages. 2.0 uses the 2014-08-25 open-test date (when the CN economy went live). Sub-patch cadence is CN-specific; the 盛趣 notice archive / [huijiwiki 版本时间表](https://ff14.huijiwiki.com/wiki/%E7%89%88%E6%9C%AC%E6%97%B6%E9%97%B4%E8%A1%A8) (Cloudflare-gated to bots, fine manually) is the source when someone wants patch-level CN granularity.
- **KR** — expansion launches 3.0→7.0 plus 7.5 from ff14.co.kr notices via press coverage ([5.0](https://www.inven.co.kr/webzine/news/?news=227896&site=ff14), [6.0](https://www.inven.co.kr/webzine/news/?news=268213), [7.5 simultaneous-update announcement](https://www.gamemeca.com/view.php?gid=1772686)). **KR runs simultaneously with Global from 7.5 (2026-04-28)** — future KR rows just mirror Global's.

New invariant test: no track may skip a middle expansion (leading gaps stay legal) — the one incompleteness shape that would mislabel eras.

akarras:
## Review: approve (code + data verified)

### The date audit you asked for — done, and the seed is clean

You asked reviewers to eyeball the hand-entered dates. I did, and started from a suspicious signal: **every pre-7.0 `x.x5` point patch in the seed falls on a Monday, while every Dawntrail-era one falls on a Tuesday.** 23 Mondays against 52 Tuesdays looked like a systematic off-by-one, and I was about to push a +1-day fix to all 22 rows.

**Don't — they're right.** I checked against the module's own cited source (ffxiv.consolegameswiki.com/wiki/Patches) and it returns those exact dates, Mondays included: 2.35→18 Aug 2014, 2.45→8 Dec 2014, 2.55→30 Mar 2015, 3.25→28 Mar 2016, 4.15→20 Nov 2017, 5.05→29 Jul 2019, 5.55→24 May 2021, 6.05→3 Jan 2022, 6.45→17 Jul 2023 — all matching the seed character for character. The Monday/Tuesday split is a real property of SE's release history, not a data-entry artifact. Recording it here so nobody "corrects" those 22 rows later.

Also verified exact: all six expansion launches (2.0/3.0/4.0/5.0/6.0/7.0), and the whole recent tail — 7.3 (2025-08-05), 7.35 (2025-10-07), 7.4 (2025-12-16), 7.45 (2026-03-03), 7.5 (2026-04-28), 7.55 (2026-07-28). Spot-checked CN independently: 7.0 金曦之遗辉 = 2024-09-27 ✅, 6.0 晓月之终途 = 2022-03-16 ✅.

### ⚠️ The PR description no longer matches the code

The body says *"CN/KR ship empty until sourced carefully"* and *"Global track seeded 2.0–7.3 majors plus Dawntrail point patches."* Commit `928be070` landed after you wrote that, and the code now has **6 CN rows, 6 KR rows, and Global through 7.55** including point patches back to 2.35. That inverts the review instruction — it tells a reviewer there are no CN/KR dates to check when there are twelve. Worth correcting in the squash message so the merged history isn't misleading.

### Code

- `WebError::BadRequest` → HTTP 400, so the new endpoint doesn't repeat the #976 "errors answered 200" shape. Unknown `?track=` correctly rejects; `track: Option<String>` sidesteps the `Query` bool trap from #983.
- `Theme::base()` is the sole constructor and always populates 6 hues, so `expansion_hues[ex % len.max(1)]` can't index an empty vec. (`.max(1)` guards the modulo but not the index — it's safe only because `base()` is the single path. Fine as-is, just fragile if a future theme is built field-wise.)
- `selected_domain` falls back to `available_domain` when unzoomed, so bands render by default rather than only after a zoom — worth stating since it wasn't obvious from the diff.
- The `specs.clear()` "only the latest pre-window patch survives" loop is correct given `visible_patches` returns date-sorted rows.
- Caption reads `show_patches`/`milestone_track` inside the existing reactive closure, so the mixed-track reason updates live.

**Tests: 243 passed / 0 failed** here (`-p ultros-api-types -p ultros-charts`, 149 + 92 + 2). `empty_milestones_render_identically_to_before` plus the zero-deletion diff makes the regression surface genuinely nil.

### Nit

`disabled=Signal::derive(|| false)` / `disabled_reason=Signal::derive(String::new)` — `OverlayRow` takes both as `#[prop(into)]`, so `disabled=false` and `disabled_reason=""` work and drop two closures.

akarras:
## Browser pass — done, and it passes

The test plan's third box was left unchecked with *"no ClickHouse on the dev box."* **ClickHouse is up on this machine** (8123/9000 listening, creds in `.env`), so I ran the pass against real data. Built and served this branch locally (`bin-features=[]` for the MSVC/jemalloc wall), footer confirms `Version: afa8d37a`.

**`/item/Cerberus/2` (Fire Shard), window 2026-01-07 → 2026-06-06** — a ~5-month span, so the `Point` LOD tier. Read straight off the rendered SVG:

| Spec claim | Observed |
|---|---|
| Bands tile the window, no gaps/overlaps | spans `68→389.2`, `389.2→716.2`, `716.2→944`; gaps `[0, 0]` |
| Clamp to plot area | starts at plot-left `68`, ends at plot-right `944` (= 960−16) |
| Alternating lightness between consecutive patches | alpha `0.050 → 0.090 → 0.050` |
| Hue per expansion | all `rgba(52,211,153,…)` = `#34d399`, the DT hue — correct, all three are 7.x |
| Drawn behind everything | band rects at node indices 1/3/5; first data path at index 19 |
| Labels on bands ≥48px | `7.4`, `7.45`, `7.5` on bands 321/327/228 px wide |
| Leading band from the latest pre-window patch | 7.4 (2025-12-16) tints the run-up, as designed |
| LOD picks the right tier | 7.45 is an `x.x5` point patch and appears — correct for a 5-month span |

**Toggle:** 3 rects + 3 labels → **off** → 0 rects + 0 labels → **on** → 3 + 3 restored. Default is on, as specified.

**`GET /api/v1/game-history`:** `cache-control: public, max-age=86400` ✅, `?track=korea` → 6 rows ✅, `?track=bogus` → **400** ✅ (not the 200-with-error-body shape #976 fixed).

`read_console_messages{onlyErrors:true}` returns nothing and `boot-progress` is absent, so hydration completed clean with the bands present.

Nothing left on the test plan. The only thing I'd still change before merge is the stale description (CN/KR are not empty; Global runs to 7.55).

akarras:
## Review

Reviewed against `origin/main` (branches from `05bb5dba`, zero commits behind).

### ⚠️ Blocking cross-PR conflict with #1050 — git merges it clean, then it doesn't compile

This is the important finding, and it is invisible to CI.

- This PR adds a `milestones` field to `DensityChartOptions`.
- #1050 adds `ultros-charts/src/charts/snapshot_tests.rs`, which builds a **exhaustive** `DensityChartOptions { .. }` literal.

Different files, so git auto-merges both with **zero conflicts**. The merged tree then fails:

```
error[E0063]: missing field `milestones` in initializer of `charts::price_density::DensityChartOptions`
   --> ultros-frontend/ultros-charts/src/charts/snapshot_tests.rs:141:10
```

**CI would not catch this** — `cargo test` is commented out in `rust.yml`, so `main` would go green with a broken test build. Whichever of the two lands second breaks it.

Fix applied on this branch (one line, matching the `web_options` helper directly above it in the same file, which already uses this form):

```rust
 &DensityChartOptions {
     width: 960.0,
     height: 540.0,
     utc_offset_minutes: 0,
     theme: Theme::site(),
+    ..Default::default()
 },
```

Using `..Default::default()` rather than `milestones: Vec::new()` so the next field added to the struct doesn't break the snapshot test again.

Verified after the fix, with all four open PRs merged together:

| Gate | Result |
|---|---|
| `cargo test -p ultros-charts` | **101 passed / 0 failed** |
| `cargo test -p ultros-app --lib` | 371 passed |

The five SVG snapshots are **byte-identical** with milestone rendering added — confirming the empty-milestones path adds no scene nodes, which is the regression this PR's own `empty-milestones` test asserts.

### Seed calendar — eyeballed as requested

Region mapping checked against the live `region` table rather than assumed:

```
1|Europe  2|Japan  3|NA-Cloud-DC  4|North-America  5|Oceania  6|中国  7|한국  8|繁中服
```

`track_for_region`'s `"中国"` / `"한국"` literals match byte-for-byte. Good — this is the failure mode that killed `/currency-exchange` for non-EN locales (#942), and keying on the region *name* is only safe because these come from Universalis rather than from localized game data.

Expansion launch dates all check out (2.0 2013-08-27, 3.0 2015-06-23, 4.0 2017-06-20, 5.0 2019-07-02, 6.0 2021-12-07, 7.0 2024-07-02). I deliberately did **not** "correct" the Monday `x.x5` dates — pre-7.0 point patches genuinely shipped on Mondays, and a day-of-week invariant is not a valid check against this data.

`version_label` / `mark_tier` / `ex_version` derivation all verified by hand against the `PatchMark` convention (`705` → `"7.05"`, `715` excluded from the Major tier, `700` → ex 5).

### Non-blocking notes

1. **Region 8 `繁中服` falls back to Global.** The Traditional-Chinese service runs its own schedule and launched years after ARR, so a TC-scoped chart will draw Global-dated bands. It degrades gracefully (only bands inside the data window draw, and TC data is recent), and the Global fallback is the documented, deliberate default — but it is a third real track, worth appending when sourced.
2. **The PR description is stale relative to the code.** It says "CN/KR ship empty" and that "appending 7.35+/7.4" is a follow-up; the merged seed actually populates `CHINA_SEED`/`KOREA_SEED` at expansion granularity and runs Global through 7.55. The code is better than the description — just noting so the description isn't taken as the spec later.

Approving with the compile fix folded in.

PR #1048, opened 2026-07-31
fix(item-explorer): key job routes on a locale-independent acronym

akarras:
## The bug

`ClassJob::abbreviation` is a **localized display string**. `ClassJobCategory`'s columns — the flags `job_category_lookup` matches on — are **English acronyms baked into the sheet schema**, identical in every locale. Four call sites used the former as a key for the latter.

German renames 23 of 36 visible jobs, French 22. For those visitors:

1. **The Job Sets nav minted dead links.** A German client's Warrior chip pointed at `/items/jobset/KRG`. `job_category_lookup` rejects `"KRG"`, and English SSR can't resolve it either — so the page rendered zero items on *both* sides of hydration. Every renamed job's own nav link was a dead page.
2. **Role grouping collapsed.** `role_for_job_abbr(&job.abbreviation)` bucketed those same jobs into `RoleGroup::Other`, flattening the popover's role sections. The existing `every_visible_job_maps_to_a_role` test passed only because the test process loads English data.

This is the sibling of #1001, which fixed `/items/category/:category` the same way. The mechanism is documented in the `resolve_jobset_param` doc comment.

## The fix

`canonical_job_acronym(ClassJobId) -> Option<&'static str>` — a locale-independent id→acronym table. Routes, hrefs, and role buckets key on it; **display labels stay localized** (a German player still sees "KRG" on the chip and in the heading, the link just goes to `/items/jobset/WAR`).

The legacy localized-abbreviation / full-name fallback is kept, carrying the identical documented locale caveat that `resolve_category_param` already accepts for legacy category links. Canonical hrefs now consolidate duplicates the way `CategoryItems` does.

## Verification

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy --all-targets -- -D warnings` (full workspace) | exit 0, 18m33s |
| `cargo test -p ultros-app --lib` | **368 passed / 0 failed** (base 363 + 5 added, 0 removed) |

**Genuine red→green.** With the pre-fix behaviour temporarily restored, exactly the two new locale tests fail, at exit 101:

```
De: job "KRG" (Krieger) has no role mapping
English SSR cannot resolve KRG, minted by a De client
```

Both pass with the fix. The tests drive the **real `data/xiv-db` locale packs** via `data_for(Language::De)` etc., not a mock — the LFS packs ship all seven locales, so this class of bug is now testable without a browser.

The other three added tests are pins/characterization rather than regression proofs, and are labelled as such: `canonical_acronym_matches_english_abbreviation` (table vs. shipped data), `only_beastmaster_lacks_a_category_column`, `a_localized_job_abbreviation_is_not_a_usable_category_key` (documents the mechanism; `collect_job_items_sorted` is unchanged, so it is green either way).

## Incidental finding — not fixed here

**`/items/jobset/BST` renders empty in every locale, English included.** Beastmaster (job id 43) ships in `ClassJob`, but `ClassJobCategory` in the current pack has only 43 bool columns ending at `PCT` — there is no BST column, so no category row can select its gear. That is a gap in the game-data pack, not a keying bug, and closing it means a pack regeneration plus an `xiv-gen` schema change (its own PR per `CLAUDE.md`). Pinned by `only_beastmaster_lacks_a_category_column` so it stays visible instead of being silently absorbed.

Also left alone deliberately: item *sets* are not compared across locales in the tests, because the CN/KO/TC packs track an older game version and genuinely hold different items. That is a separate and much broader SSR-locale issue a route key can't address.

## GlitchTip triage: none

**The GlitchTip MCP server is still unregistered** — no `glitchtip_*` tools are exposed and there is no `.mcp.json` at the worktree root or the main clone root. That is the fourth consecutive run with the backlog unreachable, so zero issues were triaged, ignored, or resolved. Re-adding the server config (with the token as an env var, per the #991 sanitization) is needed before triage can resume. This fix came from the code-hunt fallback instead.

### DIFF
--- a/ultros-frontend/ultros-app/src/routes/item_explorer.rs
+++ b/ultros-frontend/ultros-app/src/routes/item_explorer.rs
@@ -15,6 +15,7 @@ use crate::components::{add_to_list::*, cheapest_price::*, item_icon::*, meta::*
 use crate::global_state::xiv_data::tracked_data;
 use crate::i18n::*;
 use crate::routes::item_explorer_scope::{ExplorerPriceScope, use_explorer_price_scope};
+use crate::routes::item_explorer_toolbar::jobset_display_label;
 use icondata as i;
 use itertools::Itertools;
 use leptos::prelude::*;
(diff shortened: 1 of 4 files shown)


akarras:
## Review: approve

Verified locally, not just read.

**Correctness — the fix is right and the mechanism check holds up.** I confirmed `job_category_lookup` destructures exactly the columns `adv`..`pct` (43 of them, no `bst`), so the claim that ids 0–42 line up with the sheet's column order is literally true, and `canonical_job_acronym`'s 44-entry table matches. Keying routes/hrefs/role buckets on the id while leaving display labels localized is the same shape as #1001, and `tracked_data()` + `DataRevision` means the localized heading re-renders after the client's locale swap — so this follows an already-shipped precedent rather than inventing a new one.

**Tests: 368 passed / 0 failed**, run here (`cargo test -p ultros-app --lib`), matching the PR's claim exactly. `every_job_chip_link_resolves_to_items_in_every_locale` passes.

**No behavioural regression for unresolvable params.** `JobItems`/`JobSetDetail` now return `None`/`vec![]` where they previously fell through to `unwrap_or(decoded)`. That old path fed a garbage string into `job_category_lookup`, which matched nothing — so both old and new render an empty set. Equivalent, and now explicit.

### Follow-ups (not blocking, deliberately out of scope here)

- **`ultros/src/search_service.rs:95`** still mints `/items/jobset/{job.name}` — the English full name. That resolves on the English SSR side and *not* on a German client's, so a search-result click still diverges across hydration. Pre-existing and correctly acknowledged in your `resolve_jobset_param` doc comment; fixing it means lifting the acronym table into a crate the server can see, which is its own PR.
- **`ultros/src/web/sitemap.rs:187`** iterates every `class_job`, so the sitemap advertises `/items/jobset/BST` — the page this PR proves is permanently empty. Worth filtering once the BST gap is closed or excluded.

Sitemap keying is already correct (`class.abbreviation` under the English-only server = the canonical acronym), so no change needed there.

akarras:
## Review: approved ✅

Reviewed against `origin/main` (branches from `05bb5dba`, zero commits behind — no stale-base issue).

**The diagnosis holds up.** I independently confirmed the mechanism: `ClassJobCategory`'s per-job columns are English acronyms baked into the sheet schema, while `ClassJob::abbreviation` is localized, so using the latter as a key for the former can only resolve under English data. That is the same defect class as #1001 and the `/currency-exchange` crash, and keying on `ClassJobId` is the right fix.

**Verified locally** (CI does not run `cargo test` on this repo, so this needed a local run):

| Gate | Result |
|---|---|
| `cargo test -p ultros-app --lib` | **371 passed / 0 failed** (integration branch, incl. #1047/#1049/#1050) |
| Merge into `main` + the other three open PRs | clean, no conflicts |

Specific things I checked rather than took on faith:

- `canonical_job_acronym`'s 44-entry table is index-keyed, and `canonical_acronym_matches_english_abbreviation` pins it against the shipped English pack — so a game-data bump that renumbers a job fails loudly instead of silently emptying a route. Good guard rail.
- `resolve_jobset_param` and `jobset_display_label` both resolve via a **unique** match, so the `HashMap` iteration order never reaches the DOM. That matters here — unordered `HashMap` iteration reaching the DOM is exactly what caused the item-page hydration crash in #960.
- The tests genuinely drive `data_for(Language::De)` against the real LFS packs, not a mock. That is a materially stronger proof than a browser pass would have been, since a browser can only exercise one locale at a time.
- Display labels stay localized while only the *key* is canonicalised — a German player still reads "KRG" on the chip, and the link goes to `/items/jobset/WAR`. Correct split.

**On the Beastmaster finding:** agreed it is out of scope, and pinning it with `only_beastmaster_lacks_a_category_column` is the right call — it keeps a known-dead page visible instead of letting a future pack regeneration quietly change which routes are empty.

No blocking issues found. Merging.

PR #1049, opened 2026-07-31
feat(nav)!: remove the top bar, move chrome into the sidebar

akarras:
Deletes the application top bar. The sidebar becomes the sole owner of global chrome on desktop; below 1024px a three-slot fixed bottom bar carries Menu, Search and Items.

Everything the top bar hosted has a home:

| Was in the top bar | Now |
|---|---|
| `SearchBox` | An overlay — `Cmd`/`Ctrl`+K, a sidebar row, or the mobile bar |
| `LanguageNavMenu` | An inline accordion inside the account drop-up |
| `QuickThemeToggle` | The account drop-up |
| `UserMenu` | The account drop-up at the bottom of the sidebar |
| Hamburger (mobile) | The mobile bar's Menu slot |

Search and Item Explorer are promoted above the `Tools` header as a pair — Search jumps to a known item, Explorer browses when you don't. `Tools` is now an honest list of nine analyzers instead of nine analyzers plus a database.

Deletes `top_bar.rs`, `apps_menu.rs` and `LanguageNavMenu`. `AppsMenu` had been unreferenced dead code duplicating the sidebar's tool list.

## Staging

Landed in four stages so the app never sat broken between them. Stages 1–3 are additive and each left the app working with temporary duplicate controls; stage 4 is atomic because adding the mobile bar without removing the top bar doubles the chrome on phones, and removing the top bar without the mobile bar strands the drawer with no trigger.

## ⚠️ Needs manual verification before merge

CI does not run `cargo test` on this repo, and this change is almost entirely CSS and layout — **green CI is not evidence it works.** Outstanding:

- [ ] Width matrix: ≥1536px (ad rail), 1024–1535px, 1023px, 375px
- [ ] Signed-in and signed-out account drop-up
- [ ] One non-English locale (German is the tightest fit for the bar labels)
- [ ] Scroll to the bottom of a long table on a phone — bar must not cover the last row, footer must clear it
- [ ] **iOS specifically:** autofocus calls `.focus()` from inside an `Effect` rather than synchronously in the user-gesture stack, so Safari may decline to raise the keyboard
- [ ] Collapsed-sidebar drop-up — its `bottom: 6.5rem` is computed from the stylesheet by hand, never measured

## Accepted trade-offs

- **Logged-out mobile users lose the persistent sign-in button.** It's reachable via the drawer footer instead. Deliberate, in exchange for a single account surface.
- **`aria-modal` is not set on the search overlay.** Focus isn't trapped, so claiming it would tell assistive tech to ignore a page the user can still tab into. The existing `Modal` component has the same gap and should get a real focus trap at the same time.

## Pre-merge review findings, all fixed

A whole-branch review caught one Critical and three Important issues:

- **Critical:** below 1024px the overlay panel is opaque and full-viewport, painted over its own backdrop — so on a phone there was no way out at all (no backdrop click, no hardware Escape, the mobile bar covered). Now has an explicit close button.
- The site footer renders outside `AppShell`, so the bar's compensating padding never reached it.
- `.side-nav` is `overflow-y: auto`, which makes it a scroll container on both axes — the collapsed drop-up would have been clipped to a sliver rather than overhanging the content.
- The global `a:not(.nav-link)…` rule (specificity 0,5,1) silently beat every new class on an `<a>`, so the mobile bar's active-page highlight never rendered and Item Explorer didn't match the Search row it pairs with.

## Known, deliberately not fixed

`.side-nav-item` — the ordinary sidebar nav links — has the same anchor-specificity loss as the four selectors fixed above. It's a pre-existing bug on `main` and was left alone to avoid scope creep here.


akarras:
## Triage: not mergeable yet — rebase first, and rebase carefully

Reviewed as part of the scheduled PR sweep alongside #1047 and #1048. Holding off on a full code review until this is green, because the diff will change under a rebase.

**Two blockers:**

1. **`mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`** against `main` as it stands.
2. **No CI checks have reported** on `claude/top-bar-sidebar-redesign-d938db`.

**The rebase needs care — this is the stale-base trap.** This branch's merge-base is `5b21bbce`; `main` is now `05bb5dba`, four PRs ahead:

- `d6883fe5` **#1045 fix(side-nav): highlight the tool you're actually on**
- `df684e79` **#1044 feat(welcome): add the color palette picker**
- `aa216891` #1040 top-opportunities rebuild
- `05bb5dba` #1046 chart comparison grid

The overlap between what this branch touches and what `main` changed since its base is exactly where the damage would land:

```
style/tailwind.css
ultros-frontend/ultros-app/src/components/side_nav.rs
ultros-frontend/ultros-app/src/components/theme_picker.rs
ultros-frontend/ultros-app/locales/{en,de,fr,ja,cn,ko,tc}.json
```

`side_nav.rs` and `theme_picker.rs` are the two files #1045 and #1044 just landed in. A PR that removes the top bar and moves chrome into the sidebar can absorb those files wholesale during conflict resolution and **silently revert both fixes with no conflict marker to warn you**. After rebasing, verify with `git diff origin/main HEAD -- <those files>` and confirm #1045's active-tool highlighting and #1044's palette step are still present — don't trust a clean rebase.

**Ordering.** #1047 and #1048 have zero file overlap with each other and are both verified green, so they can land in any order. This one overlaps both — `item_explorer.rs` with #1048, the seven locale JSONs with #1047 — so it should land **last** and be rebased after they merge, otherwise you resolve the same locale-JSON conflicts twice.

Happy to do the full review once it's rebased and CI is green.

akarras:
## Review

Reviewed against `origin/main`, and merged together with #1047/#1048/#1050 to check cross-PR interactions (clean — no conflicts, and no semantic breakage between this PR and the other three).

### On "green CI is not evidence it works"

Agreed, and thank you for saying so explicitly. `cargo test` is commented out in `rust.yml`, so I ran the suites locally against the merged tree:

| Gate | Result |
|---|---|
| `cargo test -p ultros-app --lib` | **371 passed / 0 failed** |
| `cargo test -p ultros-charts` | 101 passed |

### The cascade fix is correct — verified against the compiled stylesheet, not the source

This is the part I most wanted to confirm, because the failure mode is silent. Checking `target/site/pkg/ultros.css` (the actual build output):

```css
a:not(.nav-link):not(.side-nav-item):not(.menu-item):not(.mobile-bar-slot):not(.btn)… { … }
```

The opt-out list survives compilation. More importantly, the *reason* it works is layer position, not specificity: `@layer theme, base, components, utilities` is declared at line 3, Tailwind's `.bottom-0` lands inside `@layer utilities`, and every rule this PR adds is emitted **unlayered** at column 0. Unlayered styles beat all layered styles regardless of specificity — so:

- `.toast-container { bottom: calc(4.25rem + env(safe-area-inset-bottom)) }` genuinely beats Tailwind's `.bottom-0` on the same element (I checked its position: inside a column-0 `@media (max-width: 1023px)`, so still unlayered). The toast clearance is live, not inert.
- The `.mobile-bar-slot` / `.menu-item` additions are the right *structural* fix rather than reaching for `!`.

### Other things I checked rather than assumed

- **`viewport-fit=cover`** was added to the viewport meta. Without it every `env(safe-area-inset-bottom)` in this PR computes to `0` on iOS and the bar sits under the home indicator. Easy to miss; good catch.
- **`.site-footer` (lib.rs:367) and `.toast-container` (toast.rs:73) both exist** — the two compensating-padding fixes actually have elements to attach to.
- **Ad rail cannot mis-place.** `.app-shell > .app-shell-ad-rail` is `display: none` by default and only gets `grid-area: ads` inside the ≥1536px query, so dropping the `"top"` row from `grid-template-areas` can't cause stray auto-placement at 1024–1535px.
- **Dead code is genuinely gone** — no remaining references to `top_bar` / `TopBar` / `AppsMenu` / `LanguageNavMenu` anywhere outside one doc comment.
- **Tests use `Owner::new()` / `owner.with()`** correctly, which is required or a bare `RwSignal::new` panics with "no Arena is active" in this crate.
- **i18n**: all seven locales carry real translations (not English stubs) for `search`, `menu`, `sign_in`, `account`, `close`.

### Your pre-merge checklist

Worked through it in a real Chrome against a local build of this branch merged with #1047/#1048/#1050, backed by the live Postgres + ClickHouse. Measured values, not impressions:

- [x] **Width matrix.**
  - **1600px** — `grid-template-areas: "side main ads"`, columns `240px 1120px 240px`, no horizontal scroll. (Rail itself is `display:none` because `:has(.ad.hidden)` fires with no AdSense fill locally — the reserved column is correct.)
  - **1280px** — `"side main"`, mobile bar `display:none`, ad rail `display:none`.
  - **~500px and 375px** — mobile bar `display:flex; position:fixed`, sits flush at the bottom (top 756 + height 56 = 812), sidebar translated off-canvas (`matrix(1,0,0,1,-280,0)`).
  - No horizontal overflow at any width.
- [x] **Signed-out account drop-up** — opens on click, `aria-expanded=true`, 213×157, fully in viewport, rows: Login with Discord / Settings / Language / Dark. Language accordion expands to all seven locales.
- [ ] **Signed-in drop-up** — not testable here, local Discord OAuth isn't wired up.
- [x] **German** (switched via the new accordion, `<html lang="de">`): bar reads **Menü / Suchen / Items**, each in a 124px slot with `scrollWidth == clientWidth` — no overflow at the tightest fit. Sidebar shows "Suchen ⌘K", account trigger "Anmelden".
- [x] **Bar doesn't cover content; footer clears it.** At 375px: `.app-shell-content` `padding-bottom: 68px`, `.site-footer` `padding-bottom: 68px`, `.toast-container` `bottom: 68px` — all three against a 56px bar. The toast one is the notable result: it proves the unlayered rule beats Tailwind's `.bottom-0`, which is the thing that would have silently failed.
- [ ] **iOS autofocus** — no iOS device. On desktop Chrome the overlay's input *is* focused after open (`document.activeElement === input`), so the Effect fires; whether Safari honours it from outside the gesture stack remains open.
- [x] **Collapsed-sidebar drop-up — the hand-computed `bottom: 6.5rem` is right.** Measured with the sidebar collapsed to 56px: panel is `position: fixed`, 240×157, **fully in viewport**, right edge at 246px so it genuinely overhangs the content, and the gap between panel bottom and trigger top is **5px** — against the 0.35rem (5.6px) the expanded state uses. Good guess.

Two more things I checked that weren't on your list:

- **The mobile bar's active-page highlight renders.** On `/items` the Items slot has `aria-current="page"`, background `color(srgb 0.65 0.55 0.98 / 0.16)` (not transparent), white text, and `text-decoration: none`. That is the anchor-specificity fix working end to end — it was the one most likely to be silently inert.
- **Search overlay**: `role="dialog"`, backdrop present, close button exactly **44×44** and hit-testable at 375px where the panel is opaque full-viewport. That is the Critical fix, and it holds.

### Not verified — flagging honestly

- **iOS Safari autofocus.** No iOS device here. The concern in the description is real and remains open; worst case the sheet opens without the keyboard and the user taps the field.
- **Signed-in account drop-up.** Local Discord OAuth isn't wired up, so only the signed-out panel was exercised.

### One finding worth a follow-up: the account row can start below the fold

`.side-nav` is `flex-direction: column; overflow-y: auto` with `.side-nav-sections { flex: 1 }`, and the account drop-up is now the **only** surface for sign-in, language and theme.

Measured at 1280×801 on the home page:

```
sidebar scrollHeight 965  vs  clientHeight 801   (overflows by 164px)
account trigger bottom = 916  →  115px below the fold
```

So on a 800px-tall laptop viewport a signed-out visitor has to scroll the sidebar to find "Sign in" at all. Previously it was a persistent top-bar button. The description lists losing the persistent mobile sign-in button as an accepted trade — this is the desktop version of the same trade, and I don't think it was intended, since 18 nav rows plus two hero rows plus the account row is simply taller than a common viewport.

Not blocking (it is reachable, and the drop-up itself works correctly once scrolled to). Worth considering `position: sticky; bottom: 0` on `.side-nav-account` so the account row pins to the bottom of the sidebar rather than scrolling away — that would also make the collapsed-state `bottom: 6.5rem` less dependent on the footer's measured height.

Also noticed while switching locales: "Groups" is untranslated in German (renders "Groups", not "Gruppen"). Pre-existing, unrelated to this PR.

### Verdict

Approving and merging. The staging discipline, the pre-merge review pass, and the honest "green CI is not evidence" framing all made this a much easier review than a change this size usually is.

PR #1050, opened 2026-07-31
fix(charts): unclip toolbar popovers, size buckets by actual data span

akarras:
## Triage of the four reported sale-history-chart symptoms

| Symptom | Root cause |
|---|---|
| "Locked into world grouping / smallest level is world" | **Bug 1** — the group-by menu *does* offer Region/Datacenter/World on region pages, but opens clipped inside the toolbar |
| "Only ever get one data point" | **Bug 2** — 30-day buckets over ~2 months of ClickHouse data |
| "Scatter/density on, cursor still locked to the candlestick positions" | **Bug 2** — the crosshair snaps to bucket centers in every mode; with 1–2 buckets there is 1–2 snap positions |
| "Worlds filter / overlay filters don't expand" | **Bug 1** — same clipped popovers |

### Bug 1: toolbar popovers clipped (client CSS)

The toolbar row used `overflow-x-auto`; per CSS, a non-`visible` `overflow-x` forces `overflow-y` to compute to `auto`, so the row became a scroll container ~27px tall. All three popovers are absolutely positioned inside it and opened invisibly. Verified live on ultros.app: menu present with all three group options, `clientHeight: 27`, `scrollHeight: 125`, menu unreachable. The row now wraps (`flex-wrap`) instead of scrolling.

### Bug 2: bucket width derived from the requested window, not the data (server)

With no `from` param the handler defaults to `now − 12y`, and the ladder duly picks 30-day buckets — but prod ClickHouse `sales` only holds ~2 months, so every item collapsed to 1–2 buckets (verified: `/api/v1/price_series/Gilgamesh/7` → `bucket_seconds: 2592000`, 2 buckets). `build_price_series` now measures the span the rows actually cover after the first pass and re-queries once at the narrower ladder step (`narrow_bucket_for_actual_span`, mirror of the existing widening loop; at most one extra query). The same items now get daily buckets.

The density mode fetch had the same default-window hole; the client now bounds it to the series' resolved domain so both charts share a window.

### Synthetic data + static snapshot tests

New deterministic fixture (`synthetic_price_series`: seeded LCG walk, two worlds, gap buckets, sub-3-sale candles, raw sales) and byte-for-byte SVG snapshots for Price/Candles/Range/Grid/Density under `ultros-charts/src/charts/snapshots/`. Regenerate after intentional rendering changes with:

```
UPDATE_SNAPSHOTS=1 cargo test -p ultros-charts snapshot
```

### Verification

- `check_ci.sh` exit 0 (fmt + workspace clippy + xiv-gen feature clippy)
- `cargo test -p ultros-charts`: 96 passed (5 new snapshots stable across runs)
- `cargo test -p ultros-app --lib`: 363 passed
- Prod API probed to confirm both the bug and that grouping/explicit windows already work server-side (a 14-day explicit window returns daily buckets; `group=datacenter` returns per-DC series) — so the server fix only needed the narrowing pass


akarras:
## Review: approved ✅

Reviewed against `origin/main` (branches from `05bb5dba`, zero commits behind).

### Bug 1 — the popover clipping diagnosis is exactly right

`overflow-x: auto` forcing `overflow-y` to compute to `auto` is a real and frequently-missed CSS rule, and a ~27px scroll container swallowing three absolutely-positioned popovers explains all three "filters don't expand / locked to world grouping" reports at once. `flex-wrap` is the right fix rather than `overflow: visible` on one axis, which CSS will not honour here.

I checked the cascade question this fix depends on: the `.side-nav-item` / opt-out rules and the toolbar styles in `style/tailwind.css` sit at column 0, i.e. **outside** any `@layer`, so they win over Tailwind's layered utilities regardless of specificity. The class swap therefore takes effect rather than being silently overridden.

### Bug 2 — the narrowing pass is correctly bounded

Traced the control flow rather than trusting the shape:

- `may_narrow` is set `false` **before** the `continue`, so the narrowing can fire at most once — no oscillation against the inner widening loop, which remains free to widen if the narrower query overshoots `MAX_BUCKETS`.
- The re-query keeps the original wide `from`/`to`, but ClickHouse only emits buckets where rows exist, so row count tracks the data span, not the window. No blowup.
- `actual_span` correctly adds one `bucket_seconds` because bucket timestamps are starts — so the single-bucket case (`first == last`) still yields a real span and narrows instead of dividing by zero.
- The response's `from`/`to` were already the *data* domain, not the requested one, so the client-side density window derived from them is genuinely bounded. The `.filter(|s| !s.is_empty())` guard matters here: `is_empty()` is `series.iter().all(|s| s.buckets.is_empty())`, and on an empty series `from`/`to` fall back to the **requested** bounds — without the guard the density fetch would inherit the 12-year window it is trying to avoid.

### Verified locally

CI does not run `cargo test` on this repo, so this needed a local run. With all four open PRs merged together:

| Gate | Result |
|---|---|
| `cargo test -p ultros-charts` | **101 passed / 0 failed** (5 SVG snapshots stable) |
| `cargo test -p ultros-app --lib` | 371 passed |

### ⚠️ Heads-up: conflicts with #1047 (handled)

`snapshot_tests.rs` builds an exhaustive `DensityChartOptions` literal; #1047 adds a `milestones` field to that struct. Git merges both **without conflict** and the result then fails to compile — and CI stays green because `cargo test` is disabled. I've folded the one-line fix (`..Default::default()`, matching `web_options` in the same file) into #1047 as the second-lander. Nothing needed on this branch.

### Browser + live-API verification (local build, all four PRs merged)

**Bug 1 — fixed, measured on the real item page** (`/item/Cerberus/52256`):

| | before (your report, prod) | after (this build) |
|---|---|---|
| toolbar `overflow-y` | `auto` | **`visible`** |
| `clientHeight` / `scrollHeight` | 27 / 125 | **27 / 27** (not a scroll container) |
| Overlays popover | opened invisibly | **208×137, fully in viewport, hit-test lands inside it** |

**Bug 2 — fixed:** `/api/v1/price_series/Cerberus/52256` now returns `bucket_seconds: 86400` (daily) where the ladder would previously have picked `2592000` (30-day). 10 buckets instead of 1–2.

### 🔴 Found while verifying: the default window is still truncated (pre-existing, not caused by this PR)

Same item, same build, three requests:

| request | domain `to` | buckets | sales |
|---|---|---|---|
| no params (what the item page actually sends) | 2026-06-20 | 10 | **98** |
| `?to=<now>` | 2026-07-31 | 17 | **148** |
| `?from=<now-400d>` (open-ended `to`) | 2026-07-25 | 16 | 145 |

**The open-ended path silently drops 50 of 148 sales — the newest ~6 weeks.**

Cause is `web.rs:671`, which predates this PR:

```rust
let to = if query.to.is_none() {
    let secs = to.timestamp() - to.timestamp().rem_euclid(bucket_seconds);
```

`bucket_seconds` there is the **pre-narrowing** width — 30 days for a default open-ended request — so `now` is floored to a 30-day-since-epoch boundary and up to a month of the newest sales is discarded before the query even runs.

This is **not a regression from this PR**: before it, the same snap applied and the chart showed 1–2 month-wide buckets, so the missing tail was invisible. Your change is a strict improvement. But it does mean the stated goal isn't fully reached — the resolution now follows the data while the *window* still follows the coarse pre-narrowing bucket.

Not blocking the merge, because the fix touches cache-key cardinality (`price_series_cache::CacheKey` uses this `to`), which is a production trade-off worth deciding deliberately rather than folding into a review. The likely shape is clamping the snap granularity, e.g. `bucket_seconds.min(DAY)` — one cache entry per item/world/day instead of per 30 days, and at most a day of tail loss. Filed as a follow-up.

### Non-blocking note

Because prod's ClickHouse `sales` only holds ~2 months, the narrowing branch will fire on essentially **every** `price_series` request until the `clickhouse_backfill` one-shot is run — i.e. two CH queries per cold cache entry instead of one. That's the correct trade (the alternative is a useless chart), the second query is over a sparse table, and the response cache absorbs the repeat. Worth knowing it flips back to one query per request once history is backfilled.

Merging.

PR #1051, opened 2026-08-01
Flip Finder: spreadsheet redesign — contained pane, resizable columns, realistic defaults

akarras:
## What changed

Redesigns the Flip Finder from a window-scrolled infinite list into a contained spreadsheet pane, addressing four UX complaints:

1. **Contained spreadsheet pane** — the table now fills the viewport below the filter bar with ONE scrollport for both axes; the column header is sticky inside it (new `fill` mode on `VirtualScroller`). The ~50 lines of scrollLeft-mirroring JS are deleted, so the mobile header/body horizontal desync is structurally impossible.
2. **Adjustable columns** — a `ColumnSpec` registry (`analyzer_columns.rs`) drives all widths via `--colw-*` CSS variables. Columns drag-resize from the header edge (pointer capture, live at 60fps, persisted to localStorage, double-click resets). Right-click / long-press opens a header context menu: sort asc/desc, hide column, reset width(s), manage columns. Breakpoint-based column hiding is gone — visibility is purely the opt-in `?cols=` system on every device; mobile relies on horizontal scroll.
3. **Clearer sell-world copy** — "Select World:" → "Sell on world:"; the landing page asks which world you'll sell your flips on. Translated across all 7 locales with a terminology-consistency review pass (ja 転売品, cn/tc 物品, ko 서버 unified on the landing page).
4. **Realistic default filters** — a visit with no filter/sort params seeds the Realistic-flips preset (`min-buy=5000`, `last-sold=1d`, `roi=30`, sort profit/day, `next-sale=1d`) as removable chips. Explicit URLs are honored verbatim; `?cols=`/region toggles don't suppress seeding. A test derives the seeded set from the built-in preset so they can't drift.

Also: spreadsheet gridlines + opaque sticky header, ARIA repair (`role="row-group"` was not a real role; now `table → row → cell`), dead CSS removal.

## Notable fixes found during review/verification

- **SSR pane height**: `use_window_size` is `INFINITY` on the server under leptos-use's `ssr` feature — without a finiteness guard the server streamed `height:2147483647px`.
- **Seeding race**: seeding via `use_navigate` loses to the world navigator's mount-time URL rebuild (caught in live browser verification — the params never landed). Seeding now writes through `query_signal` setters, which coalesce in the router's mutation queue and compose with the navigator.
- **Resize hardening**: `pointercancel` repaints the committed width (tachys skips identical-string style rewrites, so a canceled drag could strand a phantom width); zero-movement clicks don't pin defaults into storage; stored widths clamped to `[min, 2000]` with NaN/∞ screened.
- **Touch long-press hardening**: outside-close moved to bubble phase so the synthetic-click guard works; 10px movement threshold; `select-none` on header cells.

## Behavior changes to know about

- `next-sale=1d` is no longer silently appended to URLs that already carry other filters — explicit URLs render exactly what they say.
- `ScrollSource::Window` no longer has any call site (kept as a supported `VirtualScroller` mode with a documented allow).
- Mobile shows all default-visible columns (~1140px row) with horizontal scroll instead of force-hiding columns below breakpoints.

## Testing

- 307 unit tests pass; `./check_ci.sh` (fmt + clippy `-D warnings`) green on every commit.
- E2E screenshot harness run against a live server; flip-finder failures were all environmental (AdSense script noise, ClickHouse down for enrichment, analyzer warmup window on the reused server).
- Interactive browser verification: sell-world copy, spreadsheet pane rendering with gridlines, 4,334-row virtualized table — and it caught the seeding race that unit tests couldn't.
- Not yet verified on a physical touch device: long-press menu + touch drag-resize (worth a quick check on a phone).

## Follow-ups (out of scope)

- Keyboard navigation + `aria-rowcount`/`aria-sort` for the menu/grid; accessible name for the table.
- Stale `tailwind-version = "v3.4.1"` pin in Cargo.toml (build actually uses v4.1.10) — separate task chip exists.
- Consider a slimmer default `?cols=` set for phones if the default row feels wide in practice.
- Other analyzer routes (vendor-resale, venture, recipe, leve) still use breakpoint hiding and could adopt the registry pattern later.


akarras:
## Not reviewed this round — needs a merge from `main` first

Picked this up during a batch review of the open PRs. It arrived mid-run and is currently **`CONFLICTING` / `DIRTY`** against `main`, so I reviewed and landed #1047–#1050 and left this one.

### Why it conflicts

The branch point is `c4df0cb7` (#998) — **36 commits behind `main`** as of now. That is far enough back that `git diff origin/main <this branch>` reports ~150 changed files, including `.gitmodules`, `xiv-gen/ffxiv-datamining`, `ultros-frontend/ultros-xiv-icons/universalis-assets`, `data/xiv-db/*.rkyv`, and every `.github/workflows/*` file.

**Those are an artifact of the stale base, not real changes.** The true diff — `git diff $(git merge-base origin/main HEAD) HEAD` — is the 17 files you'd expect:

```
style/tailwind.css
ultros-frontend/ultros-app/Cargo.toml
ultros-frontend/ultros-app/locales/*.json  (7)
ultros-frontend/ultros-app/src/components/filter_chip.rs
ultros-frontend/ultros-app/src/components/virtual_scroller.rs
ultros-frontend/ultros-app/src/query_defaults.rs
ultros-frontend/ultros-app/src/routes/analyzer.rs
ultros-frontend/ultros-app/src/routes/analyzer_columns.rs
ultros-frontend/ultros-app/src/routes/mod.rs
docs/superpowers/{plans,specs}/2026-07-30-flip-finder-spreadsheet*.md
```

### ⚠️ Do not rebase this branch

Rebasing across a gap this size silently reverts newer `main` work with **no conflict markers** — in this case it would undo the LFS game-data-pack migration (#1033) and re-introduce the submodules. Use `git merge origin/main`, then verify with `git diff origin/main HEAD --stat` and confirm the result only touches the 17 files above.

### What landed under you that will need re-verification

Four PRs merged since this branch was cut, and two touch files this PR also touches:

- **#1049** (`feat(nav)!: remove the top bar`) — rewrites large parts of `style/tailwind.css`, deletes `top_bar.rs` / `apps_menu.rs`, and adds a fixed mobile bottom bar below 1024px with `padding-bottom: calc(4.25rem + env(safe-area-inset-bottom))` compensation on `.app-shell-content`. **This is the one that matters for a full-viewport spreadsheet pane**: your pane sizes itself to the viewport below the filter bar, and there is now a 56px fixed bar plus 68px of content padding underneath it on mobile. The pane height calculation almost certainly needs to account for that, and it is worth re-measuring on a phone width.
- **#1050** (`fix(charts)`) — `chart_toolbar.rs` swapped `overflow-x-auto` for `flex-wrap`, because a non-`visible` `overflow-x` forces `overflow-y: auto` and clipped every popover in the row. **Your header context menu is the same shape** — an absolutely-positioned popover inside a scrollable table header. Worth checking it isn't clipped by the spreadsheet pane's own scrollport.
- #1047 and #1048 shouldn't interact.

Also: the seven locale files were edited by both #1047 and #1049, so expect real (small) conflicts there.

### Note for next time

This one is a good argument for cutting agent branches from `origin/main` immediately before work starts — the 17-file change here is fine, but it's sitting behind 36 commits of churn that has nothing to do with it.

Happy to review properly once it merges cleanly.

akarras:
## Review: rejecting — the base has moved too far under this branch

This is good design work aimed at a version of the Flip Finder that no longer exists. It can't be rebased; it would have to be re-done. Detail, so the effort isn't wasted:

### The branch forked at `c4df0cb7` (#998) and is 37 commits behind

`git merge-tree` conflicts in 9 files, including `analyzer.rs` itself. That file has been rewritten **twice from the same base**: main changed 871 lines since the fork point, this branch changed 870. That's a head-on collision, not a mechanical conflict.

### It would revert four merged Flip Finder PRs

Missing from this branch: #1014 (every column filterable), #1025 (review findings on those filters), #1028 (empty state, help header, filter history, tooltips, mobile columns), #1024.

The locale files make it measurable. Comparing key sets against the shared fork point:

- main added **158** keys since the fork; this branch added **6**.
- All 158 are absent here — including `analyzer_tooltip_*` (7), `analyzer_filter_*` (5), `analyzer_confidence_gte`/`analyzer_drift_gte`/`analyzer_volume_gte`, `analyzer_empty_title`, `analyzer_rows_lacking_data`, `analyzer_last_sold_*`, `analyzer_columns_picker_desktop_only`.
- `analyzer_tooltip_confidence` is referenced by main's `analyzer.rs` and appears nowhere in this branch's.

Net effect: main's `analyzer.rs` is **3,889** lines, this branch's is **3,488** — 400 lines *smaller* despite adding a full spreadsheet redesign, because it predates the feature work.

The other 130-odd missing keys span the privacy policy (#1026), chart specs (#1042/#1046/#1047), scrip sources (#1031), top opportunities (#1040), lists (#1029), logged-out empty states (#1027), help screenshots (#1032), and the welcome palette step (#1044). Resolving the 7 locale conflicts by hand means reconciling all of that.

### The layout premise also changed

#1049 removed the top bar and moved chrome into the sidebar — this branch is missing its `menu`/`search`/`sign_in`/`region`/`set_home_world` keys. The core idea here ("the table fills the viewport below the filter bar, one scrollport for both axes") is measured against a page chrome that no longer exists, so the height math needs re-deriving regardless.

### Also: CI never ran

`statusCheckRollup` is empty — no fmt, no clippy, no Docker build. There's no evidence this compiles against anything, let alone current main.

### What's worth keeping

The design is sound and several findings in the PR body are genuinely valuable and independent of the collision — the SSR `use_window_size` `INFINITY` guard, the `query_signal`-vs-`use_navigate` seeding race, and the `pointercancel` phantom-width repaint are all real and hard-won. `docs/superpowers/specs/2026-07-30-flip-finder-spreadsheet-design.md` and the plan are still the right spec.

Suggested path: branch fresh from current main and re-apply the redesign on top of today's `analyzer.rs`, folding the `ColumnSpec` registry around the existing filter/tooltip system rather than replacing it. `analyzer_columns.rs` and the `VirtualScroller` `fill` mode should port over with little change — they're the parts that don't collide.

Closing rather than leaving it open, since no amount of rebasing gets this branch to a mergeable state.

PR #1052, opened 2026-08-01
⚡ Bolt: O(N) Top-K optimization in live_sale_ticker.rs

akarras:
💡 **What:** Replaced the `O(N log N)` sort `.sorted_by_key().take(8)` with a linear-time selection algorithm using `.select_nth_unstable_by_key(8)` followed by `.truncate(8)` and `.sort_unstable_by_key()`.
🎯 **Why:** This component fetches recent sales periodically and on load. `recent_sales.sales` can be quite large for a busy world. Sorting the entire list just to display the top 8 items forces the main thread in WASM to do $O(N \log N)$ work, which can cause UI jank. Using $O(N)$ partitioning is much faster.
📊 **Impact:** Reduces time complexity from `O(N log N)` to `O(N)` + `O(K log K)` where `K` is 8. Prevents main thread WASM blocking during UI updates.
🔬 **Measurement:** Open the dashboard with the live sale ticker and monitor CPU profiling in browser devtools. The sort step during the API fetch should be significantly faster and no longer scale non-linearly with the number of sales.

---
*PR created automatically by Jules for task [12035204378917811562](https://jules.google.com/task/12035204378917811562) started by @akarras*


akarras:
## Review: approved ✅

Small and correct. Verified merged onto current `main` (which moved by four PRs while this was open — no overlap, merges clean).

**The bounds are right, which is the thing that usually goes wrong here.** `select_nth_unstable_by_key(8, …)` panics when the index is `>= len`, and the `if first_sales.len() > 8` guard means it's only reached at `len >= 9`, where index 8 is valid. Lists of 8 or fewer skip straight to the sort. No panic path.

**The selection semantics are also right.** The key is `Reverse(sold_date)`, so a smaller key is a more recent sale. After partitioning at index 8, indices `0..=7` hold the eight smallest keys — the eight most recent — and `truncate(8)` keeps exactly those before the final `O(K log K)` sort. That is the intended top-8.

`flat_map` → `filter_map` over an `Option` is behaviour-preserving and more idiomatic.

| Gate | Result |
|---|---|
| `cargo test -p ultros-app --lib` (merged with `main`) | **371 passed / 0 failed** |
| `cargo fmt --all -- --check` | exit 0 |
| CI "Lint, checks, and tests" | pass |

### Two notes, neither blocking

1. **The `O(N log N)` → `O(N)` framing oversells it.** `get_recent_sales_for_world` returns a bounded recent-sales list, not an unbounded one — at that size the difference between a sort and a partition is well under a frame, so "prevents UI jank" isn't something this will visibly deliver. The change is still a strict improvement and costs nothing, so it's fine to take; I'd just not treat the stated impact as measured.

2. **This swaps a stable sort for an unstable one.** `itertools::sorted_by_key` is stable; `sort_unstable_by_key` and `select_nth_unstable_by_key` are not, so sales sharing an identical `sold_date` can come out in a different relative order than the API returned them. That's cosmetic here **because this runs inside `spawn_local`, i.e. client-only** — it never renders on the server, so it can't produce the SSR/CSR ordering divergence that caused the item-page hydration crash in #960. Worth keeping in mind if this block ever moves into a `Resource`.

Merging.

PR #1071, opened 2026-08-01
fix(charts): stop the open-ended price window dropping the newest month of sales

akarras:
## The bug

An open-ended `/api/v1/price_series` request — no `from`, no `to` — resolves `from` to 12 years back. The bucket ladder duly picks its widest rung, **30 days**. Both `price_series` and `price_density` then floored the **queried window's exclusive upper bound** onto that rung:

```rust
let to = if query.to.is_none() {
    let secs = to.timestamp() - to.timestamp().rem_euclid(bucket_seconds);
    chrono::DateTime::from_timestamp(secs, 0).unwrap_or(to)
} else { to };
```

The ClickHouse predicate is `sold_date < to` (half-open), so **every sale in the current 30-day bucket was excluded from the response**.

This is the default path for every item page — `selected_range` starts at `None` (`item_view.rs:1270`), so `range = None` goes out until the user drags the timeline slicer.

## Reproduced against production

Same endpoint, open-ended vs. an explicit `from`/`to` covering the same span (the explicit path was never floored). Gilgamesh, fetched 2026-08-01:

| item | shown | actual | hidden |
|---|---:|---:|---:|
| 5 | 113 | 155 | 42 (27.1%) |
| 5057 | 342 | 509 | 167 (32.8%) |
| 5106 | 188 | 277 | 89 (32.1%) |
| 5111 | 495 | 743 | 248 (33.4%) |
| 7010 | 83 | 112 | 29 (25.9%) |
| 12538 | 171 | 266 | 95 (35.7%) |
| 19938 | 394 | 565 | 171 (30.3%) |
| 27722 | 47 | 51 | 4 (7.8%) |
| **total** | **1,833** | **2,678** | **845 (31.6%)** |

Every chart's right edge sat at **2026-07-05/06** — the 30-day epoch boundary — roughly 27 days stale.

## The fix

The floor existed only so a live view would share a cache entry instead of minting a unique key every second. That is a **cache-key** concern, so it belongs on the key:

- `cache_ttl_secs(closed, bucket_seconds)` — one place deriving both the TTL and the key's grain, so the key rolls over exactly when the entry expires (one live entry per item/scope, as before).
- `open_window_cache_stamp(to_ts, grain)` — quantizes **`CacheKey.to` only**. The queried window keeps running to `now`.

Staleness is now bounded by the TTL, which already capped it at an hour. Cache cardinality is unchanged. No data is dropped.

Applied to both `price_series` and `price_density`, which had the identical floor.

`item_card.rs` (the OG PNG) was never affected — it calls `build_price_series` directly with an unfloored `to`, which is why cards stayed current while charts did not.

## Notes for review

- **This predates #1050.** Before it the chart drew 1–2 month-wide buckets, so the missing tail was invisible; #1050 made *resolution* follow the data and exposed it. Not a #1050 regression.
- The response's `from`/`to` are the *actual data span* (`build_price_series`), so the visible symptom was the chart's right edge, not an obviously-wrong axis bound.

## Verification

- `./check_ci.sh` → **exit 0** (fmt + clippy, workspace-wide).
- `cargo clippy -p ultros --all-targets -- -D warnings` → exit 0 on a *forced* re-check, so the new tests compile under test cfg.
- Three regression tests added to `price_series_tests`: key stability across the grain, the grain never exceeding the TTL at any ladder rung, and a degenerate-grain guard.
- ⚠️ **The tests could not be executed in-crate.** `cargo test -p ultros` fails to *link* on Windows/MSVC with `LNK2001/LNK2019 unresolved external symbol … ultros_app`, and this is pre-existing and unrelated — the unresolved symbols come from long-standing tests (`auth_failures_keep_their_structured_body`, `test_persistence`, `method_to_db_round_trip_webhook`). To prove the logic rather than just the compile, the two pure helpers and their tests were mirrored into a standalone file and run with `rustc --edition 2024 --test`: **4/4 pass**. Worth a run on Linux to confirm in-crate.

### DIFF
--- a/ultros/src/web.rs
+++ b/ultros/src/web.rs
@@ -437,6 +437,43 @@ fn resolve_bucket_seconds(bucket: Option<i64>, span_secs: i64) -> i64 {
     }
 }
 
+/// How long a cached response stays servable, and — for an open-ended window
+/// — the grain [`open_window_cache_stamp`] quantizes its cache key onto.
+/// Deriving both from one place means exactly one entry per item/scope is live
+/// at a time: the key rolls over on the same schedule the entry expires on.
+///
+/// Capped at an hour so an open window is never served staler than that, and
+/// floored at a minute so a hypothetical sub-minute bucket couldn't turn the
+/// cache into a no-op. A closed window is immutable, so it just takes the cap.
+fn cache_ttl_secs(closed: bool, bucket_seconds: i64) -> u64 {
+    if closed {
+        3_600
+    } else {
+        (bucket_seconds as u64).clamp(60, 3_600)
+    }
+}
+
+/// Quantize an open-ended window's end onto a `grain`-second grid, for use in
+/// the **cache key only** — never for the window actually queried.
+///
+/// An open-ended request ends at "now", so feeding that raw timestamp into the
+/// cache key mints a fresh entry every second and the cache never hits.
+/// Rounding it onto the same grid as the entry's TTL keeps one live entry per
+/// item/scope, which is all the quantization was ever for.
+///
+/// This deliberately moves the *key* and not the queried window. Flooring the
+/// window itself — which both handlers used to do, at `bucket_seconds`
+/// granularity — drags the query's exclusive upper bound backwards, excluding
+/// every sale after the boundary. An open-ended "full history" request
+/// resolves to a 12-year span, the ladder duly picks its widest step (30
+/// days), and so the newest 0–30 days of sales silently vanished from every
+/// chart. Serving a slightly stale snapshot is the cache's job and is bounded
+/// by the TTL; narrowing the window is data loss and is not.
+fn open_window_cache_stamp(to_ts: i64, grain: i64) -> i64 {
+    let grain = grain.max(1);
+    to_ts - to_ts.rem_euclid(grain)
+}
+
 /// Uniform bin height covering `[lo, hi]` inclusive in `bins` steps, floored
 /// at 1 gil so degenerate windows (every sale at one price) still bin sanely.
 fn density_bin_width(lo: u32, hi: u32, bins: u16) -> f64 {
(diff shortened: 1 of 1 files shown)


akarras:
## Review: verified independently — merging

**Bug confirmed live in production.** Reproduced the data-loss table against ultros.app today (2026-08-01), open-ended vs. an explicit window covering the same span:

| item | open | explicit | hidden |
|---|---:|---:|---:|
| 5 | 113 | 155 | 42 (27.1%) |
| 5057 | 342 | 510 | 168 (32.9%) |
| 5106 | 188 | 277 | 89 (32.1%) |
| 5111 | 495 | 743 | 248 (33.4%) |
| 7010 | 83 | 112 | 29 (25.9%) |
| 12538 | 171 | 266 | 95 (35.7%) |
| 19938 | 394 | 565 | 171 (30.3%) |
| 27722 | 47 | 51 | 4 (7.8%) |
| **total** | **1,833** | **2,679** | **846 (31.6%)** |

Every open-ended response came back with `to: 2026-07-05T00:00:00` — the 30-day epoch boundary, 27 days stale — while the explicit path returned 2026-07-31/08-01. Matches the PR's numbers (the +1 is a sale that landed since).

The mechanism checks out exactly as described: the 12-year default span puts the ladder on its widest rung (30d), the old code floored the *queried* window's exclusive bound onto that rung, and `build_price_series` then narrowed `bucket_seconds` to 86400 from the actual data span — which is why the response advertises a 1-day bucket while having been cut on a 30-day boundary.

**Fix reviewed.** Moving the quantization onto `CacheKey.to` and leaving the queried `to` at `now` is the right call — this was always a cache-key concern. Confirmed `to` (unfloored) is what reaches `build_price_series`, and that no other `rem_euclid` floor on a query window survives in `web.rs`. Cache cardinality is if anything better than before: the key grain now equals the TTL exactly, where previously the key grain (up to 30d) far outran the 1h TTL.

**Tests.** CI does not run `cargo test` (commented out in `rust.yml`), and `cargo test -p ultros` genuinely cannot link on Windows/MSVC — the crate is bin-only (no `lib.rs`), so the PR's caveat is accurate rather than an excuse. I re-verified the two helpers with my own standalone harness rather than the PR's mirror, adding properties the PR didn't cover: key-grain equals TTL on every ladder rung, negative/pre-epoch timestamps, and `i64::MIN` as a degenerate grain. **5/5 pass.**

Partial trailing buckets are not a new rendering condition — the explicit-window path was never floored, so production charts have always handled them. `item_card.rs` calls `build_price_series` directly and was correctly unaffected.

No file overlap with #1051, so no cross-PR conflict.

Approving by merge (GitHub blocks self-approval on own PRs).

Issue #1057, opened 2026-08-01
the sticky header on the item explorer on mobile doesn't show the full filter list


akarras:
![Screenshot_20260801-122048.png](https://github.com/user-attachments/assets/aa553d92-abdf-47ab-9bc8-918ae0e7ed49)

When scrolling back the other direction the header comes into frame but otherwise doesn't work well

akarras:
Reproduced at phone width, with measurements. One note up front: the screenshot is the **Flip Finder** (`/flip-finder/:world`) — HQ / ITEM / PROFIT columns, clipboard + add-to-list per row — not `/items`. Everything below is that page.

## What's measurably broken

The sticky control bar is height-locked to `STICKY_BAR_HEIGHT` (76px, `components/filter_chip.rs:21`) because the table header sticks directly beneath it at that exact offset. So it can't wrap. Its top row also can't scroll — it owns the Views and Columns popovers, and an `overflow-x` there would compute `overflow-y: auto` too and trap them in a 32px strip.

That leaves it needing to *fit*, and it doesn't. Measured against prod at a real mobile viewport, row 1 wants **506px** at every width:

| viewport | space for row 1 | overflow |
|---|---|---|
| 320px | 256px | **250px** |
| 360px | 296px | **210px** |
| 393px | 329px | **177px** |
| 414px | 350px | **156px** |
| 480px | 416px | **90px** |
| 640px | 544px | 0 |

The overflow goes past `html { overflow-x: hidden }`, so at 393px **Columns** (370→460) and **Clear all** (472→537) are not merely cut off — there is no scrollbar and no wrap, so they are unreachable. At 360px and below, half of the saved-views control goes with them.

`position: sticky` itself is working: I scrolled to 600 / 1600 / 3200 / 2000 / 400 and the bar sat at `top: 0` and the table header at `top: 76` at every one of them.

## The fix in the PR

Below `sm` the row-1 controls drop to icon-only (Views, Save view, Columns, Clear all) and the realtime indicator drops to its dot; the gap tightens to `gap-2`. Labels use `sr-only sm:not-sr-only` rather than `hidden`, so every button keeps its accessible name at every width. Nothing is added or removed at `sm` and up. Result: **0px overflow at 360 / 393 / 414**, and 3px at 320 (down from 250px). No new user-facing strings.

## Two things I'd like you to confirm

[...]

PR #1083, opened 2026-08-02
test(flip-finder): guard the mobile control bar against overflowing again

akarras:
Adds a regression guard for the Flip Finder's mobile control bar. **No behaviour change** — the fix itself is #1082.

## Why

The bar's first row can neither wrap (it is height-locked to `STICKY_BAR_HEIGHT`, which the table header consumes as its sticky offset) nor scroll (it is the containing block for the saved-views and columns popovers, and an `overflow-x` would clip them). So it has to *fit*, and when it didn't, the surplus went past `html { overflow-x: hidden }`: `Columns` and `Clear all` rendered outside the viewport with no scrollbar and no wrap, i.e. simply unreachable rather than merely cramped.

That is invisible to every check we run. It is pure layout, so it needs a browser.

## What it does

`npm run test:flip-finder-mobile-bar` drives 360 / 393 / 414 and asserts:

- the bar and its first row do not overflow their own box;
- every top-row control is hit-testable at its own centre via `elementFromPoint`, so a control pushed past the edge fails loudly instead of silently vanishing.

It checks the **outcome, not the mechanism**, so it stays valid whether the row is kept fitting by a breakpoint, by shrinking, or by truncation.

It also refuses to run against a viewport it did not get: `innerWidth` is misreported under Chrome's mobile emulation (537 for a 393px device), so it reads `documentElement.clientWidth` and bails if that isn't the width it asked for.

## Verified both directions

Red against production, naming both unreachable controls:

```
[info] 360px: 4 control(s), row overflow 168px, bar overflow 160px
[fail] the Flip Finder control bar does not fit on mobile:
  - 360px: control bar overflows its box by 168px — the surplus is clipped by html{overflow-x:hidden} and cannot be scrolled to
  - 360px: "Columns" ends at x=419, past the viewport edge
  - 360px: "Clear all filters" ends at x=496, past the viewport edge
```

Green against a locally built `cargo leptos serve` carrying a fix:

```
[info] 360px: 4 control(s), row overflow 0px, bar overflow 0px
[info] 393px: 4 control(s), row overflow 0px, bar overflow 0px
[info] 414px: 4 control(s), row overflow 0px, bar overflow 0px
[done] control bar fits, and every control is reachable, at every phone width tested
```

Being straight about that second run: the green build was this branch's original markup fix, which I have since dropped in favour of #1082. I could not re-run it against #1082's build — `cargo leptos serve` ignores `LEPTOS_SITE_ADDR` and always binds :8080, and other sessions on this machine were holding that port, so I could no longer tell whose server I was measuring. Both fixes make row 1 fit at phone widths, so it should pass, but **please re-run it on #1082 before relying on it**. The `SETTLE` default of 9s is also short for a cold local server; the passing run used `SETTLE=35000`.

## Not covered

The 768–1024px band with German labels, which is #1082's harder case (the side nav claims 240px at `lg`, so the row is barely wider at 1024px than at 768px). Switching locale needs the in-app picker rather than a cookie, so that wants its own probe.

## Relationship to the issues

- #1082 fixes the overflow. This guards it.
- #1057 additionally reports that the bar "doesn't show the full filter list" and that "scrolling back the other direction" brings the header into frame. Neither is closed by #1082 or by this. I left [a comment on #1057](https://github.com/akarras/ultros/issues/1057#issuecomment-5157819861) with measurements on both: the chip strip is 255px wide at 393px with `scrollbar-width: none`, so 4+ filters mostly hide with no affordance, and showing them all needs the height-lock broken; and `position: sticky` held at every scroll offset I tested, so the scroll-direction symptom did not reproduce and needs the reporter's browser to narrow down.

### DIFF
--- a/integration/flip-finder-mobile-bar.cjs
+++ b/integration/flip-finder-mobile-bar.cjs
@@ -1,0 +1,126 @@
+// Regression probe for the Flip Finder's sticky control bar overflowing at
+// phone widths (issues #1055 / #1057).
+//
+// The bar is height-locked to STICKY_BAR_HEIGHT (components/filter_chip.rs) so
+// the table header can stick directly beneath it, which means it cannot wrap.
+// It also must not scroll: its top row owns the saved-views and columns
+// popovers, and an `overflow-x` there computes `overflow-y: auto` too and
+// would trap them in a 32px strip. So the row has to *fit*, and when it
+// didn't the surplus went past `html { overflow-x: hidden }` — leaving
+// `Columns` and `Clear all` rendered outside the viewport with no scrollbar
+// and no wrap to reach them. At 393px that was 177px of unreachable controls.
+//
+// Anything added to that row has to keep it fitting, by any means: a
+// breakpoint, shrinking, truncation. This asserts the outcome rather than the
+// mechanism, so it stays valid across whichever the row uses. Layout is the
+// whole bug, so it can only be checked in a browser.
+//
+// Not covered: the 768-1024px band with German labels, where the side nav
+// claims 240px and the row is barely wider than it is at 768px. Switching
+// locale needs the in-app picker rather than a cookie, so it wants a separate
+// probe.
+const puppeteer = require('puppeteer');
+
+const BASE_URL = process.env.BASE_URL || 'http://127.0.0.1:8080';
+const WORLD = process.env.WORLD || 'Gilgamesh';
+const TIMEOUT_MS = Number(process.env.TIMEOUT_MS || 60000);
+const SETTLE_MS = Number(process.env.SETTLE || 9000);
+
+// Widths worth guarding. 360 is the common Android floor and 393/414 cover
+// current iPhone and Pixel classes. 320 is deliberately absent: with the
+// shell's 48px of padding it leaves 256px for the row, which the controls
+// cannot fit even icon-only, and failing there would just be noise.
+const WIDTHS = [360, 393, 414];
+
+// Enough active filters that the chip strip is populated, since it shares the
+// bar with the controls under test.
+const ROUTE = `/flip-finder/${WORLD}?profit=10000&roi=50&ppd=5000`;
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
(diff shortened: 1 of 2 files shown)


akarras:
Heads up @akarras — **this duplicates #1082**, which was opened a few minutes earlier from a parallel session and is the better fix. They conflict: both rewrite row 1 of the sticky bar in `analyzer.rs`, `saved_views.rs` and `realtime_status.rs`.

We independently reached the same diagnosis (row 1 can neither wrap nor scroll, so it must fit; both even added the same `compact` prop to `RealtimeStatus`), but #1082 goes further in two ways that matter:

- It makes the controls **shrink and truncate**, with the breakpoint only as a first line of defence. Mine relies on the `sm` breakpoint alone, and #1082 is right that a breakpoint alone is not enough — the side nav claims 240px at `lg`, so row 1 is barely wider at 1024px than at 768px, and the German labels need more than that band gives. My version would re-break there.
- It also fixes an unrelated desktop overflow from `html { width: 100vw }`.

So: **merge #1082, not this.** I'm reducing this PR to the one piece that does not overlap and that #1082 explicitly left out — a runnable regression guard.

`integration/flip-finder-mobile-bar.cjs` asserts at 360 / 393 / 414 that the bar does not overflow its box and that every top-row control is hit-testable at its own centre (`elementFromPoint`), so a control rendered past the viewport edge fails loudly instead of silently disappearing. It is red against production:

```
[info] 360px: 4 control(s), row overflow 168px, bar overflow 160px
[fail]  - 360px: "Columns" ends at x=419, past the viewport edge
        - 360px: "Clear all filters" ends at x=496, past the viewport edge
```

and green against a locally built fix:

```
[info] 360px: 4 control(s), row overflow 0px, bar overflow 0px
[info] 393px: 4 control(s), row overflow 0px, bar overflow 0px
[info] 414px: 4 control(s), row overflow 0px, bar overflow 0px
[done] control bar fits, and every control is reachable, at every phone width tested
```

Note it only covers phone widths. #1082's harder case — the 768–1024px band with German labels — is not exercised, because switching locale in the harness needs the in-app picker rather than a cookie. Worth adding later.

PR #1086, opened 2026-08-02
feat(flip-finder): land a bare URL on a default view (Realistic flips, or the user's own)

### LINKED ISSUE #1070: flip finder does not use a default view
We should default the user to realistic flips, or if possible, let the user save a default filter. That might need to be saved in the db

akarras:
Fixes #1070

## What changed

Opening the Flip Finder with no filters landed you on the raw list, topped by items whose single sale last month was a fluke. Now a bare `/flip-finder/{world}` seeds a whole **view**: the user's own default if they saved one, otherwise the **Realistic flips** preset that already existed in the Views menu (`?min-buy=5000&last-sold=1d&roi=30&sort=profit-per-day`).

Two pieces:

1. **`seed_flip_finder_default_view()`** (`query_defaults.rs`) — seeds the default view onto a bare URL and returns whether the URL *was* bare, so the caller skips the per-param `next-sale` seed it would otherwise run.
2. **A "Make this my default view" checkbox** in the existing Save-view popover, next to "Pin to this world". Persists the current query string to `localStorage["ultros.flipfinder.default_view"]`.

**No database needed.** The issue speculated it might be; the filter state already *is* the URL query string, so a default is one more localStorage string alongside the saved-views list.

## Behavior

| URL | Result |
|---|---|
| `/flip-finder/Gilgamesh` (bare) | Seeds the default view as removable chips |
| `/flip-finder/Gilgamesh?roi=50` | Honored verbatim + the existing `next-sale=1d` seed. Unchanged. |
| Clear All | Stays cleared |

The seed and the `next-sale` seed are mutually exclusive. A view already carries its own recency filter (`last-sold=1d` — enforced by an existing test on every built-in), so layering `next-sale` on top would narrow a view the user picked verbatim.

The checkbox states a fact — "this query is my default" — rather than firing an action. It initializes from storage when the popover opens, and committing it *unchecked* clears the default when it was this query (and leaves a different default alone). Without that, a default could be set but never unset.

## Two traps this had to avoid

**Clear All must stay cleared.** `seed_query_default` fires whenever its key is absent, so a whole-view default written the same way would re-apply itself the instant you hit Clear All. Instead, emptiness is decided **synchronously at component setup**, before any seeding effect has written, and the seeding effect subscribes to nothing — so it runs exactly once per route mount and never re-reads the query.

**Seeding must not use `use_navigate`.** Closed PR #1051 hit this and only caught it in a live browser: `AnalyzerWorldNavigator` (`analyzer.rs:3003`) rebuilds the URL from an **untracked** query snapshot in its own mount-time effect, so a competing `navigate` is simply overwritten and the params never land. This seeds through `query_signal` setters instead. Confirmed against `leptos_router` 0.8.15 source: `RouterContext::navigate` (`components.rs:154-168`) drains `query_mutations` and replays them onto whatever URL it is navigating to — so the seeded params survive regardless of which effect fires first.

## Notes

- Values are decoded with the router's own `Url::unescape` before being written back through a setter, which re-escapes on serialization. Without that, a saved `?name=Grade%208` would gain a `%25` on every visit.
- An empty stored default is a real answer ("land me on the whole list"), not a missing one — only an *absent* key falls back to Realistic flips.
- `analyzer_make_default_view` added to all 7 locales with real translations.

## Testing

- **`./check_ci.sh` green** (`cargo fmt --all -- --check` + `cargo clippy --all-targets -- -D warnings`).
- **378 `ultros-app` lib tests pass** locally (CI doesn't run `cargo test`), 7 of them new. The seeded set is *derived* from the built-in preset and asserted against it, so editing the preset moves the default with it — a hardcoded copy would rot on the first tweak.
- **`cargo check -p ultros-app --features hydrate --target wasm32-unknown-unknown` passes** — the `#[cfg(not(feature = "ssr"))]` storage code is never type-checked by CI's ssr-only clippy, so it was checked by hand.
- **Not browser-verified.** Another session was mid-build on this machine and a local `cargo leptos serve` would have contended for the shared target dir. The one thing a browser would have caught — the `use_navigate` race above — was instead resolved by reading the router's navigate path directly. Worth a quick manual pass on: the checkbox round-trip, and that Clear All stays cleared.


akarras:
Merge-order note: #1082 and #1083 also touch `saved_views.rs` + `analyzer.rs`. I test-merged both against this branch — **both merge clean, textually and semantically**. They restyle the two sticky-bar *buttons* (icon-collapse for mobile); this PR changes the Save-view button’s `on:click` body and adds a checkbox inside the popover. Different lines, no shared symbols or state, so merge order between them and this PR does not matter.

(#1082 and #1083 do overlap with *each other* — same mobile-overflow problem, two different solutions — but that is independent of this PR.)

PR #1108, opened 2026-08-02
🛡️ Sentinel: [HIGH] Fix XSS vulnerability in item_view JSON-LD

akarras:
🚨 Severity: HIGH
💡 Vulnerability: XSS via unescaped JSON in `<script type="application/ld+json">` tags. The `serde_json` output includes unescaped characters like `<` and `>`. If an attacker controls part of the URL (e.g. `world_val`) they can inject `</script>` to break out of the JSON-LD script block and execute arbitrary Javascript.
🎯 Impact: Attackers could execute arbitrary scripts on the user's browser, potentially stealing sessions or performing unauthorized actions.
🔧 Fix: Added a helper function to escape characters like `<`, `>`, `&`, and line terminators into unicode escape sequences (`\u003c`, etc.) before injecting the JSON string into `inner_html`.
✅ Verification: Tested locally that `escape_for_script_tag` escapes `</script>` safely without breaking the JSON output, and `cargo clippy` and `cargo fmt` passed.

---
*PR created automatically by Jules for task [5247550775441886016](https://jules.google.com/task/5247550775441886016) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/lib.rs
+++ b/ultros-frontend/ultros-app/src/lib.rs
@@ -9,7 +9,6 @@ pub(crate) mod math;
 pub(crate) mod query_defaults;
 pub(crate) mod routes;
 pub(crate) mod sales_cadence;
-pub mod script_escape;
 pub(crate) mod ws;
 
 include!(concat!(env!("OUT_DIR"), "/i18n/mod.rs"));
@@ -23,7 +22,6 @@ use crate::global_state::{
     side_nav::provide_side_nav_settings, theme::provide_theme_settings,
     toasts::provide_toast_context, xiv_data::provide_xiv_data_revision,
 };
-pub use crate::script_escape::escape_for_script_tag;
 use crate::{
     components::{
         app_shell::AppShell, on_hand_input::provide_on_hand_context, patreon::*, toast::*,
@@ -683,3 +681,4 @@ mod error_filter_wiring {
         assert!(FILTER_JS.contains("ULTROS_THIRD_PARTY_SCRIPT_HOST_RE"));
     }
 }
+pub mod script_escape;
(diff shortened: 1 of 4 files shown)


akarras:
Confirmed real, exploitable, and now fixed — with one change on top.

**Verified the vulnerability in production first.** `GET /item/Gilgamesh%3C%2Fscript%3EULTROSMARKER/12345` serves a literal `</script>` inside the JSON-LD:

```
"item":"https://ultros.app/item/Gilgamesh</script>ULTROSMARKER/12345"
```

In a real browser against prod: the JSON-LD `textContent` is **truncated** at `Gilgamesh`, `JSON.parse` on it **fails**, and the marker renders into `document.body` as 3 live text nodes. Swap the marker for `` and it executes.

Worth noting the severity is if anything understated: the server **percent-decodes** the path segment, so the encoded form above works from an ordinary clickable link. This does not require a hand-crafted raw-byte request, and `world` is never validated against the world list — `ItemView` only gates on the item id. It also silently breaks the breadcrumb structured data, so it was an SEO bug riding along.

**The one change:** the PR added a second, byte-identical copy of `escape_for_script_tag` — the original already guards the SSR bootstrap blob in `ultros/src/leptos.rs:38`. Two copies of an escaping routine is how one of them quietly stops matching the other, so I moved it to `ultros-app/src/script_escape.rs` as the single implementation; `ultros` picks it up via its existing `use ultros_app::*` and its own test still passes against it.

**Tests** go against `build_breadcrumb_json_ld` itself, not just the helper, since that is the function attacker input actually reaches — including a hostile `world` and a hostile category name. They assert no `<`/`</` survives *and* that the value round-trips back to the original URL, so escaping can never silently break structured data.

Rendered the real fixed output in a browser to confirm end-to-end: tag no longer closes early, `JSON.parse` succeeds, zero injected scripts, nothing leaks into the DOM, and the decoded URL is byte-identical to the original.

Gates: `cargo fmt --all -- --check` clean, `cargo clippy -p ultros -p ultros-app --all-targets -- -D warnings` clean, `cargo test -p ultros-app --lib` 433 passed. (CI does not run `cargo test`, so the suite was only ever going to be verified locally.)

PR #1112, opened 2026-08-02
fix(nav): hide the footer badges when the sidebar is collapsed

akarras:
## What

Two changes to the sidebar's bottom stack:

1. **The account row is now the bottom-most control.** `<AccountMenu />` moved below `.side-nav-footer`.
2. **The Discord and GitHub badges only show when the sidebar is expanded.** `.app-shell-collapsed .side-nav-footer` is now `display: none`.

## Why

Collapsed, the sidebar is 56px. Minus `.side-nav-footer`'s `0.75rem` side padding, that leaves 32px of content width — which the Discord + GitHub icon pair overflows, so GitHub rendered half-clipped by the sidebar's right edge.

The row's third child, the version hash, was *already* hidden at that width (`.app-shell-collapsed .side-nav-version { display: none }`), so the footer had nothing left worth showing in the collapsed state. Hiding the whole row is a smaller rule than the alternative and removes the overflow at its source.

## Reviewer notes

- **Why hide rather than stack.** Stacking the two badges vertically would keep them reachable while collapsed, but it adds ~24px of chrome to the mode whose entire purpose is minimal width. Both links are one click away via the collapse toggle. Happy to switch to a stacked layout if you'd rather not lose them.
- **Mobile is unaffected.** `.app-shell-collapsed` stays on the shell below 1024px, but every collapse rule lives inside `@media (min-width: 1024px)` and the drawer renders at 280px — so the drawer keeps both badges. Explicitly measured (table below).
- **The drop-up's `bottom` had to move.** `.app-shell-collapsed .side-nav-account-panel` uses `position: fixed` to escape `.side-nav`'s scrollport, so its offset is measured from the viewport, not from its parent. It previously had to clear the footer row *and* the trigger row; now the account row is bottom-most and the footer is gone at that width, so it only clears the trigger. `6.5rem` → `3.5rem`. The comment above it was updated to match.
- The account panel in the expanded state is still `position: absolute` / `bottom: calc(100% + 0.35rem)` and needed no change — it grows upward, and sitting lower in the sidebar only gives it more room above.

## Verification

I built the real stylesheet with the cargo-leptos tailwind binary (`tailwindcss-v4.1.10 -i style/tailwind.css`) and measured the sidebar markup against it in a browser — rects, not eyeballing:

| | collapsed (56px) | expanded (240px) | mobile drawer (280px) |
|---|---|---|---|
| footer `display` | `none` | `flex` | `flex` |
| GitHub past nav's right edge | — | no (468 < 640) | no (68 < 280) |
| account row bottom-most | yes | yes | yes |
| drop-up overlaps trigger | no (6.4px gap) | no (5.6px gap) | — |
| drop-up off-screen top | no | no | — |

`cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` both exit 0.

## Base

Branched from `d900a040`, which was 28 commits behind `main`. Four of those commits touched these same two files (#1082, #1085, #1091, #1092), so the branch was reset onto `d5599a92` and the change re-applied before this PR. Both files applied cleanly and the browser verification above was re-run against the rebased stylesheet. The footer/account structure was unchanged by those commits, and #1091's new what's-new badge lives in `.side-nav-sections`, not the footer.

### DIFF
--- a/style/tailwind.css
+++ b/style/tailwind.css
@@ -1938,8 +1938,14 @@ a:not(.nav-link):not(.side-nav-item):not(.menu-item):not(
 @media (min-width: 1024px) {
     .app-shell-collapsed .side-nav-label,
     .app-shell-collapsed .side-nav-section-header,
-    .app-shell-collapsed .side-nav-brand-text,
-    .app-shell-collapsed .side-nav-version {
+    .app-shell-collapsed .side-nav-brand-text {
+        display: none;
+    }
+    /* The whole footer goes, not just the version hash: 56px minus the row's
+       0.75rem side padding leaves 32px, which the Discord + GitHub pair
+       overflows — GitHub was clipped by the sidebar's right edge. Both links
+       are reachable again the moment the sidebar is expanded. */
+    .app-shell-collapsed .side-nav-footer {
         display: none;
     }
     .app-shell-collapsed .side-nav-item {
@@ -2067,15 +2073,16 @@ a:not(.nav-link):not(.side-nav-item):not(.menu-item):not(
            the viewport instead. `left` is unchanged in value: `.side-nav`
            is itself pinned to the viewport's left edge, so 0.35rem means
            the same thing whether measured from the sidebar or the
-           viewport. `bottom` is re-derived from the viewport instead,
-           approximating the collapsed footer row (~3.06rem: 2 x 0.75rem
-           padding + a 1.5rem icon row + border) plus the collapsed
-           trigger row (~3.1rem worst case, sized to the 32px avatar) plus
-           the same 0.35rem gap the expanded state uses above the trigger. */
+           viewport. `bottom` is re-derived from the viewport instead: the
+           account row is the bottom-most thing in the collapsed sidebar
+           (the footer is hidden at this width), so it only has to clear
+           the collapsed trigger row (~3.1rem worst case, sized to the
+           32px avatar) plus the same 0.35rem gap the expanded state uses
+           above the trigger. */
         position: fixed;
         left: 0.35rem;
         right: auto;
-        bottom: 6.5rem;
+        bottom: 3.5rem;
         width: 15rem;
     }
 }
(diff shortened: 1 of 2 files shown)


akarras:
Reviewed and independently verified — approving once the build job lands.

I rebuilt the stylesheet from this branch with the same tailwind binary (`tailwindcss-v4.1.10 -i style/tailwind.css`), pulled the live sidebar markup from prod, applied this PR's DOM reorder, and measured rects in a browser. **Your verification table reproduces exactly**, including both drop-up gaps:

| | collapsed (56px) | expanded (240px) | drawer (<1024px) |
|---|---|---|---|
| `.side-nav-footer` display | `none` | `flex` | `flex` |
| Discord/GitHub rendered | no | yes | yes |
| GitHub past nav right edge | — | no (68 < 240) | no (68 < 280) |
| account row is last child | yes | yes | yes |
| drop-up `position` / `bottom` | `fixed` / `56px` | `absolute` / `43.58px` | — |
| drop-up gap to trigger | **6.4px** | **5.6px** | — |
| drop-up off-screen top | no | no | — |

Checks that mattered to me beyond the table:

- **Dropping `.side-nav-version` from the collapsed `display:none` list is not a regression** — it lives *inside* `.side-nav-footer`, so hiding the footer hides it too. Confirmed it stops rendering collapsed.
- **The `6.5rem` → `3.5rem` change is load-bearing and correct.** The panel is `position: fixed` collapsed, so its offset is from the viewport; with the footer gone and the account row bottom-most it only has to clear the trigger. At `6.5rem` it would now float a footer's height too high. Measured `bottom: 56px`, top at 624 of a 900px viewport — clears the trigger by 6.4px and never leaves the screen.
- **Mobile is genuinely unaffected.** `.app-shell-collapsed` does stay on the shell below 1024px, but every collapse rule is inside `@media (min-width: 1024px)`, so the 280px drawer keeps both badges. (Worth noting `window.innerWidth` misreports under viewport emulation here — I confirmed the branch via the 280px aside width, not the reported viewport.)
- No new user-facing strings, so no locale files needed.
- Base is one commit behind `main` and touches no files that commit touched, so no stale-base risk.

On the open question in your reviewer notes — hiding rather than stacking is the right call. Collapsed mode exists to be 56px wide; spending 24px of it on two links that are one toggle away is a bad trade.

PR #1113, opened 2026-08-03
⚡ Bolt: O(N) top items extraction in alert drawer

akarras:
### 💡 What: 
Refactored the item matches search algorithm in `alert_drawer.rs` to use `select_nth_unstable_by_key` to filter the top 50 search results before fully sorting them.

### 🎯 Why: 
Sorting the full list of `matches` before taking the top 50 requires `O(N log N)` computational time, which can trigger jank on the UI thread as the user types because this is calculated on every stroke for large matches arrays. Using `select_nth_unstable_by_key` filters down the `matches` list to the top 50 in linear `O(N)` time.

### 📊 Impact: 
Significantly reduces time complexity and execution time for the alert drawer's autocomplete search bar, mitigating UI thread freezes during typing.

### 🔬 Measurement: 
Profile the CPU usage and responsiveness when typing broad queries into the alert drawer item autocomplete input before and after.

---
*PR created automatically by Jules for task [9705930609568201001](https://jules.google.com/task/9705930609568201001) started by @akarras*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -58,3 +58,6 @@ When returning the top N items from a collection, first use `select_nth_unstable
 **Learning:** In `ultros-frontend/ultros-app/src/components/live_sale_ticker.rs`, a list of recent sales was fully sorted using `sorted_by_key` before taking the first 8 elements. Because sorting is an `O(N log N)` operation and the number of recent sales can be large, finding the top 8 elements in linear time `O(N)` reduces the time complexity and prevents UI jank on the main thread in WASM.
 
 **Action:** When finding the top N elements in a collection, use `select_nth_unstable_by_key` to partition the elements in linear time, followed by `truncate(N)` and `sort_unstable_by_key` on the remaining K elements.
+## 2024-11-20 - O(N) Top-K Extraction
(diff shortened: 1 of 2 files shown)


akarras:
Verified — merging.

Same `select_nth_unstable` + `truncate` + `sort_unstable` shape I verified equivalent for #992 with a standalone harness (24 configs: boundary sizes 49/50/51, large N, and degenerate all-ties). The correctness preconditions hold here too:

- `select_nth_unstable_by_key(50, …)` is guarded by `if matches.len() > 50`, so the index is always < len and can't panic.
- The partition key and the final sort key are the same (`Reverse(i.level_item)`), so the partition selects exactly the set the sort then orders.
- Dropping `.take(50)` is safe: `truncate` caps the >50 path and the ≤50 path is already short.

**One thing I checked specifically, given #998.** `matches` is built from `tracked_data().items`, which is a `HashMap<ItemId, Item>` — and `level_item` ties are common, so which items survive the top-50 cut depends on iteration order, which `RandomState` randomizes per process. That's the exact shape of the `/scrip-sources` hydration bug.

It's **not** a bug here: `search` initializes to `""` and the closure returns `Vec::new()` on an empty query, so the dropdown renders nothing during SSR and the HashMap order never reaches server-rendered output. The tie-arbitrariness is also not a regression — the previous stable `sort_by_key` preserved a *randomized* input order, so ties were already arbitrary across processes.

Worth knowing for later though: this is one line away from being the #998 bug. If `search` ever gets seeded from a query param or a prop so the dropdown can render server-side, this list would need a stable tie-break on `item_id` the way `rank_scrip_sources` now does.

Lint/clippy/fmt and build both green.

PR #1114, opened 2026-08-03
feat(item-explorer): replace the subcategory popover with an inline accordion

akarras:
The `/items/*` subcategory picker forked two ways: an overlay popover (Weapons, Job Sets, and any group with >8 categories) and an inline horizontally-scrolling chip strip for the rest. The popover trigger read as a rough dropdown affordance for what is really the page's primary navigation, and having two different interactions depending on which pill was selected compounded it.

All five groups now render one inline accordion. It expands when a group pill is clicked, collapses when a subcategory is chosen, and starts expanded only on the bare `/items` route where nothing is selected yet.

## What changed

- `grouped_nav_popover.rs` → `grouped_nav_accordion.rs`. The overlay machinery is gone: no `absolute`/`z-[100]`, no `on_click_outside`, no Escape handler, no `NodeRef`.
- The toolbar's three-way branching collapses to one path. `group_uses_popover` and the `>8 chips` special case are deleted, along with the `item-explorer-chip-row` utility that had become unreachable.
- `role_buckets` extracted from two duplicated ~25-line bucketing loops (Weapons and Job Sets each had their own copy).
- Zero new i18n keys — every label reuses an existing one, so no locale files were touched.

## Notes for review

**Hydration.** This route has history with tachys `failed_to_cast_element` panics, so the accordion never mounts or unmounts its panel: children are always rendered and only `aria-expanded`/`data-open` change, with expansion animated via a `grid-template-rows: 0fr → 1fr` track. This is also strictly safer than what it replaces — the old `{move || …}` returned structurally different view trees from three branches, so an SSR/CSR disagreement about the selected group produced different *kinds* of tree. Now there is one shape.

**`location.pathname.track()` in the collapse effect is load-bearing.** `active_group` is a `Memo`, and a memo notifies only when its value changes. Navigating `/items/category/24` → `/items/category/25` leaves it at `Some(2)`, so tracking it alone left the accordion expanded after a subcategory had been chosen. Please don't remove the pathname track as redundant — there's a comment saying so.

**Three defects came from CSS that was reasoned about rather than measured**, each caught by a later round:
- `overflow: hidden` clips but doesn't hide — collapsed chips stayed in the tab order and the a11y tree, contradicting the header's own `aria-expanded="false"`. Fixed with `visibility`.
- `min-height: 0` zeroes the content box, not the padding box — the collapsed panel measured 14.4px, not 0.
- The wrapper's `overflow: hidden` clipped the global `:focus-visible` ring's positive `outline-offset` entirely, leaving this control with **no visible keyboard focus indicator at all**. Fixed with `outline-offset: -2px`.

## Testing

`cargo test -p ultros-app --lib` — 369 passed, 0 failed. `./check_ci.sh` — exit 0. Note that CI does not run `cargo test` for this repo, so green CI here proves only that it compiles.

Verified against a served build of this branch (confirmed via the footer version link, not assumed): pill-click opens, chip-click collapses, same-group chip→chip collapses, deep link starts collapsed, back to `/items` re-opens, header toggles, focus ring visible, no hydration panic in the console on any page, and the Job Sets panel flows with the page at 375px rather than trapping a nested scrollbar.

## Out of scope, filed separately

`role_for_job_abbr` matches on `job.abbreviation`, which is a **localized** game-data column. If non-English abbreviations differ, every job falls to the `Other` bucket for those users — and since the server renders in English while the client hydrates in the visitor's locale, that would be an element-count divergence on the default Weapons tab. Pre-existing (the popover bucketed identically), not a regression from this PR, but worth a multi-locale test to confirm.


akarras:
**Browser-verified against a served build of the merged branch** (build `b4f3e7b8` confirmed from the page chrome, not assumed). Merging.

## Verified

| behaviour | result |
|---|---|
| Bare `/items` starts expanded | ✅ `aria-expanded="true"`, panel 721px |
| Group pill click forces open | ✅ Job Sets pill expands |
| Subcategory chip click collapses | ✅ navigates to `/items/jobset/PLD`, header → `false` |
| Deep link starts collapsed | ✅ `/items/category/2` → `aria-expanded="false"`, panel 0px, header reads "Primary Tools" |
| Collapsed chips leave the a11y tree | ✅ **0 of 34** chips visible; `inner` computes `visibility: hidden` |
| Hydration | ✅ `boot-progress` absent on every route and after client-side nav |
| Role bucketing | ✅ all 7 sections render (Tank / Healer / Melee / Phys Ranged / Magical Ranged / Hand / Land) |
| Job hrefs | ✅ 34 links, **all** canonical acronyms (`PLD`, `WAR`, `GNB`, …), zero localized or percent-encoded slugs |

## On the merge

I resolved this conflict independently before noticing `b4f3e7b8` had already landed, and we reached the same resolution — worth stating because the conflict had a trap. A mechanical "take ours" on the render block would have silently reverted **two** fixes main landed after this branch forked:

- the jobset href built from the *localized* label instead of `job_chip_slug(job)` — a German client's "FST" chip navigating to a route `job_category_lookup` cannot resolve;
- `role_for_job_abbr(&job.abbreviation)` instead of `role_for_job(job)` — documented on main as bucketing 23 of 36 German and 22 of 36 French jobs into `Other`.

Both survived here. `role_grouping_is_identical_in_every_locale` is a good addition — and its observation that `every_visible_job_maps_to_a_role` *passed while the grouping was broken*, because the test process loads English data, is the kind of thing that makes the guard actually load-bearing.

Taking the `grouped_nav_popover.rs` delete over main's modification (#1094) is also correct: `item_explorer_toolbar.rs` was its only consumer, and `use_dismissable` — the shared hook #1094 extracted — survives in 5 other components.

`cargo test -p ultros-app --lib` → **433 passed, 0 failed** (CI runs no tests, so this matters), clippy `--all-targets -D warnings` clean, fmt clean.

## Measurement note for the next reviewer

The in-app browser pane runs with `document.hidden = true`, which **fully throttles rAF — 0 frames in 2 seconds**. CSS transitions therefore never advance, and the collapsed panel reads frozen at its *start* value (555px, `visibility: visible`) even though the state is correct. I nearly filed that as a bug. Forcing `transition: none` before reading computed styles snaps everything to the true target (0px / hidden). Any transition-dependent assertion in this harness needs that, or it measures the wrong thing.

PR #1115, opened 2026-08-03
fix(trends): rank the ClickHouse candidate set by recent sales, not item id

### LINKED ISSUE #1081: Market Trends only ever considers the 1500 lowest item ids
### Summary

`AnalyzerService::get_trends_v2` selects the items it asks ClickHouse about with a bare `.take(1500)` over `cheapest.item_map`:

```rust
// ultros/src/analyzer_service.rs:1112-1119
const MAX_TUPLES: usize = 1500;
let mut requests: Vec<(i32, u8, i32)> = cheapest
    .item_map
    .iter()
    .take(MAX_TUPLES)
    .map(|(key, _)| (key.item_id, key.hq as u8, world_id))
    .collect();
```

`item_map` is a `BTreeMap<ItemKey, _>` and `ItemKey` orders by `item_id` first, so `.take(1500)` is not an arbitrary sample — it is deterministically **the 1500 lowest item ids on that world**. Everything above that id cutoff is never sent to ClickHouse and therefore can never appear on Market Trends, no matter how actively it trades.

### Evidence from prod

`GET /api/v1/trends/Gilgamesh?window=30&show_suspicious=false` returns 63 items whose ids span **2 – 4650** — i.e. entirely early-ARR items. Since the cap binds, no item with an id past that window is ever considered, which silently excludes all post-ARR content (ids ~10000+).

The user-visible consequence is the confusing half of #1053: picking a category like Pugilist's Arms yields an empty table, and it reads like a broken filter rather than "the backend never offered these items."

### Proposed fix

Make the selection deterministic *and* meaningful instead of id-ordered: rank `item_map` by a relevance signal (e.g. recent sale volume or listing count) and take the top N, rather than taking whichever ids sort lowest. If the cap exists to bound SQL size, ranking keeps that bound while making the retained subset the part users actually care about.

### Also worth considering

The endpoint intermittently returns `{}` in prod — repeated identical requests alternate between a full 28KB payload and a 2-byte `{}`, and SSR HTML alternates between 63 rendered rows and 0. That looks like one replica answering before `analyzer_service` has warmed (`if !self.initiated { return None }`), and it makes the page look randomly empty on load. May deserve its own issue.

Split out of #1053, which fixes the frontend half (filters being wiped from the URL).

akarras:
Fixes #1081.

## The bug

`get_trends_v2` built its ClickHouse request set like this:

```rust
let mut requests: Vec<(i32, u8, i32)> = cheapest
    .item_map
    .iter()
    .take(MAX_TUPLES)   // 1500
```

`item_map` is a `BTreeMap<ItemKey, _>` and `ItemKey` orders on `item_id` first, so
"take the first 1500" means **take the 1500 lowest item ids in the game**. Everything
above that id boundary can never appear on Market Trends no matter how briskly it
trades. That is not a truncation of a ranked list — the cap *is* the ranking, and it
ranks by "how early in FFXIV's history was this item added".

## Measured on prod, just now

Same world, same instant. `/api/v1/movers/{world}` is a clean control: it queries
ClickHouse **directly** and never goes through the analyzer's candidate selection.

| endpoint | rows | max item id | ids > 10000 |
| --- | --- | --- | --- |
| `/api/v1/trends/Gilgamesh?window=30` | 500 | **4661** | **0** |
| `/api/v1/movers/Gilgamesh` | 10 | **51958** | 8 |

Trends is returning a *full* 500-row page — it just cannot see past item id ~4700.
Sargatanas is the same shape (500 rows, max id 4700, zero above 10000).

The sharpest version: of the true **top 50 items by unit volume** on Gilgamesh
(`/api/v1/movers/Gilgamesh?direction=volume&limit=50`, straight out of ClickHouse),
**35 have ids above 4700** — so 70% of the highest-volume items on the world are
structurally invisible to a page whose default sort is *"units traded, descending"*.

Because `movers` reads the same rollup, this also confirms the rollup **does** have
coverage for those high-id items. Trends simply never asked about them.

## The fix

Rank the candidates by the signal the page is actually about — has this item sold
recently — and *then* apply the cap. That signal is already in-process: the analyzer's
per-world sale buffer (`recent_sale_history`). No extra query, no schema change.

This is also what the v1 `get_trends` path has always done — it drives off the sale
history and looks the cheapest price up, rather than the other way round. v2 regressed
the driver when it moved to ClickHouse aggregates.

The policy lives in a new `ultros/src/trend_candidates.rs`, split out for the same
reason `resale_eligibility.rs` was: so it is unit-testable without standing up an
`AnalyzerService`. Ordering is `(last_sale desc, buffered_sales desc, item_id, hq)` —
a **total** order, so two replicas answering the same request pick the same items.

Two properties worth calling out:

- **It cannot regress a cold world.** Items the sale buffer has never seen sell sort
  last, among themselves by ascending item id — exactly today's behaviour. A world
  whose buffer is still cold behaves identically to before. There is a test for this.
- **It stays O(n).** `select_nth_unstable_by` partitions before sorting, so a board of
  tens of thousands of listed items doesn't pay a full sort.

`MAX_TUPLES` stays at 1500. The cap was never the problem; what it selected was.

Also in this diff: the sale-history lock is taken in the **same order** as `get_trends`
(sale history, then cheapest) so the two paths can't deadlock against each other, and
it is explicitly dropped before the ClickHouse round-trips — live sale ingestion needs
that lock to write and a deep scan is slow.

## Verification

⚠️ Read this bit rather than assuming the tests in the diff ran in CI — they did not.

`ultros` is a bin-only crate (`main.rs`, no `lib.rs`), so its test binary does not link
on Windows/MSVC (the pre-existing `LNK1120 ... _10ultros_app` wall), and CI has the
test step commented out in `rust.yml`. So I made the new module **dependency-free**
(`chrono` + std only) and ran its tests through a standalone harness crate that
`#[path]`-includes the real shipped file — meaning the tests below are literally the
tests in this diff, executed against the file as committed, not a copy.

Genuine red → green, one behaviour at a time:

- **RED** (module implemented with today's `take(max)` semantics): `REAL_EXIT=101`,
  **5 failed / 3 passed**. The headline test failed with the exact prod symptom —
  `the only item that actually traded was dropped in favour of two dormant low-id
  items; picked [2, 3]`. The 3 that passed are the preserved-behaviour tests, which
  should pass on both sides of the fix.
- **GREEN** (after the ranking fix): `REAL_EXIT=0`, **8 passed / 0 failed**.

Gates:

- `cargo fmt --all -- --check` → exit 0
- `cargo clippy -p ultros --all-targets -- -D warnings` → exit 0 (this pulls in
  `ultros-app`, `ultros-db` and `ultros-clickhouse`, so the whole `ssr` path
  type-checks). Full-workspace clippy not run; the diff is confined to the `ultros`
  crate.

## What this does not do

It does not touch the **second** half of #1081 — prod flapping between a populated
payload and `{}` (one replica answering before `analyzer_service` warms). That is a
separate fix and the issue should stay open for it.

### DIFF
--- a/ultros/src/analyzer_service.rs
+++ b/ultros/src/analyzer_service.rs
@@ -27,6 +27,7 @@ use ultros_db::{
 use universalis::{ItemId, WorldId};
 
 use crate::event::{BusRecv, EventReceivers, handle_bus_recv};
+use crate::trend_candidates::{TrendCandidate, select_trend_candidates};
 use thiserror::Error;
 use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
@@ -1101,21 +1102,47 @@ impl AnalyzerService {
         if !self.initiated.load(Ordering::Relaxed) {
             return None;
         }
+        // Acquired in the same order as `get_trends` (sale history, then
+        // cheapest) so the two paths cannot deadlock against each other.
+        let sale_history = self.recent_sale_history.get(&world_id)?.read().await;
         let cheapest = self
             .cheapest_items
             .get(&AnySelector::World(world_id))?
             .read()
             .await;
 
-        // Build the request tuple list from every item the cheapest map
-        // knows about on this world. Cap to a sane upper bound so a fresh
-        // world with thousands of listings doesn't blow up the SQL.
+        // Build the request tuple list from the items the cheapest map
+        // knows about on this world, capped to a sane upper bound so a
+        // world with tens of thousands of listings doesn't blow up the
+        // SQL. Which items make the cut is a real decision rather than a
+        // truncation: the map is keyed on item id, so taking the first N
+        // would ask only about the N lowest ids in the game.
         const MAX_TUPLES: usize = 1500;
-        let mut requests: Vec<(i32, u8, i32)> = cheapest
-            .item_map
-            .iter()
-            .take(MAX_TUPLES)
-            .map(|(key, _)| (key.item_id, key.hq as u8, world_id))
+        let candidates = select_trend_candidates(
+            cheapest
+                .item_map
+                .keys()
+                .map(|key| {
+                    let buffered = sale_history.item_map.get(key);
+                    TrendCandidate {
+                        item_id: key.item_id,
+                        hq: key.hq,
+                        // `SaleHistory::add_sale` keeps the buffer sorted
+                        // newest-first, so the head is the latest sale.
+                        last_sale: buffered.and_then(|s| s.first()).map(|s| s.sale_date),
+                        buffered_sales: buffered.map_or(0, |s| s.len() as u8),
+                    }
+                })
+                .collect(),
+            MAX_TUPLES,
+        );
+        // Dropped before the ClickHouse round-trips below — live sale
+        // ingestion needs this lock to write, and a deep scan is slow.
+        drop(sale_history);
+
+        let mut requests: Vec<(i32, u8, i32)> = candidates
+            .into_iter()
+            .map(|c| (c.item_id, c.hq as u8, world_id))
             .collect();
         requests.sort_unstable();
         requests.dedup();
--- a/ultros/src/main.rs
+++ b/ultros/src/main.rs
@@ -12,6 +12,7 @@ pub mod leptos;
 pub mod profiling;
 pub(crate) mod resale_eligibility;
 pub(crate) mod search_service;
+pub(crate) mod trend_candidates;
 pub(crate) mod utils;
 mod web;
 mod web_metrics;
--- a/ultros/src/trend_candidates.rs
+++ b/ultros/src/trend_candidates.rs
@@ -1,0 +1,207 @@
+//! Pure "which items should Market Trends ask ClickHouse about?" policy.
+//!
+//! Split out of `analyzer_service` for the same reason as
+//! `resale_eligibility`: so the choice is unit-testable without standing up
+//! an `AnalyzerService` (whose test binary does not link on Windows).
+//!
+//! Trends can only afford a bounded number of item tuples per request, so
+//! something has to pick. The obvious pick — iterate the cheapest-listings
+//! map and take the first N — is silently catastrophic, because that map is
+//! a `BTreeMap` keyed on item id: taking the first N takes the N *lowest
+//! item ids in the game*. Everything above that id boundary can never
+//! appear on the page no matter how briskly it trades, which is why whole
+//! categories render empty.
+//!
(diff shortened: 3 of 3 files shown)


akarras:
## Reviewed — approving and merging

Independently verified rather than taken on trust:

**Premise reproduced on prod just now** (same shape as the PR body, normal drift):

| endpoint | rows | max item id | |
| --- | --- | --- | --- |
| `/api/v1/trends/Gilgamesh?window=30` | 500 | **4664** | 0 ids > 10000 |
| `/api/v1/movers/Gilgamesh?direction=volume&limit=50` | 50 | **52254** | **36 of the top 50 by volume have ids > 4700** |

Trends returns a *full* page and still cannot see past ~4700, while `movers` reads the same rollup and reaches 52254. Confirms both that the ceiling is real and that the rollup has the coverage.

**Tests run against the file as committed** — extracted the blob from this branch into a standalone `chrono`-only harness, verified byte-identical to the committed file first, since CI has the test step commented out and the `ultros` test binary doesn't link on Windows/MSVC:

- GREEN, as shipped: `REAL_EXIT=0`, **8 passed / 0 failed**
- RED, after reverting only the fn body to the old `truncate(max)` semantics: `REAL_EXIT=101`, **5 failed / 3 passed**, headline failure verbatim — `the only item that actually traded was dropped in favour of two dormant low-id items; picked [2, 3]`

So the tests have teeth, and it is this fix that makes them pass.

**Load-bearing claims checked against the real code, not the description:**

- `buffered_sales: s.len() as u8` is safe — `SALE_HISTORY_SIZE = 6`, so the `ArrayVec` can never reach the 255 truncation point.
- "buffer sorted newest-first, so the head is the latest sale" is true — `add_sale` ends with `entries.sort_by_key(|s| Reverse(s.sale_date))` on every path, so `.first()` really is the most recent sale.
- Lock-order claim holds — `get_trends` takes `recent_sale_history` (:869) then `cheapest_items` (:870); the new code takes them in that same order, so the two paths can't deadlock.
- The new `recent_sale_history.get(&world_id)?` adds **no** new `None` path: both maps are materialised once at startup from the same `world_cache.get_inner_data()` walk, so they cover an identical world set. This does not widen the `{}`-payload flapping the PR explicitly leaves open.
- `select_nth_unstable_by_key(max, ..)` is guarded by `candidates.len() > max`, so the index is always in bounds.

No conflict with #1116 — that PR touches `main.rs` at :258-274 (sentry init); this one at :15 (module decl). Disjoint and order-independent.

PR #1116, opened 2026-08-03
chore(deps): lockfile refresh + sentry 0.48→0.49

akarras:
Recurring `big-dependency-updates` run (2026-08-03).

## Big-dependency ceilings — all UNCHANGED this run
Re-verified live against the crates.io deps API:
- **leptos** stays **0.8.20** (0.9.0-beta only). `leptos_i18n 0.6.2` still pins `leptos ^0.8` **and** `leptos-use ^0.18`, and there is no newer `leptos_i18n` release/prerelease — hard gate holds. All leptos-ecosystem crates still target 0.8.
- **sea-orm** already on 2.0; picked up the **2.0.1** patch in-range. sea-query stays 1.0.1.
- **axum 0.8.9** and **tokio 1.53.1** already latest stable.

## What changed
**In-range `cargo update` refresh:**
- `sea-orm` 2.0.0 → **2.0.1** (transitively pulls `arrow` 57 → 58 via `sea-orm-arrow`; single copy, no duplicate major)
- `rustls` 0.23.42 → 0.23.43, `http` 1.4 → 1.5, `time` 0.3.54 → 0.3.55, `clap`, `tokio-macros`, + ~40 other patch bumps
- `superboring` 0.1.12 → 0.1.14 (transitive TLS backend; adds its RustCrypto AEAD deps — all transitive, nothing we reference)
- **lodestone held at `192faccd`** (the git `async` branch) so the `ultros` reqwest-0.11 client keeps compiling; reqwest split 0.11/0.12/0.13 preserved
- **wasm-bindgen family did not move** → no Dockerfile CLI bump needed

**sentry 0.48 → 0.49** (`sentry`, `sentry-tracing`, `sentry-tower`) — the one real migration available this run:
- 0.49 made `ClientOptions` `#[non_exhaustive]` and dropped the public `traces_sample_rate` field in favor of a builder setter. Migrated `main.rs` init to `ClientOptions::new().traces_sample_rate(..)` + direct assignment of the still-public `release`/`environment`/`attach_stacktrace`/`send_default_pii` fields — **like-for-like** behavior.
- Our `default-features = false` keeps 0.49's new default-on `logs`/`metrics` telemetry **off**.
- `reqwest` stays 0.13 (sentry's transport) — **no new major**. sentry-tower's `NewSentryLayer::new_from_top()` / `SentryHttpLayer::new().enable_transaction()` API is unchanged.

## Deferred (documented reasons, re-confirmed)
- `leptos-use` 0.19 / `tower-http` 0.7 / `itertools` 0.15 — **duplicate-traps** (gated by `leptos_i18n ^0.18` / `leptos_axum ^0.6` / `tantivy ^0.14` respectively; bumping adds a 2nd copy for zero gain).
- `base64` 0.23 — duplicate-trap (transitive deps pin 0.21/0.22).
- `rkyv` 0.8 / `reqwest` 0.13 (for the 0.12 cohort) — risky migrations, deferred.
- `resvg` 0.47 → 0.48 — available, but resvg rasterizes the item-card social PNGs (`ultros/src/web/item_card.rs`) and charts; a usvg/tiny-skia major bump needs **visual** verification that a compile-only run can't provide. Candidate for a focused follow-up.

## Not fixed here
The 4 Dependabot alerts (1 high / 1 mod / 2 low) are the same `rustls-webpki` 0.101.7/0.102.8 residuals gated by `serenity 0.12.5` / `poise 0.6.2` (Discord bot on old rustls). Unfixable by lockfile bump; this PR leaves rustls majors at {0.21.12, 0.22.4, 0.23.43}, untouched.

## Verification (Windows, vendored OpenSSL, Strawberry Perl on PATH)
- `cargo clippy --all-targets -- -D warnings` — **exit 0, zero warnings**, whole workspace incl. the `ultros` server + `ultros-app` + vendored OpenSSL (10m51s)
- `npm audit` (integration/) — **0 vulnerabilities**

No live database or Docker/jemalloc build exercised locally (Windows MSVC can't build jemalloc). Leaving this for the CI Docker job + review before merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,6 +28,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "aead"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1973cfbc1a2daf9cf550e74e1f088c28e7f7d8c1e1418fb6c9dc5184b7e84c99"
+dependencies = [
+ "crypto-common 0.2.2",
+ "inout 0.2.2",
+]
+
+[[package]]
 name = "aes"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -55,11 +65,25 @@ version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
 dependencies = [
- "aead",
+ "aead 0.5.2",
  "aes 0.8.4",
  "cipher 0.4.4",
- "ctr",
- "ghash",
+ "ctr 0.9.2",
+ "ghash 0.5.1",
+ "subtle",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fdf011db2e21ce0d575593d749db5554b47fed37aff429e4dc50bc91ac93a028"
+dependencies = [
+ "aead 0.6.1",
+ "aes 0.9.2",
+ "cipher 0.5.2",
+ "ctr 0.10.1",
+ "ghash 0.6.0",
  "subtle",
 ]
 
@@ -100,9 +124,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
@@ -284,9 +308,9 @@ dependencies = [
 
 [[package]]
 name = "arrow"
-version = "57.3.1"
+version = "58.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3bd47f2a6ddc39244bd722a27ee5da66c03369d087b9e024eafdb03e98b98ea7"
+checksum = "6cfdd0833e32a9874d2b55089333ad310c0be208aafa277385ce2461dec90be3"
 dependencies = [
  "arrow-arith",
  "arrow-array",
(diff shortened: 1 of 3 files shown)


akarras:
## Reviewed — one real regression found and fixed, then merging

The lockfile half checks out: `leptos` stays pinned at **0.8.20** in the new lock (the `leptos_i18n` ceiling holds), no duplicate major introduced by the `sea-orm` 2.0.1 / `arrow` 58 move, and the wasm-bindgen family didn't move so the Dockerfile CLI pin needs no bump.

### The sentry migration was not quite like-for-like

I checked the migration against the vendored 0.49.1 source rather than the description. `ClientOptions::new()` really is `Self::default()`, so the struct-literal → builder rewrite is faithful:

```rust
pub fn new() -> Self { Self::default() }
```

But the setter is not a plain field assignment:

```rust
pub fn traces_sample_rate(self, traces_sample_rate: f32) -> Self {
    if !(0.0..=1.0).contains(&traces_sample_rate) {
        panic!("Traces sample rate {traces_sample_rate} is outside the allowed range [0.0, 1.0].")
    }
```

`GLITCHTIP_TRACES_SAMPLE_RATE` is operator-supplied and parsed loosely (`.parse().ok().unwrap_or(0.0)`). Under 0.48 that value was assigned straight to a public field, so an out-of-range setting was simply tolerated. Under 0.49 it **panics during `sentry::init`** — which in `main` runs *before* the DB is created, so the process dies at boot and Docker restart-loops it.

The trigger is the single most natural typo for something named "sample rate": `GLITCHTIP_TRACES_SAMPLE_RATE=50`, meaning 50%. `NaN` and `inf` are the same hazard — both parse successfully as `f32` in Rust, so they reach the setter and panic too.

Default path is unaffected (env unset → `0.0`), which is why nothing caught it: CI never sets the var, and a compile-only run can't see it.

Fixed in f9d169b9 by clamping instead of passing through, dropping non-finite values first so `clamp` can't be handed a `NaN`:

```rust
let traces_sample_rate = std::env::var("GLITCHTIP_TRACES_SAMPLE_RATE")
    .ok()
    .and_then(|v| v.parse::<f32>().ok())
    .filter(|rate| rate.is_finite())
    .map(|rate| rate.clamp(0.0, 1.0))
    .unwrap_or(0.0);
```

A misconfigured rate now degrades to the nearest valid setting instead of taking the site down.

### Gates

`cargo fmt --all -- --check` → **exit 0** on the fixed tree. Full clippy not re-run locally — a cold build of this branch means the vendored-OpenSSL wall, and the change is four lines inside the expression CI already type-checked green — so I'm relying on the CI clippy run against this commit rather than asserting it myself.

No conflict with #1115 (merged): that touched `main.rs` at :15 (module decl), this at :258+ (sentry init).

PR #1118, opened 2026-08-04
⚡ Bolt: Optimize top N item extraction with select_nth_unstable

akarras:
💡 **What:** Optimized top K element extraction by utilizing `select_nth_unstable_by_key` before sorting the final elements.
🎯 **Why:** Previously, the application was using `sort_by_key` or `sort_unstable_by_key` to fully sort the slice of size $M$ before performing `.truncate(N)` or `.take(N)`, taking $O(M \log M)$ time. `select_nth_unstable_by_key` allows finding the top $N$ elements in linear $O(M)$ time, reducing unnecessary sorting work. 
📊 **Impact:** Provides significant CPU time improvements whenever the limit is significantly smaller than the total results.
🔬 **Measurement:** Verify CPU profile to confirm `sort_by_key` time drops during heavy DB sales history access patterns.

---
*PR created automatically by Jules for task [5315684271983234542](https://jules.google.com/task/5315684271983234542) started by @akarras*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -61,3 +61,8 @@ When returning the top N items from a collection, first use `select_nth_unstable
 ## 2024-11-20 - O(N) Top-K Extraction
 **Learning:** In WASM/frontend contexts, running a full `O(N log N)` sort on search results reactively as the user types can cause unnecessary main-thread blocking. Finding the top K elements can be optimized by using `select_nth_unstable_by_key(K)`, which partitions the array in `O(N)` time, followed by truncating the array to K and performing `sort_unstable_by_key` only on those K elements.
 **Action:** When extracting the top N items from a potentially large collection, avoid full sorts in hot reactive loops. Use `.select_nth_unstable_by_key(N)` followed by `.truncate(N)` and sorting only the remaining elements.
+## 2024-08-04 - Optimize top N item extraction with select_nth_unstable
+**Learning:**
+In `ultros-db/src/sales.rs` and `ultros/src/discord/ffxiv/analyze.rs`, taking the top N elements of a slice using a full sort (`sort_by_key`/`sort_unstable_by_key` followed by `.truncate()` or `.take()`) takes $O(M \log M)$ time. When returning the top N items from a potentially large collection, avoid full sorts.
+**Action:**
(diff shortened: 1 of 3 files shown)


akarras:
Reviewed and verified the top-N rewrite.

**Correctness of the transform.** `select_nth_unstable_by_key(limit)` places index `limit` at its sorted position with everything before it `<=` by key; the key is `Reverse(sold_date)`, so indices `0..limit` are the `limit` newest. `truncate(limit)` keeps exactly those, then the final sort orders them. The `len() > limit` guard also covers the panic condition (`select_nth` panics when `index >= len`), and the `==`/`<` cases correctly fall through to a plain sort.

**Empirical check** — CI does not run `cargo test` (the step is commented out in rust.yml), so a behavior-preserving claim needed evidence rather than reading. I ran the old and new algorithms side by side over 20,000 randomized cases with deliberately saturated tie density (as few as 1 distinct timestamp across up to 500 sales), since FFXIV timestamps are second-granularity and stack purchases produce many sales sharing a second:

```
sold_date multiset differs : 0            <-- the user-visible contract
ordering violations        : 0
SET of rows differs        : 7812/20000   (tie-break only)
```

The returned *timestamps* and their order are identical in every case.

Worth stating plainly though: this is **contract-preserving, not byte-identical**. The old code used stable `sort_by_key`; the new path is unstable, so when ties straddle the cutoff a different sale record carrying the *same* timestamp can be returned. That is legal for "the N most recent sales", and the blast radius is one boundary row out of 200-1000 — it cannot move `compute_summary`'s median meaningfully. Noting it for the record, not as a blocker.

**`limit == 0` divergence.** `get_compact_sale_history` gained a `limit_usize > 0` guard that `get_sale_history_from_multiple_worlds` did not. At `limit == 0` that guard skips the truncate entirely and returns *every* sale instead of none. Unreachable today — the sole caller (`extended_sale_history`) does `.clamp(1, MAX_LIMIT)` — so not a live bug, just an inconsistency between the two sites that matters if that clamp ever moves.

`total_results` is still captured before truncation in `analyze.rs`, so the Discord embed's count is unchanged. Nit, not worth a round trip: the new `.jules/bolt.md` entry is dated 2024-08-04, should be 2026.

Merging.

PR #1119, opened 2026-08-04
fix(vendor-resale): hide listings priced far above what the item actually sells for

akarras:
## The bug

`/vendor-resale/<world>` ranks by ROI against the cheapest **listing**, but a listing only pays out if somebody actually buys it. Gil-trader and troll listings sit orders of magnitude above what an item really clears for, and because they inflate ROI the most, **they sort straight to the top**.

On a live load of `/vendor-resale/Gilgamesh` (default view, ROI-sorted), **every row above the fold was unbuyable**:

| Item | Vendor price | "Market price" | ROI shown | Median recent sale | Ratio |
|---|---|---|---|---|---|
| Copper Wristlets | 229 | 150,000,000 | 62,226,974% | 849 | 176,678× |
| Steel Jig | 384 | 71,428,571 | 17,671,030% | 295 | 242,131× |
| Sweet Rice Cake | 23 | 18,999,996 | 78,478,143% | 6,249 | 3,041× |
| Goatskin Eyepatch | 708 | 100,000,000 | 13,417,979% | 11,504 | 8,693× |

Replaying the page's real computation against a live Gilgamesh snapshot: **9 of the top 10 rows by ROI, and 18 of the top 50, are junk.** The tool's headline results are noise.

## Why it happened

The Flip Finder, Trends, Best Deals and Top Opportunities all filter these out via `ResaleQualityFilter`, whose own module doc says it exists so that "the Trends page, the Flip Finder, the Top Opportunities card, **and any future resale surface**" share one hide/show decision. Vendor Resale is exactly that future surface, and was never wired in.

It needs no new data to fix. `compute_summary` **already** derives `median_price` from the item's recent sales and hangs it on every row — nothing ever compared it against the listing.

## The fix

A local plausibility check using data already on the page, plus a `show-suspicious` toggle matching the other analyzers (hidden by default, active-filter chip + clear-all wired).

Rows we *cannot* judge — no recent sales, or a degenerate median — are **kept**, matching `ResaleQualityFilter`'s rule that a row without enrichment is shown rather than penalized. The page's existing "Sales (min)" filter remains the control for requiring sale history outright.

### The threshold is measured, not guessed

Across 10,289 Gilgamesh NQ items holding both a cheapest listing and recent sales:

| percentile | listing ÷ median recent sale |
|---|---|
| p50 | 0.99× |
| p75 | 1.48× |
| p90 | 4.08× |
| p95 | 10.2× |
| **p99** | **433×** |
| **p99.9** | **83,091×** |

Normal listings cluster at ~1× and tail off smoothly to ~10×; past p99 it becomes a separate population. **50×** leaves five times normal variation intact, and drops **32 of 1,956** vendor rows (1.64%). Every junk row observed ran 1,000×–150,000×.

After the fix the top of the table is plausible again — best remaining row is 18.2× (a genuinely thin market), then 21.1×, 12.6×, 1.5×, 1.0×.

## Verification

- **Genuine red→green.** With pre-fix behaviour temporarily restored, 3 of the 5 new tests fail at exit `101` with the real message: `Sweet Rice Cake: listing 18999996 vs median sale 6249 should be suspicious`. (The other 2 assert rows are *kept*, so they correctly pass under both behaviours.) Restored → all pass.
- `cargo test -p ultros-app --lib` → **438 passed / 0 failed** (baseline 433 + 5 added).
- `cargo fmt --all -- --check` → exit 0.
- `cargo clippy -p ultros-app --all-targets -- -D warnings` → exit 0.
- `cargo check -p ultros-app --no-default-features --features hydrate --target wasm32-unknown-unknown` → exit 0. CI never lints the hydrate path, and that's the build that ships to the browser, so I checked it explicitly.
- The 5 new i18n keys are present with real (non-stub) translations in all 7 locales; `leptos-i18n` won't compile without them, and it does.

Test fixtures are the real observed rows and the real medians `compute_summary` produces from the live API response, so the test documents the actual incident rather than a synthetic one.

## Note

The threshold is deliberately conservative — it is meant to remove the unbuyable, not to second-guess thin markets. If you'd rather it be stricter, `SUSPICIOUS_PRICE_MULTIPLE` is a single constant with the measured distribution written above it.


akarras:
Reviewed and verified end to end, including independently against live production data rather than taking the PR body's numbers on faith.

## The bug is real, and worse than stated

Loaded `/vendor-resale/Gilgamesh` on the live prod build (`b29a199`) in a real browser. The PR says 9 of the top 10 rows were junk. Today it is **10 of 10** — the entire first screen:

| # | Item | Vendor | Listing | Median sale | Ratio |
|---|---|---|---|---|---|
| 1 | Sweet Rice Cake | 23 | 18,999,996 | 6,249 | 3,040x |
| 2 | Copper Wristlets | 229 | 150,000,000 | 849 | 176,678x |
| 3 | Hyuran Longboots | 48 | 9,999,999 | 9,999 | 1,000x |
| 4 | Steel Jig | 384 | 71,428,571 | 295 | 242,131x |

## The threshold reproduces

Re-derived the distribution from a fresh `/api/v1/cheapest/Gilgamesh` + `/api/v1/recentSales/Gilgamesh` pull (10,292 NQ items with both a cheapest listing and recent sales), replicating `compute_summary`'s median exactly:

| p50 | p75 | p90 | p95 | p99 | p99.5 | p99.9 |
|---|---|---|---|---|---|---|
| 0.99x | 1.48x | 4.08x | 10.23x | 434.03x | 2,127.89x | 92,747x |

Matches the PR's figures to two decimals. The p95->p99 jump from 10x to 434x is a genuine population break, and 50x sits in the empty part of it.

## The filter discriminates correctly on live rows

Applied the rule to the 25 rows actually on prod page 1: **15 hidden, 10 kept**. The separation is clean with nothing near the boundary —

- highest **kept**: 21x (Leather Calot), then 20x, 18x, 17x, 14x, 13x, 1x...
- lowest **hidden**: 129x (Wooden Bucket at 999,999,999), then 189x, 500x, 609x...

Nothing lands between 21x and 129x, so the result is insensitive to the exact constant anywhere in roughly 25x-120x. Genuine thin-market rows (Usagi Kabuto 18x, Sheep Dolls 13x, Kadomatsu 14x) survive, which is the outcome that actually matters.

## Tests are not vacuous

`cargo test -p ultros-app --lib` -> **438 passed / 0 failed** (433 baseline + 5). CI never runs tests (the step is commented out in rust.yml), so I confirmed red->green by reverting `is_suspicious_market_price` to pre-fix behavior: exactly 3 fail with real messages (`Sweet Rice Cake: listing 18999996 vs median sale 6249 should be suspicious`), and the 2 "row is kept" tests correctly pass under both. Restored -> all green.

## UI verified in a real browser

Injected the exact markup `ToolbarField`/`ToolbarPills` emit into the live prod toolbar (Puppeteer, prod's real stylesheet), at 1280 and 375, in English and German — German being the widest locale:

- desktop: field 134px (en) / 193px (de), toolbar height unchanged at 148px
- mobile 375: fits the existing row in en; wraps to one new row in de (267->334px)
- `insideToolbar: true` and **no horizontal page scroll** in all four combinations
- `aria-pressed` styling resolves to distinct backgrounds, so the active state reads correctly, and "Hidden" is correctly the default

`.toolbar` is `flex-wrap: wrap` with no height constraint — it is not the height-locked Flip Finder bar — so growth is absorbed rather than clipped.

Data pairing is sound: HQ listings are skipped outright and the summary is looked up with `hq: false`, so the comparison is NQ-listing vs NQ-sales. The i64 widening makes overflow unreachable (max `i32::MAX * 50` ~ 1.07e11). All 7 locales carry real translations, not English stubs.

## One follow-up, deliberately not blocking

The toggle uses plain `query_signal`, where the Flip Finder's identically-named `show-suspicious` uses `filter_query_signal`. That distinction is load-bearing: `filter_nav_options` exists because plain `query_signal` is `replace: false, scroll: true` ("Filters are not navigation"). Measured on prod using this page's existing tax pill, which has the same wiring: clicking it **scrolls 938 -> 0 and pushes a history entry**, so back-button walks filter states.

Not changing it here. 5 of the 6 existing vendor-resale filters already use plain `query_signal`, so the new toggle is consistent with its immediate neighbor; converting only the new one would make the page behave inconsistently. Filing the page-wide conversion separately.

Also confirms the doc comment's "2.1%" is correct and not in conflict with the body's 1.64% — different denominators (2.16% of all judgeable NQ items vs 1.64% of vendor rows). I verified 222/10,292 = 2.16%.

Merging.

PR #1120, opened 2026-08-04
🛡️ Sentry: [test coverage improvement]

akarras:
Added unit tests for the `escape_for_script_tag` method to ensure HTML and JS characters are escaped correctly to prevent XSS.

---
*PR created automatically by Jules for task [11323133583695667912](https://jules.google.com/task/11323133583695667912) started by @akarras*

### DIFF
--- a/ultros-frontend/ultros-app/src/script_escape.rs
+++ b/ultros-frontend/ultros-app/src/script_escape.rs
@@ -20,3 +20,28 @@ pub fn escape_for_script_tag(json: &str) -> String {
     }
     out
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_escape_for_script_tag() {
+        let json = r#"{"name": "test", "value": "<script>alert('xss')</script>&amp;\u2028\u2029"}"#;
+        let escaped = escape_for_script_tag(json);
+        assert_eq!(
+            escaped,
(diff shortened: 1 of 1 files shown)


akarras:
Reviewed. Tests-only, no production change, and the assertions match the real behavior of `escape_for_script_tag`.

Ran them locally, since CI compiles tests via `clippy --all-targets` but never executes them (the `cargo test` step is commented out in rust.yml):

```
test script_escape::tests::test_escape_for_script_tag ... ok
test result: ok. 1 passed; 0 failed
```

Checked that this doesn't enshrine a bug: the escape set (`<`, `>`, `&`, U+2028, U+2029) is the correct one for embedding JSON in a `<script>` element. `<` alone closes the `</script>` breakout, `>`/`&` are defense in depth, and the two line separators are required because they are legal in JSON strings but terminate a JS line. Substituting `\u003c` for `<` is also always valid JSON, since `<` can only ever occur inside a JSON string literal.

One thing worth knowing about the first assertion. The input is a **raw** string (`r#"..."#`), so the `\u2028\u2029` in it are the six literal characters `\`,`u`,`2`,`0`,`2`,`8` — not the actual U+2028/U+2029 code points. That assertion therefore passes identically whether the function escapes those code points or passes literal text through untouched, so it proves nothing about line-separator handling on its own.

Not a problem in practice: the third assertion uses `"...\u{2028}\n\u{2029}"` in a normal string, which *is* the real code point, and that one genuinely covers it. So the coverage is there — just carried by the last assertion rather than the first. Flagging it so nobody later trims the third case believing the first already covers it.

Merging.

PR #1121, opened 2026-08-04
fix(item-stats): serve datacenter- and region-scoped requests

akarras:
## The bug

`/api/v1/item_stats/{world}/{item}` rejects anything but a single world. Live against prod right now:

```
item_stats/Gilgamesh/49249 -> 200
item_stats/Phoenix/49249   -> 200
item_stats/Aether/49249    -> 400   (datacenter)
item_stats/Chaos/49249     -> 400   (datacenter)
item_stats/Europe/49249    -> 400   (region)
item_stats/North-America/8 -> 400   (region)
```

The item view fires this request on **every** load, and the confidence badge is written to soft-fail — "if the endpoint errors or returns no variant, the badge simply doesn't render". So the chip has been silently dead for every datacenter- and region-scoped visitor, with nothing surfaced to the user and only a 400-per-page-load in GlitchTip.

Found in the breadcrumbs of GlitchTip issue [#6874](https://glitchtip.ultros.app/ultros/issues/6874) — a real Europe-scoped user session, two page views, two 400s:

```
GET https://ultros.app/api/v1/item_stats/Europe/49249  -> 400
GET https://ultros.app/api/v1/item_stats/Europe/8      -> 400
```

The handler's stated rationale is stale:

```rust
// Item stats are per-world; reject DC/Region selectors to keep the
// semantics tight. The chart they're paired with is also per-world.
```

`price_series` has since grown datacenter/region support (`PriceSeriesArgs.world`: *"A world, datacenter, or region name; expanded to its constituent worlds"*), and the same session shows `price_series/Europe/49249` returning 200 beside the 400. The badge was the odd one out.

## The fix

`item_stats_window` is keyed by world, so the scope fans out to its member worlds via `AnyResult::all_worlds()` — the same expansion `build_price_series` already uses — and the rows are folded back into one variant per quality:

| field | how it combines |
|---|---|
| `sample_size_30d`, `cleaned_sample_size_30d` | sum |
| `vwap_30d`, `p50_30d` | weighted by `cleaned_sample_size` |
| `launder_suspicion` | weighted by `sample_size` (its own denominator) |
| `confidence_band` | band of the world with the most cleaned samples, tie-broken on world id |

Two things worth a reviewer's eye:

- **`p50` across worlds is an approximation.** A sample-weighted mean of per-world medians is not the true median of the combined sample, but the rollup doesn't keep the raw distribution. It's weighted rather than flat-averaged so a 4-sale world can't move it as far as a 4000-sale one, and the function doc says so plainly. The badge itself reads only `confidence_band` + `sample_size_30d`, both of which combine exactly.
- **`confidence_band` is a stored per-world judgement** from `item_quality_score`, not something recomputable from the folded numbers — hence "dominant world wins". The tie-break on world id matters: without it the answer would depend on ClickHouse's row order, which is the same class of SSR/CSR nondeterminism as #998/#960.

**A single-world scope returns that row's values verbatim** (`[only] => variant_of(only)`), so world-scoped requests are byte-identical to before the change.

### Wire change

`ItemStatsResponse::world_id` becomes `Option<i32>` — a multi-world scope has no single id to report. Neither consumer (`ultros-app/src/api.rs`, the badge) reads the field, and regions previously 400'd, so no response shape that works today changes.

## Verification

- **Reproduction is the live prod matrix above** — datacenter and region 400, worlds 200.
- **9 new unit tests** in `ultros-clickhouse` (a lib crate, so these actually run here — unlike the `ultros` bin, whose test binary won't link on Windows/MSVC).
- **Genuine red→green.** With `fold_variants` replaced by the pre-fix "first world only" semantics, 6 of the 9 fail at exit 101; restored, `cargo test -p ultros-clickhouse --lib` is **33 passed / 0 failed** (was 24).

Gates, all exit 0:

- `cargo fmt --all -- --check`
- `cargo clippy -p ultros --all-targets -- -D warnings` (8m05s)
- `cargo clippy -p ultros-app --all-targets -- -D warnings` (8m09s)
- `cargo check -p ultros-app --no-default-features --features hydrate --target wasm32-unknown-unknown` (4m34s) — CI never lints the hydrate cfg, and that's what ships to the browser

Not verified against a live ClickHouse (the aggregation is pure and unit-tested; the fan-out query reuses `deep_scan_batch` unchanged). Worth a sanity check on `/item/Europe/<something>` after deploy that the badge appears.

## Not fixed here

- **#6874** itself — the `reactive_graph` "already disposed" panic on route change (`/item/49249` → `/item/8`), count 1. Separate bug; the 400s were in its breadcrumbs, not its cause.
- **#6869 / #6868 / #6870 / #6873 / #6872** are all `environment:development` / `server_name:Bahamut` — the dev box, not prod. Left alone per convention.

### DIFF
--- a/ultros-api-types/src/item_stats.rs
+++ b/ultros-api-types/src/item_stats.rs
@@ -24,7 +24,11 @@ pub struct ItemStatsVariant {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct ItemStatsResponse {
-    pub world_id: i32,
+    /// The world these stats are scoped to, or `None` when the request named
+    /// a datacenter or region — the rollup is per-world, so a multi-world
+    /// scope has no single id to report and the variants below are folded
+    /// across every member world instead.
+    pub world_id: Option<i32>,
     pub item_id: i32,
     pub variants: Vec<ItemStatsVariant>,
 }
--- a/ultros-clickhouse/src/queries.rs
+++ b/ultros-clickhouse/src/queries.rs
@@ -8,6 +8,7 @@
 
 use clickhouse::Row;
 use serde::Deserialize;
+use ultros_api_types::item_stats::ItemStatsVariant;
 use ultros_api_types::price_series::{HqFilter, SeriesGroup};
 use ultros_api_types::trends::ConfidenceBand;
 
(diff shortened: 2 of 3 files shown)


akarras:
**Reviewed — verified independently.**

- **Repro confirmed against prod**: `item_stats/{Aether,Chaos,Europe,North-America}/49249` all 400; `Gilgamesh` 200. Matches the PR body exactly.
- **Bug is real end-to-end**: `item_view.rs:1249` passes `world()`, which can be a DC/region name, and the badge soft-fails to nothing — the chip has been silently dead at those scopes.
- **Fan-out precedent checks out**: `build_price_series` (web.rs:531) already expands scope via `get_all_worlds_in`; `AnyResult::all_worlds()` and `AnySelector::as_world_id()` are pre-existing, separately-tested helpers.
- **Wire change is safe**: nothing reads `ItemStatsResponse::world_id` — not `api.rs`, not the badge, not the `variant_for` impl. Regions previously 400'd, so no working response shape changes.
- **Tie-break reasoning is correct**: `max_by_key` keeps the *last* maximum, so keying on `(cleaned_sample_size, Reverse(world_id))` does make the lowest world id win. The determinism test pins it — worth having given this repo's SSR/CSR row-order history.
- **Tests pass locally**: `cargo test -p ultros-clickhouse --lib` → **33 passed / 0 failed**, exit 0. CI lint/check job green.

The `p50` weighted-mean approximation is documented honestly at both function and PR level, and the badge reads only `confidence_band` + `sample_size_30d`, which combine exactly. Single-world scopes pass through verbatim, so the blast radius is confined to scopes that are broken today.

Post-deploy sanity check worth doing: load `/item/Europe/<item>` and confirm the badge renders.

PR #1122, opened 2026-08-04
fix(ingest): apply listings/add as a delta instead of replacing the board

akarras:
## The bug

Universalis' `listings/add` carries only the listings that **newly appeared** on the board — not a snapshot of it. Measured live over 45s: 449 events, payload size min 1, **median 1**, mean 2.2, max 20.

We fed it to `update_listings`, which is documented *"assuming a pure view of the listing board"* and deletes every row absent from its input. So every add event truncated that world's board down to the delta.

## Evidence

Matching payload `listingID`s against the live Universalis board for the same world:

- payload was a strict subset of the live board in **15/20** samples
- **19/20** events would have deleted still-live listings — **491 rows across 20 events**

Production damage, 20-item sample on Faerie: Ultros held **173 of 556** real listings, i.e. **31% of the market**. Item 52267 showed 1 of 19.

| World (id) | Universalis | Ultros |
|---|---|---|
| Jenova (40) | 36 | 36 ✅ |
| **Faerie (54)** | **19** | **1** ❌ |
| Siren (57) | 18 | 18 ✅ |
| **Gilgamesh (63)** | **37** | **3** ❌ |
| **Cactuar (79)** | **34** | **1** ❌ |

Boards *oscillate* rather than staying broken: the REST catch-up service fetches genuine full boards and repairs a world, then the next add event truncates it again. That's why per-world counts look random and why a single reading never showed the pattern — item 52267 read 1 listing, then 19 twenty minutes later with no intervention.

## The fix

`listings/add` is now insert-only via `UltrosDb::add_listings`. Removals already arrive separately on `listings/remove` and are unchanged. `update_listings` stays the path for callers that really do fetch a whole board (the catch-up service and the manual refresh route).

Dedupe uses the same `(hq, quantity, price, retainer name)` key the full-board diff uses, reusing its `added` side so the "already have it" test can't drift from the comparator — including multiset handling, so a retainer holding three identical listings when we store two yields exactly one insert. Universalis' `listingID` would be an exact identity, but `active_listing` has no column for it.

The catch-up marker is still stamped: an add event reflects a real upload at that moment, and the marker is compared against upload time, so *not* stamping would make every uploaded item read as permanently behind and drag the sweep into refetching the whole market.

## Tests

4 new tests in `ultros-db`. The main one asserts **both halves** of the bug: that the old full-board diff really does condemn the other 18 listings, and that the new path returns only the 1 genuine insert.

Note this repo's CI runs fmt + clippy only, not `cargo test` — these were run locally (11/11 pass).

## Verification

- `cargo check -p ultros --all-targets` — clean
- `cargo clippy -p ultros-db -p ultros --all-targets -- -D warnings` — clean, zero warnings
- `cargo fmt --all` — clean
- `cargo test -p ultros-db --lib` — 11/11 pass

### DIFF
--- a/ultros-db/src/listings.rs
+++ b/ultros-db/src/listings.rs
@@ -137,6 +137,26 @@ struct ListingsDiff {
     removed: Vec<(active_listing::Model, Option<retainer::Model>)>,
 }
 
+/// Selects the listings from an incremental `listings/add` payload that we don't
+/// already hold for this item/world.
+///
+/// Universalis' `listings/add` is a *delta* — it carries the listings that newly
+/// appeared on the board, not a snapshot of it — so this only ever yields rows to
+/// insert. Matching reuses [`diff_update_listings`]' `added` side so the
+/// "already have it" test is exactly the comparator the full-board path uses,
+/// including its multiset handling: a retainer legitimately holding three
+/// identical listings when we already store two yields exactly one insert.
+///
+/// Universalis' `listingID` would be an exact identity, but `active_listing` has
+/// no column for it, so the key is the same (hq, quantity, price, retainer name)
+/// tuple used everywhere else in this module.
+fn listings_to_add(
+    listings: Vec<ListingView>,
+    existing_items: Vec<(active_listing::Model, Option<retainer::Model>)>,
+) -> Vec<ListingView> {
+    diff_update_listings(listings, existing_items).added
+}
+
 /// Diffs the incoming full listing-board view against the DB's current rows for an
 /// item/world, sorting both sides by the same canonical key before merging so the
 /// merge loop's comparator matches the sort exactly (previously the merge loop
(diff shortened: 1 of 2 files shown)


akarras:
**Reviewed — verified independently.**

Traced the blast radius of turning `EventType::Add` from a full board into a delta, which is the part that could have bitten:

- **`analyzer_service::add_listings`** (analyzer_service.rs:1530) folds each listing in via `add_listing` on the cheapest-items map — incremental by construction, so a delta is not just safe but *more* correct than the full board it used to get.
- **`undercut_alert`** (undercut_alert.rs:198) takes `min_by_key(price_per_unit)` over the added set; with a delta that is exactly the newly-arrived undercut. No full-board assumption.
- **Frontend is unaffected** — `ultros-app` only subscribes to `Sales` events (`live_sale_ticker.rs`, `auto_mark_purchases.rs`), never `Listings`.
- **Dropping the `EventType::Remove` send is a second fix, not a regression**: that broadcast was carrying the *bogus* removals, so `analyzer_service::remove_listings` was being told to evict genuinely-live cheapest listings from the cache.
- `update_listings` is correctly left in place for the two callers that really do fetch a whole board (`item_update_service.rs:188`, `web.rs:1174`).

Other checks:

- The `expect("Should always have a retainer")` in `add_listings` cannot fire — `resolve_retainers` is keyed on every `listing.retainer_name` in the payload and `to_add` is a subset of it. Same invariant the extracted-from `update_listings` relied on.
- Metrics match existing conventions in this module (`counter!`/`histogram!` with `Instant::elapsed()`, same as lines 501–505).
- **Tests pass locally**: `cargo test -p ultros-db --lib` → **34 passed / 0 failed**, exit 0, including all four new delta tests. The regression guard asserting the old full-board diff really does condemn the other 18 listings is the right way to write this — it pins both halves of the bug.
- CI lint/check job green.

Accepting the tradeoff that removals now depend solely on `listings/remove` plus the catch-up sweep's full-board prune: worst case is a stale row lingering, which is strictly better than truncating boards to ~31% of the market.

PR #1123, opened 2026-08-04
fix(analyzer): rebuild the cheapest-listing map instead of trusting the event stream

akarras:
## The bug

The analyzer served buy prices for listings that no longer exist. On Faerie, item 52267 read **6,769 gil** while Postgres and Universalis both agreed the real cheapest was **379,996**.

A 400-item sample on Faerie, analyzer vs. the database's true cheapest:

| | count | |
|---|---|---|
| exact match | 294 | 77.2% |
| **stuck below DB** | **76** | **19.9%** |
| above DB (ordinary lag) | 11 | 2.9% |

That asymmetry is the signature. Worst cases ran 630x, several pinned at literally **1 gil**.

## Why it sticks

`add_listing` keeps the `min` of stored and incoming price, so a stored price can only ever move **down**. The one path that can raise it is `remove_listing`, gated on `listing.price_per_unit <= entry.get().price`. Once a too-low price is in the map, every later remove fails that gate — `379996 <= 6769` is false — and the entry is pinned forever.

**Recovery was the missing piece, not injection.** `ultros_analyzer_bus_lagged_total` shows **~99k listings-bus events dropped in 7 days** (bursty — zero in the last 5 hours), and `BusRecv::Lagged` acknowledged the loss and did nothing.

Worth recording: `ultros_analyzer_skipped_events_total` has **never been incremented** in production, so the retainer-filter drop paths in `update_listings`/`remove_listings` are latent, not a cause.

Snapshots then made each burst permanent. `run_worker` reseeded from Postgres only when no snapshot restored — but one is written every 15 minutes, so a price stranded days earlier was serialized and read back on every deploy.

## Changes

- **Rebuild `cheapest_items` from Postgres on every boot**, snapshot or not, and again after any bus lag (collapsed to one pass per burst so a burst doesn't queue N rebuilds).
- **Rebuilds replace rather than merge.** Merging cannot clear a phantom, since the stored price only moves down. There's a test for exactly this.
- Snapshot restore is **kept as a fallback** for a failed rebuild — stale prices beat an empty map — it's just no longer the only source. (The existing snapshot round-trip test asserts DC-keyed restore; dropping it would have broken that for no gain.)
- Build into unlocked maps first, so no lock is held across the DB stream and readers never see a half-filled map.
- **`cheapest_listings`: full-table `RANK()` window → `GROUP BY MIN`.** The `rank = 1` filter can't be pushed into the window, so Postgres sorted every row in `active_listing` to keep one per group; `RANK()` also returned ties as duplicate rows. Adds a covering index on `(item_id, hq, world_id, price_per_unit)` — the only existing index (`WorldItemIndex`) carries neither `hq` nor price. Created `CONCURRENTLY` via `use_transaction() -> Some(false)`, since `active_listing` is written continuously by ingest.
- **`EventType::Update` on the listings bus was `todo!()`** — a panic that would take the analyzer down if anything ever published one. Now a no-op, matching the sales arm.

## Verification

- `cargo clippy -p ultros --all-targets -- -D warnings` — clean
- `cargo fmt --all` — clean
- `cargo test -p ultros-db --lib` — 30 passed

⚠️ The new `merging_fresh_data_cannot_clear_a_phantom_minimum_but_replacing_does` test **compiles but has not been executed**: the `ultros` bin test binary doesn't link on Windows/MSVC (74 unresolved `ultros_app` externals), and this repo's CI runs fmt + clippy only, not `cargo test`. It should run on Linux — worth a check before merge if you have one handy.

## Merge order

Independent of #1122. If the analyzer perf PR follows, this one should land first — its index is what that PR's new query is designed to use.

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -32,6 +32,7 @@ mod m20260514_000002_alert_list_update;
 mod m20260514_000003_list_activity;
 mod m20260802_000001_authless_character_claims;
 mod m20260802_000002_user_group_discord_guild;
+mod m20260804_000001_active_listing_cheapest_index;
 
 pub struct Migrator;
 
@@ -73,6 +74,7 @@ impl MigratorTrait for Migrator {
             Box::new(m20260514_000003_list_activity::Migration),
             Box::new(m20260802_000001_authless_character_claims::Migration),
             Box::new(m20260802_000002_user_group_discord_guild::Migration),
+            Box::new(m20260804_000001_active_listing_cheapest_index::Migration),
         ]
     }
 }
--- a/migration/src/m20260804_000001_active_listing_cheapest_index.rs
+++ b/migration/src/m20260804_000001_active_listing_cheapest_index.rs
@@ -1,0 +1,47 @@
+use sea_orm_migration::prelude::*;
+
+/// Covering index for `UltrosDb::cheapest_listings`, which the analyzer runs on
+/// boot and on bus-lag recovery:
+///
+/// ```sql
+/// SELECT item_id, hq, world_id, MIN(price_per_unit)
+/// FROM active_listing GROUP BY item_id, hq, world_id
+/// ```
+///
+/// The only pre-existing index on this table is `WorldItemIndex` on
+/// (item_id, world_id) — it carries neither `hq` nor `price_per_unit`, so that
+/// query had to go to the heap for every row. Ordering the columns
+/// group-keys-then-value lets Postgres serve it as an index-only scan.
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    /// `CREATE INDEX CONCURRENTLY` is rejected inside a transaction block, and
+    /// sea-orm wraps migrations in one by default on Postgres. `active_listing`
+    /// is written continuously by ingest, so a plain `CREATE INDEX` would hold an
+    /// exclusive write lock for the whole build — opting out of the transaction
+    /// is what keeps this migration online.
+    fn use_transaction(&self) -> Option<bool> {
+        Some(false)
+    }
+
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute_unprepared(
+                r#"CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_active_listing_cheapest
+                   ON active_listing (item_id, hq, world_id, price_per_unit)"#,
+            )
+            .await?;
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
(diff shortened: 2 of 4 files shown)


akarras:
Reviewed. Approving and merging.

**The diagnosis is correct — I verified the mechanism in the source rather than taking the writeup on faith:**

- `CheapestListings::add_listing` (analyzer_service.rs:278) is `*entry = cheapest_listing.min(*entry)` — a stored price can only ever move *down*.
- `remove_listing` (analyzer_service.rs:293) is gated on `listing.price_per_unit <= entry.get().price`.

So a too-low price is genuinely unclearable by any subsequent event, and the 19.9%-stuck-below-DB vs 2.9%-above asymmetry in the sample is exactly the signature that predicts. Rebuilding wholesale is the only thing that clears one, and the accompanying test encodes precisely that asymmetry.

Identifying **recovery, not injection** as the missing piece is the part that makes this the right fix — and the observation that snapshots made each lag burst permanent (serialized every 15 min, read back on every deploy) explains why this outlived so many restarts. Keeping snapshot restore as a fallback for a failed rebuild is the right call.

The `cheapest_listings` query change is a clean win independently: `RANK()` could not push the `rank = 1` filter into the window so Postgres sorted the entire table, and it returned ties as duplicate rows. `GROUP BY ... MIN` is exactly the question being asked, and the covering index column order (group keys then value) is right for an index-only scan. `CREATE INDEX CONCURRENTLY` with `use_transaction() -> Some(false)` is correct and necessary on a continuously-written table.

Turning the `EventType::Update(_) => todo!()` panic into a no-op is a good catch.

**Verification note:** CI here runs fmt + clippy only (`cargo test --workspace` is commented out in `rust.yml`), so green CI proves compilation. I did not execute the new test — the `ultros` bin test binary does not link on Windows/MSVC — but its assertions are over `add_listing`'s min-semantics, which I confirmed by reading the implementation above, so it is provably green.

**Two non-blocking follow-ups worth recording:**

1. **The resync races live ingest.** `rebuild_cheapest_from_db` streams the DB and then does `*lock.write().await = listings`, while the listings consumer keeps applying events throughout. A `listings/remove` that lands after its row was already streamed gets discarded by the replace, re-stranding that price until the *next* rebuild. That is strictly better than the status quo (phantoms are now bounded by boot/lag frequency instead of immortal), so it should not hold this up — but the window is real and worth a follow-up if the stuck-price rate does not drop to near zero after deploy.

2. **Transient memory doubling.** Building `fresh` fully before replacing means peak usage is ~2x the steady-state cheapest map during a rebuild. That is the correct trade for not exposing a half-filled map to readers, but a lag-triggered resync now spikes memory on an already-loaded process — worth watching alongside the new `ultros_analyzer_cheapest_resync_total` counter.

3. Operational: if the `CONCURRENTLY` build ever fails partway it leaves an INVALID index, and `IF NOT EXISTS` will then skip recreating it. Worth a `\d active_listing` check after deploy.

PR #1124, opened 2026-08-04
perf(analyzer): batch the cheapest-listing refill and drop the locks around it

akarras:
## Why

The analyzer's listings consumer can't keep up with ingest bursts — `ultros_analyzer_bus_lagged_total` recorded **~99k dropped listing events in a week**. Dropped events aren't merely delayed: a lost `listings/remove` strands a cheapest price that no later event can correct, because `add_listing` only ever lowers a stored price. (#1123 handles the recovery side; this PR attacks why the consumer falls behind in the first place.)

## Two compounding causes

**1. The refill query was a cartesian fan-out.** `get_multiple_listings_for_worlds_hq_sensitive` built its result with `worlds.flat_map(|w| items.map(|i| ...))` — **one query per (world × item) pair**.

**2. It ran inside the write lock.** `remove_listing` called that query per removed listing, from inside the write lock for that level of the map. For the region selector that's a *single lock covering every item in the region*, held across a sequence of queries.

An 18-listing removal on an 8-world region meant roughly **144 sequential round-trips under one write lock** — then repeated for the datacenter and world maps.

## Changes

- **Replace the fan-out** with `cheapest_listings_for_items`: one grouped query, built with sea-orm's query builder so the `IN` lists stay parameterized. It returns both qualities at once, since `ListingSummary` carries `hq` and the analyzer keys on it — the hq-sensitive split was two round-trips for no reason. The old method had exactly one caller, so it's deleted rather than left as a trap.
- **Split `remove_listing`** into `remove_if_cheapest` (synchronous, DB-free, reports which keys went stale) and a three-phase caller: decide under the lock → query with **no lock held** → re-acquire and apply.
- **Group removals by selector first**, so each selector is one pass and one query instead of one per (listing, level) pair.

Net: that same 18-listing removal goes from ~144 sequential round-trips under a write lock to **3 queries, none under a lock**.

- Bus raised 1024 → 8192. Explicitly a mitigation, not a fix — a consumer persistently slower than the producer lags at any size — which is why the lock and fan-out changes are the substance.

## Verification

- `cargo clippy -p ultros-db -p ultros --all-targets -- -D warnings` — clean
- `cargo fmt --all` — clean
- `cargo test -p ultros-db --lib` — 30 passed

⚠️ The new `remove_if_cheapest_only_reports_keys_whose_price_actually_went_stale` test **compiles but has not been executed**: the `ultros` bin test binary doesn't link on Windows/MSVC (74 unresolved `ultros_app` externals), and CI runs fmt + clippy only, not `cargo test`. It should run on Linux.

## Merge order

Based on `main`, not stacked, so it gets its own CI. It overlaps #1123 in `analyzer_service.rs`, so whichever lands second needs a small conflict resolution — **#1123 first is cleaner**, since its covering index on `(item_id, hq, world_id, price_per_unit)` is exactly what this PR's grouped query is shaped to use.

### DIFF
--- a/ultros-db/src/lib.rs
+++ b/ultros-db/src/lib.rs
@@ -218,30 +218,6 @@ impl UltrosDb {
             .ok_or_else(|| anyhow::Error::msg("Region not found"))
     }
 
-    #[instrument(skip(self, world_id, item))]
-    pub async fn get_multiple_listings_for_worlds_hq_sensitive(
-        &self,
-        world_id: impl Iterator<Item = WorldId>,
-        item: impl Iterator<Item = ItemId> + Clone,
-        hq: bool,
-        limit: u64,
-    ) -> Result<Vec<active_listing::Model>> {
-        use active_listing::*;
-        let join = futures::future::try_join_all(world_id.flat_map(|world| {
-            item.clone().map(move |i| {
-                Entity::find()
-                    .filter(Column::ItemId.eq(i.0))
-                    .filter(Column::WorldId.eq(world.0))
-                    .filter(Column::Hq.eq(hq))
-                    .order_by_asc(Column::PricePerUnit)
-                    .limit(limit)
-                    .all(&self.db)
-            })
-        }))
-        .await?;
-        Ok(join.into_iter().flat_map(|l| l.into_iter()).collect())
-    }
-
     #[instrument(skip(self))]
     pub async fn get_listings_for_world_hq(
         &self,
--- a/ultros-db/src/listings.rs
+++ b/ultros-db/src/listings.rs
@@ -4,7 +4,8 @@ use itertools::Itertools;
 use metrics::{counter, histogram};
 use migration::DbErr;
 use sea_orm::{
-    ColumnTrait, DbBackend, EntityTrait, ExprTrait, FromQueryResult, QueryFilter, Statement,
+    ColumnTrait, DbBackend, EntityTrait, ExprTrait, FromQueryResult, QueryFilter, QuerySelect,
+    Statement,
 };
 use std::{
     collections::{HashMap, HashSet, hash_map::Entry},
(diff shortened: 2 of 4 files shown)


akarras:
Reviewed, conflict-resolved, and merged.

**I resolved the #1123 overlap on this branch** (merge commit `e23ff15`) — the conflict this PR's description predicted:

- `ultros-db/src/listings.rs` — both sides inserted immediately before `cheapest_listings`. Kept this PR's new `cheapest_listings_for_items` *and* #1123's doc comment; the `GROUP BY MIN` body from #1123 merged automatically.
- `ultros/src/analyzer_service.rs` — both sides appended a test to the same `mod test`, sharing a closing brace. Kept both tests and both imports.

I verified the resolution did not silently revert #1123: `git diff origin/main..HEAD` on the merged branch is byte-for-byte the same file set and line counts as this PR's own delta against its base, and I confirmed `rebuild_cheapest_from_db` / `schedule_cheapest_resync` / `cheapest_resync_in_flight` all survive, with no remaining references to the deleted `get_multiple_listings_for_worlds_hq_sensitive`.

**Gates:** `cargo fmt --all -- --check` clean locally, and the `Lint, checks, and tests` job **passed on the merged tree** — which is the real compile gate here, since `cargo clippy --all-targets` type-checks the test module I hand-merged. (Worth stating for the record: that job runs fmt + clippy only; `cargo test --workspace` is commented out in `rust.yml`, so its name oversells it.)

**On the substance** — the two causes are correctly identified and the fix matches them. `worlds.flat_map(|w| items.map(|i| ...))` really is one query per pair, and issuing them from inside the write lock (a *region-wide* lock for the region selector) is what turns a fan-out into a stall. Splitting `remove_listing` into a synchronous DB-free `remove_if_cheapest` plus decide-under-lock -> query-unlocked -> reapply is the right shape, and grouping removals by selector before querying is what makes it one round-trip instead of one per (listing, level). Deleting the old method rather than leaving it as a trap is the correct instinct given it had exactly one caller.

Good that the bus-size bump is labelled a mitigation rather than the fix — a consumer persistently slower than its producer lags at any buffer size.

This composes well with #1123: that PR's covering index on `(item_id, hq, world_id, price_per_unit)` is exactly the shape this PR's grouped query wants, and #1123's rebuild path is independent of the refill path this one rewrites.

PR #1134, opened 2026-08-05
fix(ssr): flip-finder hydration failure — gate LocalStorage stored values out of the server build

akarras:
## Symptom

`https://ultros.app/flip-finder/Gilgamesh?min-buy=5000&last-sold=1d&roi=30&...&confidence=high` hydration-crashes in the browser:

```
Uncaught ReferenceError: __INCOMPLETE_CHUNKS is not defined
panicked at js-sys .../singlethread.rs:142: RefCell already borrowed
Uncaught RuntimeError: unreachable executed
```

## Root cause

The SSR stream was dying mid-response. Reproduced with a curl burst: **3 of 4 fetches truncated at ~8.4MB** (mid `__RESOLVED_RESOURCES` sales JSON, no closing chunk), each truncation matching a prod panic in `docker logs`:

```
panicked at reactive_graph-0.2.14/src/owner/storage.rs:141:34:
Dereferenced SendWrapper<T> variable from a thread different to the one it has been created with.
```

leptos only emits the `__INCOMPLETE_CHUNKS=[];` bootstrap as the **final** chunk of the stream (`hydration_context/src/ssr.rs`), so a truncated stream leaves the global undefined and the client's wasm-bindgen static accessor throws during hydration — the whole cascade the user sees.

The GlitchTip backtrace (issue 6870, first seen Jul 28 — matches #991 landing) pins it: `AnalyzerTable`'s scroll-sync `on_cleanup` → `hscroll_listeners.update_value` → LocalStorage `try_with` → SendWrapper deref. `StoredValue::new_local` wraps its value in a `SendWrapper` pinned to the tokio worker thread that created it, but leptos re-runs/disposes the Suspense owner across `.await` points on the multithreaded runtime — cleanup on a different worker thread is a guaranteed panic, probabilistically (~2/3 of requests here) killing the response stream.

## Fix

The three `new_local` sites are pure client listener bookkeeping (scroll-sync closures, window scroll/resize callback, per-row ResizeObserver). Compile them out of the SSR build with `#[cfg(feature = "hydrate")]`:

- `analyzer.rs` — `hscroll_listeners` + its `on_cleanup`/`Effect`
- `virtual_scroller.rs` — the `if is_window` listener block and the per-row `resize_observer` block

They emit no markup, so SSR HTML is byte-identical and hydration is unaffected; the hydrate (client) build keeps the code unchanged. `Fenwick::add` becomes hydrate-only, hence the targeted `cfg_attr(allow(dead_code))`.

**Durable rule this encodes: no `StoredValue::new_local` / `signal_local` may be created in an SSR-compiled component path.**

## Verification

- `./check_ci.sh` → exit 0 (fmt + workspace clippy `-D warnings`, incl. the `ultros` bin)
- `cargo check -p ultros-app --no-default-features --features hydrate --target wasm32-unknown-unknown` → 0 warnings
- `cargo test -p ultros-app` → 439 passed
- The companion `arc_memo.rs:334` unwrap-None panic seen in the same log windows looks like collateral of the aborted owner; worth re-checking prod logs after this deploys.

### DIFF
--- a/ultros-frontend/ultros-app/src/components/virtual_scroller.rs
+++ b/ultros-frontend/ultros-app/src/components/virtual_scroller.rs
@@ -1,9 +1,11 @@
 use leptos::prelude::*;
 use std::hash::Hash;
 use std::{cell::RefCell, rc::Rc};
+#[cfg(feature = "hydrate")]
+use web_sys::ResizeObserver;
 use web_sys::wasm_bindgen::JsCast;
 use web_sys::wasm_bindgen::closure::Closure;
-use web_sys::{HtmlDivElement, ResizeObserver, window};
+use web_sys::{HtmlDivElement, window};
 
 struct Fenwick {
     n: usize,
@@ -21,6 +23,9 @@ impl Fenwick {
         self.bit.clear();
         self.bit.resize(n + 1, 0.0);
     }
+    // Only called from the hydrate-gated ResizeObserver effect; the SSR
+    // build never measures rows, so the tree is only ever read there.
+    #[cfg_attr(not(feature = "hydrate"), allow(dead_code))]
     fn add(&mut self, mut idx: usize, delta: f64) {
         // fenwick tree is 1-based internally
         idx += 1;
@@ -231,6 +236,14 @@ where
     // Window-scroll mode: the container no longer scrolls, so its `on:scroll`
     // handler is inert. Drive `scroll_offset` and `window_height` from the
     // page instead.
+    //
+    // Client-only: the `LocalStorage` StoredValue below must never exist in
+    // the SSR build — Suspense re-runs/disposes the owner across `.await`
+    // points on the multithreaded runtime, so its cleanup (or drop) can land
+    // on a different worker thread than the one that created the SendWrapper,
+    // which panics and truncates the response stream (see analyzer.rs's
+    // scroll-sync block for the full story).
+    #[cfg(feature = "hydrate")]
     if is_window {
         let sticky_offset = match source {
             ScrollSource::Window { sticky_offset } => sticky_offset,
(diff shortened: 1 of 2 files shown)


akarras:
Reviewed and verified locally. Approving and merging.

**The diagnosis holds up.** `StoredValue::new_local` wraps its value in a `SendWrapper` bound to the tokio worker that created it; leptos disposes the Suspense owner across `.await` points on a multithreaded runtime, so cleanup can land on a different worker — a guaranteed panic that aborts the response mid-body. Since leptos emits the `__INCOMPLETE_CHUNKS` bootstrap only as the *final* chunk, a truncated stream is exactly the `ReferenceError` -> `RefCell already borrowed` -> `unreachable` cascade reported. Gating the three sites out of the SSR build is the right fix, not a workaround: all three are pure client listener bookkeeping that emit no markup, so SSR HTML is unchanged and hydration is unaffected. I confirmed the gated `is_window` block is inert under SSR anyway (Effects do not run server-side, and `window()` is `None`), so there is no behavioral loss.

**Verification** (CI runs fmt + clippy only — `cargo test --workspace` is commented out in `rust.yml`):

- `cargo check -p ultros-app --no-default-features --features hydrate --target wasm32-unknown-unknown` — **exit 0**. This matters more than usual here: the entire change is `#[cfg(feature = "hydrate")]`, so the SSR-only build CI lints cannot prove the client build still compiles. It does.
- `cargo test -p ultros-app --lib` — 448 passed / 0 failed.
- Verified in the merged tree that the only `new_local` in `analyzer.rs` (line 1464) sits inside the gate at 1459.

**Merge-order note:** #1136 rewrites this same region of `analyzer.rs` and was branched before this PR, so it does not carry the gate. I tested the merge — it resolves clean and the gate survives with the `new_local` still inside it — but this PR must land first.

PR #1136, opened 2026-08-05
perf(flip-finder): update the table in place on a realtime market tick

akarras:
## The problem

Every relevant websocket listing event rebuilt the whole Flip Finder — which, from the user's side, is what happens each time someone searches that item in game.

`AnalyzerTable` took the three resolved market boards as plain values (`sales: RecentSales`, `…: CheapestListings`), so the `{move || …}` closure inside `<Suspense>` had to dispose and rebuild it whenever a board refetched. Worth stating plainly since the file's comments assumed otherwise: **`<Suspense>` was not the cause** — it builds its children exactly once (`EitherKeepAlive` only toggles which branch shows). The closure returning a freshly built `<AnalyzerTable/>` was.

Each rebuild discarded:

- the `VirtualScroller`'s scroll position and measured row heights
- the accumulated ClickHouse `enrichment` map — so every visible row re-fetched its quality and sparkline data
- the `requested` dedupe set backing that fetch
- the realtime subscription that had just delivered the event

and the Suspense boundary dropped to `AnalyzerTableSkeleton` on top, flashing the table away underneath whoever was reading it.

## The fix

Hoist the expensive derivation into one `market` memo in `AnalyzerWorldView` and pass `AnalyzerTable` a `Memo<Option<ProfitTableHandle>>`. A tick now invalidates `market → profits → filtered_rows → sorted_data`, and the `VirtualScroller`'s keyed `<For>` diffs only the rows that moved. Everything else stays put.

Details a reviewer should know:

- **`ProfitTableHandle` compares by `Arc::ptr_eq`.** `Memo` requires `PartialEq`, and comparing ~20k rows element-wise costs more than the rebuild that produced them. Identity is the real question — a rebuild always yields a fresh `Arc`.
- **`market` holds the previous table across a refetch.** A leptos resource reports `None` while loading, *including on a refetch of data it already has*; following that would empty the table for the length of every fetch — the flash this PR exists to remove. It tracks the world the boards belong to, so a world change drops the table rather than showing stale rows under a new heading (world changes still show the skeleton, as before).
- **`<Transition>` rather than `<Suspense>`, gated on a `bool` memo via `<Show>`.** Gating on the value memo would reintroduce the rebuild.
- **The registration closure is load-bearing.** A memo runs under its own owner, so its resource reads do *not* register with the suspense boundary — without a registration point the server would stream the skeleton instead of awaiting. The sibling closure inside the `<Transition>` reads each resource and renders nothing; `dry_resolve` invokes closure views (`tachys/src/reactive_graph/mod.rs`), which is what registers the tasks. Please don't delete it as dead code.

## Narrowed + debounced refetch

One shared `market_refresh_version` meant every event refetched the world board, the region board, *and* every connected region's board — megabytes of JSON for a change that touched one row, with no debounce at all.

`classify_market_update` maps an event's `world_id` to the boards it can actually have moved, feeding three counters instead of one. Ticks coalesce over 400ms and flush once with the scopes unioned. `Stale` (which names no world) and unresolvable worlds fail open to all three boards — a board that silently stops refreshing is a worse failure than an extra fetch.

## Cleanup

`name_chip_pending` and `hydrated` move back into `AnalyzerTable`. They were hoisted to the parent only to survive the remount, and several comments across the file describing that remount as a fact of life are now gone.

## Testing

- `cargo test -p ultros-app --lib` — 445 passed, including 6 new tests covering the scope classifier (own world / same region / other region / `Stale` / unresolvable world) and the burst-merge
- `cargo clippy -p ultros-app --features hydrate --target wasm32-unknown-unknown` — clean for `analyzer.rs`. Note CI never lints the `hydrate` cfg; the 6 lints it surfaces are pre-existing, in `on_hand_input.rs` and `ws/realtime.rs`, and untouched here.

**Not verified in a browser.** The behavioural claims are from the type-level change plus reading the leptos 0.8.20 / tachys sources, not from watching a live tick land. The SSR path in particular (`<Show>` inside `<Transition>` still rendering the resolved table rather than the skeleton) is reasoned from `to_html_async_with_buf` and `dry_resolve`, so it's worth a look at a served page before this goes out.

### DIFF
--- a/ultros-frontend/ultros-app/src/routes/analyzer.rs
+++ b/ultros-frontend/ultros-app/src/routes/analyzer.rs
@@ -156,7 +156,10 @@ use std::{
 use ultros_api_types::{
     cheapest_listings::CheapestListings,
     recent_sales::{RecentSales, SaleData},
-    websocket::{FilterPredicate, SocketMessageType, is_analyzer_market_update_relevant},
+    websocket::{
+        EventType, FilterPredicate, ServerClient, SocketMessageType,
+        is_analyzer_market_update_relevant,
+    },
     world_helper::{AnyResult, AnySelector, WorldHelper},
 };
 use web_sys::wasm_bindgen::JsCast;
@@ -309,6 +312,57 @@ fn sort_rows(rows: &mut [CalculatedProfitData], mode: SortMode, dir: SortDir) {
 
 #[derive(Clone, Debug)]
 struct ProfitTable(Vec<Arc<ProfitData>>);
+
+/// Cheap-to-compare handle on a built [`ProfitTable`].
+///
+/// The table is rebuilt from scratch every time the market boards refetch and
+/// holds ~20k rows, so it lives in a `Memo` — and `Memo` needs `PartialEq`.
+/// Comparing the tables element-wise on every realtime tick would cost more
+/// than the rebuild that produced them, and identity is the question actually
+/// being asked (a rebuild always yields a fresh `Arc`), so this compares by
+/// pointer instead.
+#[derive(Clone, Debug)]
+struct ProfitTableHandle(Arc<ProfitTable>);
+
+impl ProfitTableHandle {
+    fn new(table: ProfitTable) -> Self {
+        Self(Arc::new(table))
+    }
+
+    fn rows(&self) -> &[Arc<ProfitData>] {
+        &self.0.0
+    }
+}
+
+impl PartialEq for ProfitTableHandle {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
+    }
+}
+
+/// The market boards as the table sees them.
+#[derive(Clone, Debug, PartialEq)]
+struct MarketState {
+    /// World these boards belong to. Tracked so a world change can drop
+    /// `table` rather than carry it — every other transition keeps it.
+    world: String,
+    /// The built table, held across a refetch.
+    ///
+    /// A leptos resource reports `None` while it is loading, including on a
+    /// refetch of data it already has. Following that to `None` would empty
+    /// the table for the length of every realtime tick, which is the flash
+    /// this is here to avoid; the previous table stays up until the new one
+    /// is ready.
+    table: Option<ProfitTableHandle>,
+    /// At least one board resolved to an error. Only actionable when there is
+    /// no table to fall back on.
+    failed: bool,
+}
+
+/// How long realtime market ticks are coalesced before the affected boards are
+/// refetched. A busy world delivers relevant listing events in bursts, and each
+/// one used to trigger its own round of fetches.
+const MARKET_REFRESH_DEBOUNCE_MS: u32 = 400;
 
 fn listings_to_map(listings: CheapestListings) -> HashMap<ProfitKey, (i32, i32)> {
     listings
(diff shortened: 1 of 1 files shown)


akarras:
Reviewed and merged. This is careful work and the diagnosis is right — but the important part of this review is that **I was able to verify the one claim you flagged as unverified.**

## The SSR registration closure does work — confirmed in the dependency source

You reasoned this one from reading and asked for a look at a served page. I checked it against the actual crate instead, which I think settles it more firmly than a page load would:

`tachys-0.2.18/src/reactive_graph/mod.rs:131` —
```rust
impl<F, V> RenderHtml for F where F: ReactiveFunction<Output = V>, V: RenderHtml + 'static {
    fn dry_resolve(&mut self) { self.invoke().dry_resolve(); }
```

`dry_resolve` on a closure view calls `self.invoke()` — it **runs the closure body** before recursing. So the sibling closure's `register_*.get()` reads really do execute during the dry-resolve pass and register with the enclosing `<Transition>`, which is exactly what you claimed and exactly why the server awaits the boards instead of streaming `AnalyzerTableSkeleton`. The closure returning `()` is fine — the unit impl's `dry_resolve` is a no-op, but the reads have already happened by then.

Your comment asking that it not be deleted as dead code is well placed; I'd go further and say that source citation belongs in the comment, because "this closure renders nothing" is going to look deletable to the next reader.

## Verified independently

- `cargo test -p ultros-app --lib` on the merged tree — **448 passed / 0 failed** (439 base + 3 from #1135 + your 6). The `market_scope` module covers own-world / same-region / other-region / `Stale` / unresolvable / burst-merge.
- `cargo check -p ultros-app --no-default-features --features hydrate --target wasm32-unknown-unknown` — exit 0, and the full `cargo leptos` wasm build also succeeded. Worth doing since CI never lints that cfg.
- Merges clean onto current `main` (now carrying #1123/#1124/#1134/#1135), and I confirmed **#1134's SSR gate survives the merge** with the lone `new_local` still inside it. That mattered: this branch predates #1134 and does not carry the gate itself.
- `ProfitTableHandle`'s `Arc::ptr_eq` is the right call — a rebuild always yields a fresh `Arc`, so identity answers the question element-wise comparison of ~20k rows would answer far more expensively.
- Holding the previous table across a refetch is the crux, and the reasoning is correct: a leptos resource reports `None` on a refetch of data it already has, so following it would reintroduce the exact flash this removes. Keying `MarketState` on the world so a world change drops the table instead is the right exception.
- Good discipline on `StoredValue::new` (not `new_local`) for `pending_scope`/`flush_id` — that keeps the debounce out of #1134's SendWrapper trap — and on `try_update_value`/`try_with_value` for the disposal case.
- Failing open to `MarketScope::ALL` on an unresolvable world is the correct bias.

## What I did *not* verify

**No browser run.** I built the merged tree locally to check a served page, but a concurrent session took over the worktree mid-build, and I chose the source-level proof above over restarting a 40-minute build. So the behavioural claims — no skeleton flash, scroll position and ClickHouse enrichment surviving a tick — remain reasoned rather than observed. I'm comfortable merging because the failure mode if the memo plumbing were wrong is *reverting to today's behaviour* (a rebuild), not a break; the one claim that could have been a genuine regression was the SSR path, and that one is now confirmed.

Worth a glance at GlitchTip for `analyzer.rs` after this deploys, and at the flip-finder page under a live tick.

## Heads-up on merge order

#1137 and #1138 both touch `analyzer.rs` and were branched before this. Landing this one first is deliberate — it is the hub of that cluster, so the two small PRs rebase onto it rather than forcing a 531-line rebase the other way. Expect both to need a rebase now.

PR #1137, opened 2026-08-05
fix(flip-finder): render every enabled column at every width

### LINKED ISSUE #1111: remove desktop only filters from flip finder
It serves no purpose to have desktop only columns when users can customize their view.

akarras:
Closes #1111.

## The bug

The Flip Finder's columns picker offers eleven optional columns. Nine of them carried a breakpoint gate (`hidden md:flex`, `hidden lg:block`, `hidden xl:flex`), so on a phone ticking them changed nothing — the checkbox toggled, the URL changed, and the column stayed `display: none`. The picker even shipped a "desktop only" note next to those nine, which explained the dead checkbox rather than fixing it.

Measured on prod at narrow width with **every** column enabled (`?cols=profit_per_day,velocity,drift,confidence,world,last_sold,roi,datacenter,trend,sales_per_day,volume_30d`):

| column | rendered width |
|---|---|
| HQ, Item, Profit, Profit/day, ROI, Buy Price | 44 / 228 / 112 / 112 / 112 / 112 |
| Velocity, Drift, Confidence, World, Datacenter, Trend, Sales/day, 30d Volume, Last Sold | **0 — `display: none`** |

## The fix

The results grid is already a pair of synchronized horizontal scrollports (`.analyzer-hscroll`), so a narrow viewport can *scroll* to a column rather than hide it. Every optional header, cell and skeleton column loses its breakpoint gate; `?cols=` becomes the only thing that decides what renders.

## What that simplifies

The width reservation existed in three breakpoint buckets (`--analyzer-extra-cols-{base,md,xl}`) for exactly one reason: reserving a hidden column's width would have given a phone a scroll range whose far end was blank. With nothing hidden that concern disappears and the reservation is one number.

- `ExtraColumnWidths` + `extra_column_widths_px` → `optional_column_width_px`, a plain sum over all eleven columns.
- Four `@media` blocks in `style/tailwind.css` → one rule. The baseline is now just the four always-on columns: `30.75rem` = HQ 44px + Item 14rem + Profit 7rem + Buy Price 7rem, exact rather than the old rounded-up 38/61/68rem tiers.
- `col_hidden_note_class` and the picker's "desktop only" note are gone, along with `analyzer_columns_picker_desktop_only` in all seven locale files.

Also picked up: Trend, 30d Volume and Last Sold were missing `shrink-0` on their body cells. That never showed because those cells were `display: none` at precisely the widths where the row is tightest.

## Tests

`the_default_column_set_adds_no_extra_width`, `every_opt_in_column_reserves_width`, `breakpoint_hidden_columns_reserve_no_width_below_their_breakpoint` and `hidden_note_matches_the_width_buckets` all encoded the bucketing contract that no longer exists. Replaced with `no_optional_columns_reserves_no_extra_width`, `every_optional_column_reserves_width` (now covers all eleven, not just the non-default ones) and `column_reservations_add_up`, which pins that the reservation is a plain sum.

## Verification

- `cargo test -p ultros-app --lib analyzer::` — **68 passed, 0 failed**

Note this widens the default mobile scroll range: with the default six optional columns the row min-width goes from 604px to 1076px on a phone. That is the point of the issue — the columns are reachable now instead of silently dropped — but it is a visible change to the default phone view, so flagging it.

Heads-up on overlap: #1136 is open against `analyzer.rs` too. It touches the realtime/sort paths and the table wrapper; this touches column class strings, the width helper and the picker. They should merge cleanly but whichever lands second is worth a look.

### DIFF
--- a/style/tailwind.css
+++ b/style/tailwind.css
@@ -2373,41 +2373,15 @@ a:not(.nav-link):not(.side-nav-item):not(.menu-item):not(
    rows, so the two scrollports scroll through the same range and the row
    striping covers it.
 
-   The base value tracks the columns each breakpoint actually reveals (several
-   are `hidden md:flex` / `lg:flex`). The `--analyzer-extra-cols-*` variables
-   are set inline by the route and carry the optional columns the user has
-   switched on beyond the default set — the stylesheet cannot know those. They
-   are bucketed by the breakpoint at which the column first renders (`base` =
-   always, `md` = `hidden md:flex` columns, `xl` = `hidden xl:flex` columns) so
-   a viewport only reserves width for columns it actually shows; adding a
-   hidden column's width would extend the scroll range into blank space. */
+   No breakpoint enters into it: every column the user switched on renders at
+   every viewport width, and a narrow screen scrolls to reach them. The 30.75rem
+   here is the four always-on columns (HQ 44px + Item 14rem + Profit 7rem + Buy
+   Price 7rem); `--analyzer-optional-cols` is set inline by the route and carries
+   whatever `?cols=` turned on, which the stylesheet cannot know. */
 .analyzer-table {
-    --analyzer-row-min-width: calc(38rem + var(--analyzer-extra-cols-base, 0px));
-}
-@media (min-width: 768px) {
-    .analyzer-table {
-        --analyzer-row-min-width: calc(
-            61rem + var(--analyzer-extra-cols-base, 0px) +
-                var(--analyzer-extra-cols-md, 0px)
-        );
-    }
-}
-@media (min-width: 1024px) {
-    .analyzer-table {
-        --analyzer-row-min-width: calc(
-            68rem + var(--analyzer-extra-cols-base, 0px) +
-                var(--analyzer-extra-cols-md, 0px)
-        );
-    }
-}
-@media (min-width: 1280px) {
-    .analyzer-table {
-        --analyzer-row-min-width: calc(
-            68rem + var(--analyzer-extra-cols-base, 0px) +
-                var(--analyzer-extra-cols-md, 0px) +
-                var(--analyzer-extra-cols-xl, 0px)
-        );
-    }
+    --analyzer-row-min-width: calc(
+        30.75rem + var(--analyzer-optional-cols, 0px)
+    );
 }
 .analyzer-grid-row {
     min-width: var(--analyzer-row-min-width, 0px);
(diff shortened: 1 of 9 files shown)


akarras:
Heads-up from the PR-review pass that just landed #1136 (`perf(flip-finder): update the table in place on a realtime market tick`, +531/-99 in `analyzer.rs`).

This PR was branched before that, and it deletes 230 lines from the same file — the classic shape for a merge that resolves clean textually and then fails to compile. **I checked, and it is fine:**

- `git merge-tree origin/main <this branch>` — clean, no conflicts.
- `cargo check -p ultros-app` on a tree merging current `main` + this PR + #1138 — **Finished, 0 errors** (7m37s).

So no rebase is required on either count. Note that your CI ran against a base without #1136, so that green check does not by itself cover the combination — the check above does.

One thing to be aware of when you next touch this file: #1136 changed `AnalyzerTable`'s props from plain values to signals (`filter_outliers`, `region`, `cross_region_enabled` are now `Signal<_>`, and the table takes `profits: Memo<Option<ProfitTableHandle>>` instead of the three raw boards), and moved `name_chip_pending`/`hydrated` back into `AnalyzerTable`. The component is no longer rebuilt on every realtime market tick, so state declared inside it now survives ticks.

akarras:
**Reviewed — merged-tree verified.**

Checked the things the diff can't show on its own:

- **Baseline arithmetic is right.** 30.75rem = HQ 2.75rem (44px) + Item 14rem + Profit 7rem + Buy Price 7rem. ✓
- **The markup sweep is complete.** The only `hidden md:`/`lg:`/`xl:` left in `analyzer.rs` are the two `sticky-bar-button-label` spans, which are the icon-only-on-mobile pattern for the height-locked bar and correctly out of scope.
- **`analyzer_columns_picker_desktop_only` is gone from all seven locale files**, not just `en`.
- Every width in the new `WIDTHS` table matches its `w-[..]`/`w-28` in the markup.

Confirmed the premise on prod at 375px before reading the fix: with every column enabled the nine gated columns really do render at width 0.

Agreed with the call on the flagged behaviour change (default mobile row min-width 604px → 1076px). That's the issue being fixed, not a side effect — a reachable-by-scrolling column beats a silently dead checkbox.

CI green (lint + build). On a local integration branch of #1139+#1137+#1140+#1141: `cargo fmt --check` clean, `clippy -p ultros-app --all-targets -D warnings` clean, **451 tests passed / 0 failed**.

LGTM. Should land before #1138, which depends on it.

Issue #1133, opened 2026-08-04
UI kit migration: core work complete; remaining visual cleanup


akarras:
The original UI component migration is complete apart from the remaining visual cleanup in #1132. This is the older migration tracker; use #1349 for shared analyzer data, columns, windows, and filters.

**Status checked against main and issue history on 2026-09-08:**

- [x] Shared sort header: #1125.
- [x] Shared control bar: #1126.
- [x] Toolbar-route conversion and old Toolbar removal: #1127.
- [x] Shared data-table infrastructure: #1080.
- [x] Currency Exchange: #1128.
- [x] Retainers: #1129.
- [x] Alerts and drawer consolidation: #1130.
- [x] Lists/List View: #1131.
- [x] Loading-state and skeleton sweep: #1223, the completed portion of #1132.
- [ ] Remaining card surfaces and Settings/Job Set Detail treatment: #1132.

These migrations shared the UI components; they did not move every page onto `MarketGrid`. The later Trends, Currency Exchange, Item Explorer, and Recipe work is explicitly tracked in #1349. The explorer's current filters/table were further improved in #1316; keep its next migration in #1346 rather than reopening the old work here.

Close this tracker when #1132 is resolved. The home dashboard keeps its separate visual treatment.

akarras:
Status check against main as of 2026-08-05 (verified in-repo, not just by PR titles):

**Done**
- [x] Phase 1 — #1125 shared `SortHeader` (PR #1140, closed)
- [x] Phase 2 — #1126 control bar extracted to `components/control_bar.rs` (PR #1143, closed); #1144 also put it in the loading skeleton

**Not started, now unblocked**
- [ ] Phase 3 — #1127: all seven routes still import `toolbar::{Toolbar, ToolbarField, ToolbarPills}`; `components/toolbar.rs` still exists. Control bar dependency is satisfied — this is next and mostly mechanical.
- [ ] Phase 4 — #1080: no `components/data_table.rs` yet; the five raw `<table>` sites are all still present (currency_exchange.rs:765, list_view.rs:1386, retainers.rs:121+209, alert_rules_panel.rs).

**Not started**
- [ ] #1128 Currency Exchange (raw table + hand-rolled chips still in place; #1142 fixed the icons but didn't touch structure)
- [ ] #1129 Retainers, #1130 Alerts (both drawer files still exist), #1131 Lists, #1132 hygiene (groups.rs animate-pulse gray still breaks light mode, at lines 519/694 now)

[...]

Issue #1080, opened 2026-08-02
Extract a reusable sortable data-table component (item explorer + currency exchange)


akarras:
Split out of #1058. That issue proposed reusing the item explorer's "spreadsheet style" table on the currency exchange page. While implementing it I found the explorer's table is not reusable in its current form, so the currency exchange PR adopts the explorer's *visual language* by hand instead. This issue tracks doing the extraction properly.

### Why it isn't reusable today

- `ItemList` (`ultros-frontend/ultros-app/src/routes/item_explorer.rs:643`) is a private component hard-typed to `Memo<Vec<(&'static ItemId, &'static Item)>>`. Nothing else can render rows through it.
- `SortHeader` (`item_explorer.rs:581`) is private too, and it is the only piece that knows how to build a sort href that preserves the rest of the query string.
- It hard-depends on explorer-specific contexts — `CheapestPrices` and `ExplorerPriceScope` (`item_explorer.rs:649-652`) — and panics via `.expect(...)` if they are absent.
- The "spreadsheet" is not a component at all: it is an inline `<div role="table">` (`item_explorer.rs:850-895`) whose column template lives in hand-duplicated Tailwind class strings.

### The maintenance hazard

The header row's grid template at `item_explorer.rs:855` and the body row's at `item_explorer.rs:890` are separate string literals that must stay character-identical or the columns stop lining up. Each exists in a single-world and a multi-world variant, so a column change means editing four long strings by hand with no compiler or test to catch a mismatch. The currency exchange PR adds another hand-copy of the same look, which makes this worse — that is the debt this issue should retire.

### Suggested direction

Extract something like `components/data_table.rs` exposing a public sortable table whose columns are described once as data (label, i18n key, grid width, alignment, sortable-or-not, cell renderer) and whose header and body derive their grid template from that single description. Then port `ItemList` and the currency exchange results table onto it. Worth confirming up front whether a `role="table"` div grid or a real `<table>` is the better substrate — the explorer uses the former, the currency exchange page a real `<table>`, and only one should survive.

### Out of scope for #1058

The currency exchange PR deliberately does not attempt this. Extracting a generic table is a refactor across at least two routes with real hydration-order risk (see the `hydrated` gate comment at `item_explorer.rs:684`), and bundling it with a UX cleanup would make both harder to review and to revert.

akarras:
[...]

PR #1140, opened 2026-08-05
refactor(ui): extract a shared SortHeader, fix vendor resale's stuck sort

### LINKED ISSUE #1125: UI kit: extract a shared SortHeader component (4 copies today, one has a live arrow bug)
Part of the UI-kit unification effort (see umbrella issue).

`SortHeader` — the sortable column link whose arrow reflects the actual sort direction and omits the default `?dir=` — exists as four independent implementations:

- `ultros-frontend/ultros-app/src/routes/analyzer.rs:1001-1054` (the reference)
- `ultros-frontend/ultros-app/src/routes/scrip_sources.rs:392-445` — byte-identical to the analyzer's except the fallback mode/dir
- `ultros-frontend/ultros-app/src/routes/item_explorer.rs:582-600` — string-keyed variant with different props
- `ultros-frontend/ultros-app/src/routes/vendor_resale.rs:636-660` — inline `QueryButton` + a **hardcoded** `BiSortDownRegular`, so `?dir=asc` renders ascending rows under a descending arrow. This is exactly the bug the analyzer's doc comment says its version was written to fix.

### Task

Extract `components/sort_header.rs`, generic over the route's sort mode (something like `SortMode: Display + Default` plus a default-direction hook), and port all four sites onto it. The vendor-resale arrow bug disappears as a side effect.

Keep the query-string-preserving href behavior — that's the part every copy gets subtly different. Item explorer's port can be minimal (just the header cell); its full table move is tracked in #1080.

akarras:
Phase 1 of the UI-kit unification umbrella (#1133). Closes #1125.

Four routes had grown their own sortable column header and disagreed on the details that matter: which direction a fresh click applies, whether the arrow reflects the direction actually in effect, and whether `?dir=` is written when it is redundant.

## What's shared now

`components/sort_header.rs` owns the contract, generic over the route's sort enum via a small trait:

```rust
pub trait SortColumn: Copy + PartialEq + Display + Send + Sync + 'static {
    fn fallback() -> Self;                                  // what `?sort=` absent means
    fn default_dir(self) -> SortDir { SortDir::Desc }       // per-column, overridable
}
```

`SortDir` is shared instead of redeclared per route (three copies of the same `asc`/`desc` `FromStr`+`Display` are gone). The flip rule and the href builder are free functions, so the query-string-preserving behaviour — the part every copy got subtly different — is pinned by tests without standing up a router.

## Ports

| route | before |
|---|---|
| flip finder | the reference implementation; moved out verbatim |
| scrip sources | a near-identical fork, per-column defaults preserved (costs read best-first ascending) |
| item explorer | string-keyed variant. Keeps its `role="columnheader"` cell and `aria-sort`; its `page` reset now goes through `reset_keys`. Local `SortDirection` — which parsed `"dsc"` but printed `"desc"` — deleted |
| vendor resale | inline `QueryButton` + hardcoded down arrow |

## Behaviour changes

- **Vendor resale sorting was stuck descending.** The route parsed `?sort=` but never `?dir=`, so the one order it could produce was also the only one the (hardcoded) arrow claimed. It now reads `?dir=` and the header can reach ascending.
- **Item explorer's arrows** become the shared icon set instead of `↑`/`↓` text, and the link picks up the shared active/inactive brand colours — the intended convergence, and the reason its full table conversion is still tracked under #1080.
- Each route's own sort now falls back through `SortColumn::fallback()` rather than a duplicated literal, so rows are ordered by exactly what the header highlights.

## Verification

- `./check_ci.sh` — clean (fmt + clippy `-D warnings`)
- `cargo test -p ultros-app --lib` — 446 passed. Seven new tests in `sort_header` cover the flip rule, per-column defaults, default-dir omission, param preservation and `reset_keys`; the analyzer's copies of those tests (which tested a re-implementation of the rule, not the code that shipped) are replaced by one that pins the flip finder's per-column defaults.

### DIFF
--- a/ultros-frontend/ultros-app/src/components/mod.rs
+++ b/ultros-frontend/ultros-app/src/components/mod.rs
@@ -70,6 +70,7 @@ pub mod select;
 pub mod side_nav;
 pub mod skeleton;
 pub mod small_item_display;
+pub mod sort_header;
 pub mod sparkline;
 pub mod stats_display;
 pub mod theme_picker;
--- a/ultros-frontend/ultros-app/src/components/sort_header.rs
+++ b/ultros-frontend/ultros-app/src/components/sort_header.rs
@@ -1,0 +1,303 @@
+//! Sortable column header, shared by every tool table.
+//!
+//! Four routes had grown their own copy of this — the Flip Finder's (the
+//! reference), Scrip Sources' near-identical fork, the Item Explorer's
+//! string-keyed variant, and Vendor Resale's inline `QueryButton` with a
+//! hardcoded down arrow. They disagreed on the details that matter: which
+//! direction a fresh click applies, whether the arrow reflects the direction
+//! actually in effect, and whether `?dir=` is written when it is redundant.
+//!
+//! The contract here:
+//!
+//! * clicking an inactive column sorts by it in that column's
+//!   [`SortColumn::default_dir`]; clicking the active column flips it,
+//! * the arrow always reflects the direction actually applied,
+//! * `dir` is omitted from the href when it matches the column's default, so
+//!   the common case stays a clean `?sort=…` and bookmarks don't accumulate a
+//!   redundant param,
+//! * every other query param survives — that's the part each copy got subtly
+//!   different.
+
+use leptos::prelude::*;
+use leptos_router::hooks::use_location;
+use leptos_router::location::Location;
+use leptos_router::params::ParamsMap;
+
+use crate::components::icon::Icon;
+use icondata as i;
+
+/// `?dir=` — sort direction. Absent means the active column's
+/// [`SortColumn::default_dir`].
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Default)]
+pub enum SortDir {
+    Asc,
+    #[default]
+    Desc,
+}
+
+impl SortDir {
+    /// The direction a click on the already-active column moves to.
+    pub fn flipped(self) -> Self {
+        match self {
+            SortDir::Asc => SortDir::Desc,
+            SortDir::Desc => SortDir::Asc,
+        }
+    }
+}
+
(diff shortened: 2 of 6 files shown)


akarras:
**Reviewed — merged-tree verified.**

The shared contract is sound and the tests pin the parts that actually differed. Traced the logic for consistency:

- `dir` resolves via the *active* column's `default_dir()`, and for an inactive column `next_dir` ignores `current` entirely and returns `mode.default_dir()` — so the two paths can't disagree. ✓
- Omitting `dir` when it equals `mode.default_dir()` round-trips correctly: a bare `?sort=cost` re-reads as `Cost.default_dir()`. ✓
- `item_explorer`'s `column_aria_sort` computes active/direction the same way `SortHeader` does, so ARIA can't drift from the arrow. `fallback() = ItemLevel` matches the old `default=true`. ✓
- `scrip_sources` per-column defaults preserved exactly (`CostPerScrip | Cost => Asc`). ✓

**Both bug fixes are real**, not just cleanup:
- vendor resale genuinely never parsed `?dir=` — the hardcoded down arrow was the only honest thing about it.
- item explorer's old `SortDirection` parsed `"dsc"` but printed `"desc"`, so its own emitted URL failed to parse back and silently fell to the default.

One note, **not a blocker**: `label` is a `String` from `t_string!`, so headers won't re-render on a live locale switch the way the old inline `t!` did. There's clear precedent for this exact pattern on component props (`checked_label=t_string!(...).to_string()` in `add_recipe_to_list.rs` and `add_set_to_list.rs`), so it's consistent with the codebase rather than a regression this PR introduces — worth a separate pass across all of them if live locale switching should update in place.

Integration branch of #1139+#1137+#1140+#1141: fmt clean, clippy clean, **451 tests passed / 0 failed**.

LGTM.

PR #1141, opened 2026-08-05
fix(flip-finder): show that the filter strip has more filters off-screen

### LINKED ISSUE #1057: the sticky header on the item explorer on mobile doesn't show the full filter list

akarras:
Closes #1057.

## The title says sticky header; the sticky header is fine

I checked that first. At 375px on prod, across six scroll positions, the table header pins exactly where it should:

| scrollY | control bar | table header |
|---|---|---|
| 4000 | top 0, h 76 | **top 76** |
| 3000 | top 0, h 76 | **top 76** |
| 1500 | top 0, h 76 | **top 76** |
| 600 | top 0, h 76 | **top 76** |
| 200 | top 30 | top 123 |
| 0 | top 230 | top 323 |

No clipping, no drift. The sticky-bar work that landed after this issue was filed (#1082/#1083/#1105) appears to have taken care of whatever the screenshot caught.

## What is actually broken is "the full filter list"

`.filter-chip-row` scrolls horizontally but deliberately draws no scrollbar — the control bar is height-locked at 76px (the table header sticks directly beneath it at that offset), so a scrollbar gutter would eat into the 32px row the chips sit in. The consequence was never handled: nothing on screen says the strip scrolls.

Measured on prod at 375px with eight filters set (`?profit=1000&roi=50&confidence=medium&drift=5&min-volume=10&min-buy=500&max-price=900000&quality=hq`):

```
.filter-chip-row  scrollWidth 1002   clientWidth 237
chips at x = 32, 156, 254, 360, 508, 620, 715, 895
```

Everything past `Budget ≤ 900000` — Quality, Drift, Confidence, 30d volume — is off-screen with no scrollbar, no fade and no arrow. The strip *is* swipeable, so the filters are reachable; you just have no way to know they exist.

## The fix

Fade the strip at whichever edge has chips behind it. Two custom properties, `--chip-fade-start` and `--chip-fade-end`, feed a `mask-image` declared beside the strip's own rules; `analyzer.rs` sets them from the row's scroll geometry.

Both default to `0px`, which collapses each gradient stop pair onto the same position — a zero-width fade. **A strip whose chips all fit is byte-for-byte unchanged.**

Re-derived on three triggers: `scroll`, window `resize`, and whenever `active_filters()` changes — that last one matters because adding or removing a chip moves `scrollWidth` without firing either listener.

Verified the logic against prod by driving the real element:

| scrollLeft | remaining right | `--chip-fade-start` | `--chip-fade-end` |
|---|---|---|---|
| 0 | 765 | `0px` | **`24px`** |
| 300 | 465 | **`24px`** | **`24px`** |
| 765 | 0 | **`24px`** | `0px` |
| back to 0 | 765 | `0px` | **`24px`** |

## Verification

- `cargo check -p ultros-client --target wasm32-unknown-unknown` — clean
- Stylesheet built with cargo-leptos' own `tailwindcss v4.1.10`; mask rule emitted intact

That wasm check is not optional here. The new block is `#[cfg(feature = "hydrate")]`, and CI's clippy compiles `ssr` (the default feature) — so **CI cannot see this code at all**. It caught two real errors on the first run: `el.style()` resolved to tachys' `ElementExt::style` (in scope via the leptos prelude, matching `HtmlDivElement` directly) instead of the inherent `web_sys::HtmlElement::style`, which needs a deref step and therefore loses method resolution. Fully qualified now, with a comment saying why.

Listeners are parked in a `StoredValue` and removed in `on_cleanup`, matching the horizontal scroll-sync block directly above it — a forgotten listener would keep firing after the component is disposed.

## Related

#1139 (mobile edge-to-edge padding) widens this strip from 237px to 284px on a phone, which reduces how often the fade is needed but does not remove the need for it — 284px still holds about two chips.

### DIFF
--- a/style/tailwind.css
+++ b/style/tailwind.css
@@ -2346,8 +2346,33 @@ a:not(.nav-link):not(.side-nav-item):not(.menu-item):not(
     flex: 1;
     /* The bar is height-locked, so a reserved scrollbar gutter would eat
        into the chips themselves. Scrolling still works via trackpad,
-       shift+wheel and focus. */
+       shift+wheel, touch and focus. */
     scrollbar-width: none;
+
+    /* ...which leaves nothing on screen to say the strip scrolls at all. At
+       375px with eight filters set, the chips run 1002px inside a 237px
+       viewport and the five past the fold are simply invisible (#1057). These
+       two variables drive an edge fade; analyzer.rs sets them from the row's
+       own scroll geometry, so the fade appears on the side that has more
+       chips and nowhere else. Both default to 0, which collapses each
+       gradient stop pair onto the same position — a zero-width fade, i.e. no
+       visual change at all when everything fits. */
+    --chip-fade-start: 0px;
+    --chip-fade-end: 0px;
+    -webkit-mask-image: linear-gradient(
+        to right,
+        transparent 0,
+        #000 var(--chip-fade-start),
+        #000 calc(100% - var(--chip-fade-end)),
+        transparent 100%
+    );
+    mask-image: linear-gradient(
+        to right,
+        transparent 0,
+        #000 var(--chip-fade-start),
+        #000 calc(100% - var(--chip-fade-end)),
+        transparent 100%
+    );
 }
 .filter-chip-row::-webkit-scrollbar {
     display: none;
--- a/ultros-frontend/ultros-app/src/routes/analyzer.rs
+++ b/ultros-frontend/ultros-app/src/routes/analyzer.rs
@@ -1347,6 +1347,96 @@ fn AnalyzerTable(
         });
     }
 
+    // --- Filter chip strip: edge fades ---------------------------------------
+    // The strip scrolls but shows no scrollbar (the bar is height-locked, so a
+    // gutter would eat the chips), which left nothing on screen to say there
+    // were more filters off to the right — at 375px with eight filters set the
+    // chips run ~1000px inside a ~240px viewport. `--chip-fade-{start,end}`
+    // drive a mask declared in the stylesheet; both are 0 unless there is
+    // actually something to scroll to on that side.
+    let chip_row = NodeRef::<leptos::html::Div>::new();
+    #[cfg(feature = "hydrate")]
+    {
+        let chip_listeners = StoredValue::new_local(
+            None::<(
+                web_sys::HtmlDivElement,
+                Closure<dyn FnMut()>,
+                Closure<dyn FnMut()>,
+            )>,
+        );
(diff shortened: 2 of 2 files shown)


akarras:
**Reviewed — found and fixed a real CI failure, then verified the behaviour on prod.**

### CI was red, and the PR body's reasoning about why it couldn't be is wrong

Clippy failed on `0d75dcc2` with two errors:
- `unnecessary_cast` — `el.scroll_left() as f64`. web_sys' `Element::scroll_left()` already returns `f64` (CSSOM View has `scrollLeft` as an unrestricted double while `scrollWidth` is a long — which is why the casts on the *next* line are correct and this one isn't).
- `useless_conversion` — `let el: web_sys::HtmlDivElement = el.into();`, where `NodeRef<Div>::get()` already yields `HtmlDivElement`.

The body says the block is `#[cfg(feature = "hydrate")]` and therefore *'CI cannot see this code at all'*. CI demonstrably did see it — the errors are reported against `ultros-app (lib)`. Worth correcting, because that assumption is what made the local `clippy … -- -D warnings` result look trustworthy.

Pushed both fixes as `84b46b3d`.

### Behaviour verified against prod

Drove the real `.filter-chip-row` at 375px with the eight filters from the body and ran the port of `apply_fades`:

| scrollLeft | remaining right | `--chip-fade-start` | `--chip-fade-end` |
|---|---|---|---|
| 0 | 723 | `0px` | `24px` |
| 300 | 423 | `24px` | `24px` |
| 723 (max) | 0 | `24px` | `0px` |
| back to 0 | 723 | `0px` | `24px` |

Matches the body's table and reverses cleanly. Reproduced the premise too: scrollWidth 1007 vs clientWidth 236, so five chips really are unreachable with no affordance.

Also checked:
- **The zero-fade default is genuinely a no-op** — at `0px` each gradient stop pair collapses onto the same position, so a strip that fits is unchanged. Confirmed the computed `mask-image` on a live page.
- **The listener can't be stranded.** `<div class="filter-chip-row" node_ref=chip_row>` is unconditional (the `Show` is *inside* it), so the node the `StoredValue` holds is never replaced. The `on_cleanup` symmetry is right.
- Tracking `active_filters()` to re-derive is necessary and correct — adding a chip changes `scrollWidth` without firing scroll or resize.

Integration branch of #1139+#1137+#1140+#1141 **with these fixes**: fmt clean, clippy `-D warnings` clean, **451 tests passed / 0 failed**.

LGTM once CI re-runs green on `84b46b3d`.

PR #1142, opened 2026-08-05
Extract item icons from the game files (fixes missing currency-exchange icons)

akarras:
## Why

https://ultros.app/currency-exchange renders blank icons for most currencies: universalis-assets only ships Lodestone-crawled icons for **marketable** items, so gil, company seals, tomestones, scrips, MGP, and most untradable exchange rewards have no icon at all.

## What

Icons are now read straight out of a local FFXIV install — no crawling, no assets repo, one toolchain (ironworks, all Rust):

- **New `icon-extract` crate**: opens the install via ironworks (SqPack + tex), prefers the native-2x `_hr1` variant (80px), decodes Dxt1/Argb8 (a full census of all 26k referenced icons found exactly those two formats; anything else fails loudly rather than shipping a corrupt image).
- **`game-data-pack`**: the universalis-assets git source is gone. Icons come from the install (auto-discovered; `--game-path` to override; `--skip-icons` on machines without the game). `data/manifest.toml` records the client version under `[icons]` as provenance. A >10% unreadable-icon tripwire refuses to write a gutted pack if extraction is systemically broken.
- **Pack format**: icons stored once per *icon* id with an `items.map` entry translating item→icon (items share icons heavily: 25,507 unique icons cover 49,764 items — storing per-item doubled the pack). Large is now the native 80px (sharper on hidpi than the old 60px), Medium stays 40px, Small is dropped — `ultros-xiv-icons` serves Medium bytes for Small requests. Display CSS boxes unchanged.

## Numbers

Regenerated against ironworks git HEAD, so these are what actually ships:

| | before | after |
|---|---|---|
| items with icons | 17,208 | **50,773** |
| unique icons | — | 26,223 |
| items with no icon | 1,009 | **0** |
| Large resolution | 60px (from 128px crawl) | 80px (native 2x) |
| LFS pack | 22MB | 55MB |

The generator reports **"every named item has an icon"** — no absent, no
unreadable, and no `Icon = 0` rows among named items. The xiv-db CSV packs
regenerated byte-identical.

## Root cause of the missing icons: BC7, not a stale client

An earlier revision of this PR said 1,009 items had no icon because the
generating client was a patch behind the pinned CSVs. That was wrong. The
client was current — `2026.07.16.0001.0000` is a build stamp, not a patch
number, and the game will not launch on an outdated client anyway.

The real cause: `ironworks 0.4.1`'s `file::tex::Format` predates the entire
Bcn2 group, so it cannot parse a **BC7** texture — ~9% of `ui/icon`. On top of
that, `GameInstall::icon` matched `Err(_) => continue`, so "present but
unparseable" and "absent from the install" collapsed into one result, and the
failure reported itself as a stale client.

Measured on a current install, sampling icon ids 1..250000:

| | ironworks 0.4.1 | git HEAD |
|---|---|---|
| readable | 1546 | **1735** |
| present-but-unreadable | **189** | **0** |
| highest readable id | 234951 | 246458 |

151 of the 189 were BC7; the other 38 were existing formats that also failed
under the older sqpack handling.

### Fixes

- Reads return `IconRead::{Image, Absent, Unreadable}`, matched on
  `ironworks::Error::NotFound` structurally rather than by string. An
  indexed-but-unreadable icon **aborts the run** naming path and reason instead
  of quietly shrinking the pack. Genuinely-absent icons keep the >10% tripwire,
  since CSVs legitimately run ahead of a client.
- Every format `ui/icon` actually uses is decoded: BC1, BC3, **BC7**, Bgra8,
  Bgr5a1, Bgra4 (plus BC2/BC4/BC5). Each verified by decoding real icons to PNG
  and looking at them.
- `examples/probe.rs` classifies an id range as readable / absent /
  present-but-unreadable; `examples/dump.rs` writes one sample per format.

### Dependency note

`ironworks` is pinned to git rev `3d5db8ff` — **no crates.io release contains
the Bcn2 formats**, so CI fetches it from GitHub at build time. The bump is not
a drop-in: the `ffxiv` feature is gone (install discovery is now
`sqpack::Install`), `Format` variants are renamed DXGI-style, and `File::read`
requires `Read + Seek`.

## Verification

- 40 unit tests across `icon-extract` / `game-data-pack` / `ultros-xiv-icons`, including a currency-coverage test that fails against the old pack
- Decoded output eyeballed (gil, Poetics, MGP, Dxt1 control) both from raw extraction and from the final pack
- `./check_ci.sh` green (fmt + workspace clippy + xiv-gen feature clippy); xiv-db `.rkyv` packs regenerated byte-identical

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,8 +13,9 @@ There are **no git submodules** in this repo anymore. FFXIV game data (item/reci
 
 - **Fresh clone**: run `git lfs install && git lfs pull` once. Without it, the `data/` files are LFS pointer text, not real content, and the build fails with an actionable error message rather than a cryptic panic.
 - **Worktrees**: no setup needed — `git worktree add` checks out LFS content the same as a normal clone as long as `git lfs install` has been run once on the machine.
-- **Regenerating packs**: `cargo run --release -p game-data-pack -- --pinned` rebuilds the packs from the pins already recorded in `data/manifest.toml` (reproducible, no version bump). Pass `--latest` instead to bump the pins to the newest upstream data and regenerate against that.
-- **`data/manifest.toml`**: records exactly which upstream commit/release each pack was generated from — this is the source of truth for "what version of game data is this."
+- **Regenerating packs**: `cargo run --release -p game-data-pack -- --pinned` rebuilds the packs from the pins already recorded in `data/manifest.toml` (reproducible for the CSV packs, no version bump). Pass `--latest` instead to bump the pins to the newest upstream data and regenerate against that.
+- **Icons need a local FFXIV install**: the icon pack is extracted straight out of the game's SqPack files (via the `icon-extract` crate), not fetched from a repo. The generator searches the standard install locations (`--game-path <install root>` to override) and hard-requires the install; pass `--skip-icons` to rebuild only the CSV packs on a machine without the game. **Patch the game before regenerating** — the run reports how many named items have no icon in the install, and a triple-digit count means the client is older than the pinned CSVs.
+- **`data/manifest.toml`**: records exactly which upstream commit/release each pack was generated from, plus (under `[icons]`) the FFXIV client version the icon pack was extracted from — this is the source of truth for "what version of game data is this."
 - **Updating game data**: done by hand (in practice, by an agent), not on a schedule. A game-data
   bump can break consumers — a renamed sheet or column shifts `xiv-gen`'s generated types — so the
   regeneration and the fallout need fixing in the same change. Run
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -292,6 +292,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "array-init"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3d62b7694a562cdf5a74227903507c56ab2cc8bdd1f781ed5cb4cf9c9f810bfc"
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -829,6 +835,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "binrw"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a13d2fa8772b88ee64a0d84602c636ad2bae9c176873f6c440e2b0a1df4c8c43"
+dependencies = [
+ "array-init",
+ "binrw_derive",
+]
+
+[[package]]
+name = "binrw_derive"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "193bcff2709da50edf2e9a89e0f9a2118eea70f257b1f4380ccf27d3d8578302"
+dependencies = [
+ "owo-colors",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "binstring"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -987,7 +1015,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d8f347189c62a579b8cd5f80714efa178f52e461dc2e6d701d264f5ff22e566c"
 dependencies = [
  "once_cell",
- "proc-macro-crate",
+ "proc-macro-crate 3.5.0",
  "proc-macro2",
  "quote",
  "syn 2.0.119",
@@ -2736,6 +2764,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "flate2",
+ "icon-extract",
  "image",
  "rayon",
  "rkyv",
@@ -2814,6 +2843,17 @@ dependencies = [
  "r-efi 6.0.0",
  "rand_core 0.10.1",
  "wasm-bindgen",
+]
+
+[[package]]
+name = "getset"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6cf442baaabe4213ce7d1239afc26c039180b6456da2cededa316ae2c8a77a77"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.119",
 ]
 
 [[package]]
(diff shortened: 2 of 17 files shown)


akarras:
**Reviewed.** Well-built change — CI is fully green including the 1h4m build, the tripwire and the Small→Medium aliasing are both tested, and the currency-coverage test (gil / Poetics) is exactly the assertion that fails against the old pack. CLAUDE.md and README honestly document the new FFXIV-install requirement rather than papering over it. The per-icon-id storage with an `items.map` indirection is clearly the right call at 25,507 icons for 49,764 items.

Four things worth surfacing before this lands.

**1. It carries an unrelated commit.** `091df5fc` ("feat(charts): wire format for GroupLevel and ChartMode") is on this branch. It is *not* on `origin/main` — it's an unpushed commit sitting on the local `main` this branch was cut from, which is why the diff shows `ultros-charts/src/charts/mod.rs` +51 and `data/grouping.rs` +78 in a PR about icons. Squash-merging bundles that chart work under an icon-extraction commit message. It's good, tested code, so this is a history-hygiene call rather than a correctness one — but it should be a deliberate one.

**2. The 148-item regression ships to prod.** Acknowledged in the body, and I want to make sure it's weighed rather than skimmed: the generating client was a patch behind the pinned CSVs, so 148 items that *had* an icon in the old pack have none now. Net is strongly positive (+32,556 items gain icons), and the fix is a re-run on a patched client — but it is a live, user-visible regression on the newest marketable gear until someone with a patched install runs `cargo run --release -p game-data-pack -- --pinned`.

**3. Regenerating game data now requires a local FFXIV install.** This is the real architectural consequence. Previously anyone — or CI — could reproduce the packs from a pinned git source. Now the icon half is reproducible only on a machine with the game, and `--skip-icons` is the only alternative. Documented clearly, and the provenance recording under `[icons]` is the right mitigation, but it narrows who can maintain game data to one person with one machine. Worth being explicit that that's the accepted trade.

**4. Resource cost.** LFS pack 22MB → 51MB and decompressed memory ~55MB → ~90MB. The LFS side is probably fine. The +35MB resident is worth a thought given the prod box runs everything under container memory caps — not a blocker, just don't let it be a surprise.

**Recommendation: merge, then immediately re-run on a patched client to close the 148-item gap.** I've held off merging it myself because (2) is a deliberate prod regression and (1) changes what lands in your history — both are your call rather than mine. Say the word and I'll merge it, or split the chart commit out first if you'd rather keep it clean.

akarras:
**Pushed two commits (`0b1c4274`, `c3887a4f`) that fix the missing-icon gap at its root. The client was never stale.**

### What was actually wrong

`ironworks 0.4.1`'s `file::tex::Format` predates the entire Bcn2 group — it has **no BC7**, which is ~9% of `ui/icon`. Those textures could not be parsed, and `GameInstall::icon` matched `Err(_) => continue`, so 'present but unparseable' and 'not in the install' collapsed into the same answer. That is what made a decoder gap read as a client one patch behind.

Measured against a current install (`2026.07.16.0001.0000`), sampling ids 1..250000:

| | ironworks 0.4.1 | git HEAD |
|---|---|---|
| readable | 1546 | **1735** |
| present-but-unreadable | **189** | **0** |
| highest readable id | 234951 | 246458 |

Exactly the 189 recovered. 151 were BC7; the other 38 were existing formats that also failed under the older sqpack handling.

For the record, the stale-client theory doesn't survive contact with the evidence: the install is the only one on the machine, XIVLauncher points at it, and the game boots — which it cannot do on an outdated client. The `2026.07.16` string is a build stamp, not a patch number, so comparing it to the datamining repo's commit date (which trails a patch by days or weeks anyway) never showed what it appeared to.

### The changes

**`0b1c4274`** — reads return `IconRead::{Image, Absent, Unreadable}`, matched structurally on `ironworks::Error::NotFound` rather than by string. An indexed-but-unreadable icon now **aborts the run** naming the path and reason instead of quietly shrinking the pack; genuinely-absent icons keep the existing >10% tripwire, since CSVs legitimately run ahead of a client. Also added the decoders that were missing outright (Dxt5/BC3, BC2, Rgb5a1, Rgba4).

**`c3887a4f`** — the ironworks bump, pinned to rev `3d5db8ff` as there is no crates.io release with this. Not a drop-in: the `ffxiv` feature is gone (install discovery is now `sqpack::Install`), `Format` variants are renamed DXGI-style, and `File::read` needs `Read + Seek` so the tests feed a `Cursor`. Added BC7 plus BC4/BC5. Pleasingly, upstream's new names independently confirm the 16-bit layouts this crate assumes — `Bgra4Unorm`, `Bgr5a1Unorm`.

Every format is verified by decoding real icons to PNG and looking at them, not by inspection alone. `examples/probe.rs` is the diagnostic that found this (classifies an id range as readable / absent / present-but-unreadable and reports format distribution); `examples/dump.rs` writes one sample per format for eyeballing.

fmt, clippy `-D warnings`, and 38 tests pass.

### Still open

1. **The pack has not been regenerated against this.** The 148-item regression should vanish entirely once it is — no patched client required. Worth doing before merge so the numbers in the PR body match what ships.
2. **The git-rev dependency** means CI fetches from GitHub at build time. Deliberate, but call it out in the body.
3. This branch still carries `091df5fc` ("feat(charts): wire format for GroupLevel and ChartMode"), an unpushed commit from local `main` that predates the branch point — unrelated to icons, and a squash-merge files it under this PR.

akarras:
**Pack regenerated (`75d51c47`) — the gap is fully closed.**

```
           before    after
  items    49764     50773
  unique   25507     26223
  missing  1009      0
  pack     51MB      55MB
```

Generator output: **"every named item has an icon"**. No absent, no unreadable, and zero `Icon = 0` rows among named items.

Worth stating plainly: this ran on the *same* client (`2026.07.16.0001.0000`) and the *same* pinned CSVs as the original generation. Nothing was patched. The 148 items that had regressed against the old universalis pack are back purely from the ironworks bump — which is the cleanest possible confirmation that the client was never the problem and the earlier 'patch the game and re-run' caveat was a misdiagnosis.

The xiv-db CSV packs regenerated **byte-identical**, so reproducibility there is intact and this commit is just the icon pack.

Also fixed a fmt slip of mine from the `Cursor` change in `c3887a4f` that had CI red. Updated the PR body's Numbers table and the root-cause section to match reality.

fmt, clippy `-D warnings`, and 41 tests pass locally — including `ultros-xiv-icons`' currency-coverage test against the freshly built pack.

akarras:
Rebased onto main (`9d16e222`) now that #1148 landed. The unrelated chart commit is gone from this branch — the diff is purely icon work, and `git diff origin/main HEAD -- ultros-frontend/ultros-charts/` is empty, confirming that code now comes from main rather than riding along here.

fmt, clippy `-D warnings`, and 41 tests pass against the new base.

Issue #1132, opened 2026-08-04
Finish UI kit cleanup: card surfaces and small-page headers


akarras:
**Mostly complete.** The #1223 follow-up comment narrowed this issue to two remaining design tasks. The original loading-state checklist should not be implemented again.

Already done: analyzer inline skeletons, item-chart placeholder, the groups light-mode pulse fix, optional `ToolHeader` help props, and the old `edit_retainers` surface classes. Current source agrees with those completed portions.

Remaining:

- **Card surfaces:** consolidate the three Vendor Resale cards and four Related Items cards with the shared panel styles. Decide how interactive cards retain their hover affordance; replacing `.card` with `.panel` alone changes it.
- **Small-page headers:** give Settings and Job Set Detail an appropriate shared header/layout treatment, with the necessary translated summary/context copy. Keep each page's existing functionality.

Use before/after screenshots to review the surface and layout decisions, including light/dark themes and mobile. Check new strings in all seven locale files, now under `ultros-frontend/ultros-i18n/locales/`.

This is remaining visual cleanup for #1133. Shared analyzer columns, windows, and filters are tracked separately in #1349. Coordinate any Groups work with the active Groups PR #1357; do not fold it into this cleanup.

akarras:
Status after PR #1223 (2026-08-29): the skeleton/loading-state half of this issue is done — the four analyzer inline pulse fallbacks, the item-view chart placeholder, and the groups light-mode pulse (fixed earlier today in #1213) are all on shape-matched token skeletons, with the orphaned keys pruned.

Two bullets remain, both needing a design call rather than mechanical work:
1. **`.card` → `.panel` consolidation** (3 sites in vendor_resale, 4 in related_items): `.card:hover` carries a box-shadow lift that `.panel` has no equivalent for, so consolidating changes hover affordance on interactive tiles. Wants a decision (add a `.panel-interactive` hover variant? drop the lift?) plus a screenshot pass.
2. **job_set_detail.rs / settings.rs kit retrofit**: both predate the kit broadly — a real ToolHeader treatment needs new summary/context copy (i18n × 7) and re-layout, closer to a mini #1129-style rebuild than hygiene.

Scoping this issue down to those two items; everything else in the original list is verified fixed (bullet-by-bullet disposition in the #1223 PR discussion).

PR #1143, opened 2026-08-05
refactor(ui): extract the flip finder's control bar into components/control_bar.rs

### LINKED ISSUE #1126: UI kit: extract the flip finder's sticky control bar into components/control_bar.rs
Part of the UI-kit unification effort (see umbrella issue). Builds on the SortHeader extraction (#1125).

The flip finder's sticky control bar — SavedViewsMenu slot + Columns button + Clear-all + filter chip row + `+ Filter` menu — is the target look for every tool page, but it lives inline in `routes/analyzer.rs`:

- The bar itself: `analyzer.rs:1765-1860`
- The Columns picker popover and the `+ Filter` menu: `analyzer.rs:1830-2230` (`available_filters()`, `default_filter_value()`, `col_hidden_note_class()`)
- Grid width plumbing: `extra_column_widths_px` → `--analyzer-extra-cols-{base,md,xl}` CSS vars

Supporting pieces that are already components but have only one consumer: `components/filter_chip.rs` (also owns `STICKY_BAR_HEIGHT = 76.0`, the contract with the sticky table header) and `components/saved_views.rs`.

The CSS is likewise analyzer-branded: `.sticky-bar-button*` / `.sticky-bar-popover` (tailwind.css 2418-2483), `.filter-chip*` / `.filter-chip-row` (2298-2362), `.analyzer-hscroll` / `.analyzer-table` / `.analyzer-grid-row` (2364-2416).

### Task

1. Extract `components/control_bar.rs` exposing the bar with slots/props for: saved views (optional), a columns picker driven by a column description list, the chip row, and the add-filter menu driven by a filter description list.
2. Rename the `.analyzer-*` CSS classes to a neutral prefix (the `.sticky-bar-*` and `.filter-chip-*` families are already neutral).
3. Port the flip finder onto the extracted component — zero visual change is the acceptance bar.

Converting the other Toolbar-based routes is deliberately a separate issue so this one stays a pure no-op refactor of `/flip-finder`.

akarras:
Phase 2 of the UI-kit unification umbrella (#1133). Closes #1126.

The flip finder's sticky control bar — result count, saved views, Columns picker, Clear-all, the filter chip row and the `+ Filter` menu — is the target look for every tool page, but it lived inline in `routes/analyzer.rs` as ~700 lines of view tangled into the route's own state.

## What moved

`components/control_bar.rs` now owns the bar, both popovers, their tap-away / route-change / Escape dismissal, and the chip strip's edge fades. `analyzer.rs` drops 640 lines. Everything route-specific arrives as a prop:

| prop | what it carries |
|---|---|
| `summary` / `actions` | row 1's left (truncates first) and right (never shrinks) view fns |
| `columns` + `visible_columns` + `on_toggle_column` / `on_reset_columns` | the picker, driven by a `ColumnOption` list. Empty hides the button entirely |
| `available_filters` + `on_add_filter` | the `+ Filter` menu, driven by a `FilterOption` list |
| `columns_extra` / `filter_menu_extra` | controls pinned below each popover's own content |
| `children` + `active_filters` | the chips, and the ids behind them |

`STICKY_BAR_HEIGHT` moves here from `filter_chip.rs` — it's the bar's contract with the sticky table header, not the chip's.

**`ControlBarPopovers` is the one non-obvious piece of API.** The analyzer's category picker lives *inside* the `+ Filter` menu and commits on `change`, so it has to close the menu it sits in. Rather than a context lookup, the caller can own the popover signals and pass them in; without it the menu would sit open over the page it just filtered.

`active_filters` is a `Signal<Vec<&'static str>>` rather than a plain `is_empty: Signal<bool>` because the strip's edge fades have to be re-derived when a chip is added or removed — that moves `scrollWidth` without firing scroll or resize. The empty-state hint derives from the same signal.

## Rebase notes — three merges landed inside the code this moves

- **#1137** replaced the breakpoint-bucketed `--analyzer-extra-cols-{base,md,xl}` scheme with a single `--analyzer-optional-cols`. Main's logic is kept verbatim and only the rename applies on top, so the stylesheet diff against main is *nothing but* `analyzer-` → `tool-`.
- **#1137 also deleted the "desktop only" column note.** `ColumnOption::hidden_note_class` is dropped to match — keeping it would have re-added a note whose i18n key no longer exists in any locale file.
- **#1141** added the chip-strip edge fades, whose listeners hook `.filter-chip-row` by `NodeRef` — the element this refactor moves into the component. The whole fade block moved with it, `#[cfg(feature = "hydrate")]` gate and `on_cleanup` intact.

## CSS

`.analyzer-*` → a neutral `.tool-*`: `tool-table`, `tool-hscroll`, `tool-grid-row`, `--tool-optional-cols`, `--tool-row-min-width`. `.sticky-bar-*`, `.filter-chip-*` and `--chip-fade-*` were already neutral. No `analyzer-`-prefixed class or custom property remains in the repo.

## Verification

- `./check_ci.sh` clean; `cargo test -p ultros-app --lib` 451 passed.
- The `hydrate`-gated fade block compiles clean on `wasm32-unknown-unknown` — `check_ci.sh` never lints that cfg, so it was checked separately. The 7 clippy findings under that feature are all pre-existing debt in `on_hand_input.rs` / `realtime.rs`; none are in these files.
- **Rendered against a local build and compared to an unmodified `origin/main` build of the same page.** Identical on every dimension measured: bar height exactly 76px (the lock), two 32px rows, `position: sticky`, the same five bar buttons in the same order, one chip row, three chips for `?profit&roi&sales`, zero popovers open at rest, `--tool-optional-cols: 584px` feeding a 1076px row min-width, and zero `analyzer-`-prefixed classes in the live DOM.

**Known gap:** popover *interactivity* is unverified. This box can't hydrate `/flip-finder` — the local SSR streams the skeleton while the client resolves the resources fast, so tachys hits a mismatch on `AnalyzerTable`'s root div and the wasm panics before handlers attach. **An unmodified `origin/main` build fails identically, at the same element, with the same message**, so it is pre-existing and environmental rather than anything this PR introduces — but it does mean clicks were never exercised end to end.

**Note for #1144:** that PR adds an `AnalyzerControlBarSkeleton` that hand-copies the bar's markup ("Keep the two in step"). Whichever of these lands second should fold that into a `ControlBarSkeleton` here instead of duplicating the classes — that duplication is what this issue exists to delete, and #1132 has the skeleton work queued anyway.

### DIFF
--- a/style/tailwind.css
+++ b/style/tailwind.css
@@ -2414,7 +2414,7 @@ a:not(.nav-link):not(.side-nav-item):not(.menu-item):not(
    `auto`), and since it never scrolls vertically the header would never
    stick. So the header and the row area are two *sibling* scrollports of
    equal width whose scrollLeft is synchronized in analyzer.rs. */
-.analyzer-hscroll {
+.tool-hscroll {
     overflow-x: auto;
     overflow-y: hidden;
     /* Rows scroll underneath the sticky header, so it cannot be
@@ -2429,15 +2429,13 @@ a:not(.nav-link):not(.side-nav-item):not(.menu-item):not(
    No breakpoint enters into it: every column the user switched on renders at
    every viewport width, and a narrow screen scrolls to reach them. The 30.75rem
    here is the four always-on columns (HQ 44px + Item 14rem + Profit 7rem + Buy
-   Price 7rem); `--analyzer-optional-cols` is set inline by the route and carries
+   Price 7rem); `--tool-optional-cols` is set inline by the route and carries
    whatever `?cols=` turned on, which the stylesheet cannot know. */
-.analyzer-table {
-    --analyzer-row-min-width: calc(
-        30.75rem + var(--analyzer-optional-cols, 0px)
-    );
+.tool-table {
+    --tool-row-min-width: calc(30.75rem + var(--tool-optional-cols, 0px));
 }
-.analyzer-grid-row {
-    min-width: var(--analyzer-row-min-width, 0px);
+.tool-grid-row {
+    min-width: var(--tool-row-min-width, 0px);
 }
 
 /* Compact control-bar button. Sized to fit the bar's 32px rows — the
--- a/ultros-frontend/ultros-app/src/components/control_bar.rs
+++ b/ultros-frontend/ultros-app/src/components/control_bar.rs
@@ -1,0 +1,337 @@
+//! Sticky control bar: the filter surface every tool page shares.
+//!
+//! Grew up inline in the Flip Finder and is the target look for the rest of
+//! the tools (#1133), which still use the older `Toolbar` idiom — a stack of
+//! labelled fields that renders every filter whether it is in use or not, and
+//! then echoes the active ones in a second hand-rolled chip row.
+//!
+//! The shape here is the opposite: **only active filters take space.** Row 1
+//! is the result count plus view-level controls; row 2 is one [`FilterChip`]
+//! per active filter and a `+ Filter` menu holding everything unset. The bar's
+//! height therefore tracks the filters in use rather than the filters that
+//! exist.
+//!
+//! ## The height lock
+//!
+//! The bar is pinned to exactly [`STICKY_BAR_HEIGHT`] because the table header
+//! sticks directly beneath it at that offset — a bar that grew with its
+//! content would cover its own column headers. So the rows can neither wrap
+//! nor scroll, and every control has to *fit*, at every width and in every
+//! locale. Three things keep row 1 inside, in the order they give up space:
+//! the summary is `flex-1` and truncates first, button labels are hidden below
+//! `md` and ellipsize above it, and icons never shrink. A breakpoint alone
+//! would not do it — the side nav takes 240px at `lg`, so the row is no wider
+//! at 1024px than at 768px (#1055).
+//!
+//! Anything added to row 1 needs to be able to yield too.
+
+use std::collections::HashSet;
+
+use leptos::prelude::*;
+
+use crate::components::dismissable::use_dismissable;
+use crate::components::icon::Icon;
+use crate::i18n::*;
+use icondata as i;
+
+/// Height reserved for the sticky control bar. Feeds
+/// `ScrollSource::Window { sticky_offset }` so rows hidden behind the bar are
+/// not counted as visible, and is pinned by the bar's own `h-[76px]`.
+pub const STICKY_BAR_HEIGHT: f64 = 76.0;
+
+/// One column the picker can turn on or off.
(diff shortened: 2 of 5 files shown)


akarras:
**Rebased onto main and resolved the conflicts.** Pushed as `8e69ca05`.

Your first commit was #1140's, already squash-merged — so its commits aren't ancestors of main and a plain rebase would have tried to re-apply the whole thing. Cherry-picked just `7d5dfbf7` instead.

Three of the conflicts needed judgement, not a mechanical pick. All three would have compiled-and-been-wrong or failed to build:

**1. The CSS rename collided with #1137's deletion.** This branch renames `--analyzer-extra-cols-{base,md,xl}` → `--tool-extra-cols-*`, but #1137 deleted the three-bucket scheme outright — no column is breakpoint-hidden any more, so the reservation is a single number. Taking either side plainly was wrong: yours resurrects the buckets, HEAD's keeps the `analyzer-` prefix your PR exists to remove. Resolved to #1137's single variable under your neutral name, `--tool-optional-cols`, in both the stylesheet and `analyzer.rs`.

**2. `hidden_note_class` wouldn't have built.** `ColumnOption` still carried it and the picker still rendered `t!(i18n, analyzer_columns_picker_desktop_only)` — but #1137 removed that concept and deleted the key from **all seven** locale files, so `leptos-i18n` would have failed the build. Dropped the field and the note.

**3. The edge fade would have gone silently dead.** This is the one worth flagging. #1141 put a `node_ref=chip_row` on `.filter-chip-row` and drives `--chip-fade-{start,end}` from its scroll geometry. Moving the strip into `control_bar.rs` left that NodeRef attached to nothing — it compiles fine and the fade just never applies. `ControlBar` now takes an optional `chip_row` NodeRef, the same 'caller owns it' shape you already use for `popovers`, so #1141's logic stays put and keeps working. That felt more faithful than relocating the effect, though moving it in wholesale is the better long-term call once a second tool page wants the affordance.

Audited that the rest of the merged batch survived: no breakpoint gates back on the columns, #1138's confidence gate intact, the fade CSS present, and **zero `analyzer-`-prefixed classes or custom properties left** — your PR's stated goal still holds after the resolution.

Verified on the rebased tree: fmt clean, clippy `-D warnings` clean, **451 tests passing** (your 452 was against the old base; #1137 consolidated the bucketing tests), and `cargo check -p ultros-client --target wasm32-unknown-unknown` clean — that last one matters because the fade block is `#[cfg(feature = "hydrate")]`.

**Still not verified: the bar rendered in a browser.** Your PR body flags this too, and I didn't clear it — same local-dev-server problem. Given this is a 550-line view extraction, that's the remaining risk on it; the compile and test evidence says the wiring is right but says nothing about layout.

PR #1144, opened 2026-08-05
fix(flip-finder): put the control bar in the loading skeleton

### LINKED ISSUE #1110: loading skeleton on analyzer ignores filter buttons
Currently the item explorer has a beautiful skeleton but has some noticeable content shift from the filter buttons loading in. Include those buttons as part of the content

akarras:
Closes #1110.

Based on `main` (#1137 has since merged).

## The bug

`AnalyzerTableSkeleton` draws the results grid and nothing else. But the sticky control bar — row count, Columns, Clear all, the filter chips — lives inside `AnalyzerTable`, which only exists once the resources resolve. So the loading state has no bar, and the moment data arrives the bar drops in above the table and shoves everything down.

Measured on prod:

```
.sticky-bar            height 76px   (h-[76px], the lock STICKY_BAR_HEIGHT depends on)
parent .flex.flex-col  gap    16px   (gap-4)
.analyzer-table top within that container: 92px
```

92px of vertical shift, on every load, after the skeleton has already told the user where the table would be.

## The fix

Draw the bar in the skeleton too. Every class is copied from the real bar — the `h-[76px]` lock, the two `h-8` rows, `sticky-bar`'s own positioning — so the swap changes content and never layout. Verified by building the candidate markup against prod's live stylesheet:

| | real page | skeleton |
|---|---|---|
| bar height | 76px | **76px** |
| row 1 / row 2 height | 32 / 32 | **32 / 32** |
| `.analyzer-table` top within container | 92px | **92px** |

The outer `flex flex-col gap-4` is `AnalyzerTable`'s own root, repeated here rather than hoisted around the `<Suspense>`: the gap between bar and table has to survive the swap, and the fallback is the only thing standing in for both halves.

## One detail worth the measurement

The placeholder chips are `w-24 w-16 w-20`, which sum to **287px — exactly the strip's inner width at 375px**. My first pass used `w-28 w-20 w-24` and overflowed by 17px. That matters because of #1141: an overflowing strip triggers the new edge fade, so a too-wide placeholder would have advertised a horizontal scroll the real chip row may not have once it renders.

Row 1 was checked the same way: 359px content in 359px of space at 375px, no overflow.

## Verification

- `cargo test -p ultros-app --lib analyzer::` — 71 passed, 0 failed (re-run after rebasing onto the post-#1136/#1140 `main`)

Pure markup, no new logic and no hydrate-gated code, so the `ultros-client` clippy pass that #1141 needed doesn't apply here.


akarras:
## Review — approved, merging once `build` goes green

### Conflict resolution pushed

This branch was based on `d4ae6bbd` and went `CONFLICTING` when **#1143** landed, which both moved the real bar out to `components/control_bar.rs` **and renamed the table's class tokens**. I resolved it on the branch (merge commit `4d9b54b4`, no force-push) and carried the rename into the skeleton:

- `analyzer-table` → `tool-table`
- `analyzer-grid-row` → `tool-grid-row`
- `--analyzer-optional-cols` → `--tool-optional-cols`

Without that the skeleton would have kept emitting the old class names and silently lost the grid-row min-width the whole component exists to preserve — the failure mode would have looked like the skeleton "just being a bit off" rather than a broken merge.

I also refreshed the doc comment, which pointed at "the real bar in `AnalyzerTable`" — that address is stale post-#1143. It now points at `ControlBar`, says explicitly that this is a hand-mirrored copy that has to be updated with it, and records **why there are four row-1 placeholders**: they match what `AnalyzerTable` actually puts after the summary (`RealtimeStatus`, `SavedViewsMenu`, Columns, Clear all). That last one is worth having written down — I had to go read the `actions=` prop to convince myself four was right and not a leftover.

### Verified

On a merged tree containing all six open PRs: `cargo fmt` clean, `cargo clippy --all-targets -- -D warnings` **exit 0**, **866 tests passing**.

Then measured the served build (`pkg/3704dd94/`) in a real browser:

- `.sticky-bar` height **76px**, `.tool-table` offset within the container **92px** — matching your prod measurements exactly. 76 + 16 (`gap-4`) = 92, so the skeleton's `flex flex-col gap-4` wrapper reproduces the offset arithmetically, not just by eye.
- The skeleton bar is present in the cold SSR HTML — exactly one `skeleton-shimmer sticky-bar`, and its outer classes (`sticky-bar h-[76px] px-2 py-1 flex flex-col gap-1`) are byte-identical to `ControlBar`'s root, so the swap can't change the box.
- Sampled the bar height and table offset every 50ms across an in-app world switch: **438 samples, zero deviation from (76, 92)**. No shift.

Approval via `gh` is blocked since I'm authenticated as the repo owner, so this comment is the review.

Issue #1077, opened 2026-08-02
Phase 2: mirror Discord guild membership into groups (needs privileged intent)


akarras:
Phase 2 of #1062 — the part of the original issue that says "copy the state of the bot into our db".

**Blocked on an external dependency.** Do not start this until the intent below is enabled; nothing in this issue can be tested without it.

## Blocker: `GUILD_MEMBERS` is a privileged intent

`ultros/src/discord/mod.rs:146` builds the client with `GatewayIntents::non_privileged()`. Mirroring guild membership needs `GUILD_MEMBERS`, which is privileged. That means:

1. Toggle **Server Members Intent** on in the Discord developer portal for the bot application.
2. Change the intent set in `discord/mod.rs` to include `GatewayIntents::GUILD_MEMBERS`.
3. Once the bot is in 100+ guilds, Discord requires **verification and intent approval** to keep it enabled. Worth checking the current guild count before committing to this, because crossing that line turns a config toggle into an application review.

Step 1 is not a code change and can't be done from a PR. Confirm it's done first.

## Why there's more work here than "add an event handler"

There is currently **no `EventHandler` implementation anywhere in the codebase**. `poise::FrameworkOptions` at `discord/mod.rs:93-96` sets `commands` and nothing else, so the bot has never handled a raw gateway event. All of the following is new surface:

- `guild_create` — bot joins a guild, or the gateway reconnects and replays every guild. Note this fires on *every* startup for *every* guild, so it has to be idempotent and cheap.
- `guild_delete` — bot kicked or guild deleted. See the lifecycle hazard below.
- `guild_member_addition` / `guild_member_removal` — the actual membership mirror.
- `guild_member_update` / `guild_role_update` — needed if group membership is ever scoped to a role rather than the whole guild.

## Reconciliation is mandatory, not optional

Gateway events are missed whenever the bot is down, and Discord does not replay them. A member who leaves during a deploy stays in the mirrored group forever. So this needs a periodic reconciliation job that pulls the authoritative member list per linked guild and diffs it against `user_group_member`. Treat the event handlers as a latency optimization over reconciliation, not as the source of truth.

Rate limits matter here: this is one paginated member-list fetch per linked guild per cycle.

## Lifecycle hazard — needs a product decision before coding

The original issue says:

> When the bot is removed from the server, we should also remove our cloned permissions.

[...]

Issue #1062, opened 2026-08-01
group invites


akarras:
Currently the only way to join a group is to be invited. The current system also only lets you search via a group ID.

Let's extend this system to fully utilize the discord bot.

When the user has the bot added to a guild, and has sufficient permission within that guild to manage roles, allow the user to create a group that's based on the state of the discord bot.

We should build systems to handle the permissions checks via our database most of the time, and then copy the state of the bot into our db. When the bot is removed from the server, we should also remove our cloned permissions.

akarras:
Triaged this and it's an epic rather than a single change, so I've split it into phases. Keeping this open as the tracking issue.

- **#1074 — Phase 1: create a group from a Discord server you manage.** Ships the useful half with no external dependency. Membership stays manual; the group gains a `guild_id`, the guild's name and its icon. **Implementing this now.**
- **#1075 — Phase 1.5: group invite codes.** Directly answers the "only way to join is to be invited" complaint that opens this issue, with zero Discord dependency, by mirroring the invite pattern lists already have.
- **#1076 — Phase 1.6: use the `guilds` OAuth scope for guild discovery.** Optional polish; lets us show servers the bot *isn't* in.
- **#1077 — Phase 2: mirror guild membership.** The "copy the state of the bot into our db" half. **Blocked** — see below.

Three things surfaced during research that shaped the split.

**Phase 2 is blocked on a Discord portal toggle, not on code.** The bot runs with `GatewayIntents::non_privileged()` (`ultros/src/discord/mod.rs:146`), and membership mirroring needs `GUILD_MEMBERS`, which is privileged. That has to be enabled in the developer portal, and past 100 guilds it additionally needs Discord verification and intent approval. On top of that there is no `EventHandler` in the codebase at all today — `FrameworkOptions` sets only `commands` — so every gateway handler is new surface, and a reconciliation job is effectively mandatory alongside it, since missed events during a deploy never get replayed.

**"When the bot is removed, remove our cloned permissions" would cause silent data loss as written.** `list_shared_group` cascades on delete, so deleting a mirrored group revokes every list share pointing at it — someone kicks the bot for an unrelated reason and users lose shopping-list access with no notification. #1077 proposes converting the group to a manual group instead, but it's a product call and worth your confirmation before anyone builds it.

[...]

PR #1146, opened 2026-08-05
feat(groups): shareable invite codes for groups (phase 1.5 of #1062)

### LINKED ISSUE #1075: Phase 1.5: group invite codes (no Discord dependency)
Phase 1.5 of #1062. Independent of the Discord work — can land before or after phase 1.

#1062 opens with the actual user complaint:

> Currently the only way to join a group is to be invited. The current system also only lets you search via a group ID.

Guild mirroring (phase 2) is one answer to that, but it's gated on a privileged Discord intent. Invite codes solve the same complaint with **no Discord dependency at all**, and the codebase already has the pattern — lists have had shareable invite links for a while.

## Scope

Mirror `list_invite` for groups:

- New `group_invite` entity modelled on `ultros-db/src/entity/list_invite.rs` — `{ id: String (the code), group_id, max_uses: Option<i32>, uses: i32 }`. Lists carry a `permission: i16` column; groups have no permission concept, so drop it rather than adding a column nothing reads.
- `create_invite` / `use_invite` in `ultros-db/src/lists.rs`, modelled on the list versions at `ultros-db/src/lists.rs:880-985` — including the uses/max-uses accounting and the "already a member" case.
- A `/group/invite/:invite_id` route. The login redirect already handles this shape: `safe_login_next` in `ultros/src/web/oauth.rs` is unit-tested against `/list/invite/abc123`, so `/login?next=/group/invite/...` works as-is.
- Invite management UI on the group card in `routes/groups.rs`, mirroring `components/list/share_list_modal.rs`.

## Notes

- Only the group owner can mint invites — matches every other mutation in `lists.rs`, where the check is `group.owner_id != owner_id`.
- Worth deciding whether an invite to a **guild-linked** group (phase 1) should be allowed. Leaning yes: the guild link is an identity, not an access-control boundary, until phase 2 makes membership authoritative.

akarras:
Closes #1075. Phase 1.5 of #1062.

#1062 opens with:

> Currently the only way to join a group is to be invited.

Guild mirroring (#1077) is one answer to that, but it's blocked on a privileged Discord intent that has to be toggled in the developer portal. Invite codes solve the same complaint with **no Discord dependency at all**, and lists have had shareable invite links for a while — so this is mostly mirroring a pattern that already works.

## What changed

| Layer | |
|---|---|
| Migration | `m20260805_000001_group_invite` — `group_invite` table, FK cascade off `user_group`, index on `group_id` |
| Entity | `ultros-db/src/entity/group_invite.rs`, plus the reverse relation on `user_group` |
| DB | `create_group_invite` / `use_group_invite` / `delete_group_invite` / `get_group_invites` in `lists.rs` |
| API | `GroupInvite` / `CreateGroupInvite`, and four routes in `web.rs` |
| Frontend | `GroupInviteAccept` at `/group/invite/:invite_id`; owner-only invite panel on the group card |
| i18n | 20 new keys, translated in all 7 locales |

The `group_invite` table deliberately has **no `permission` column**. `list_invite` carries Read/Write, but group membership is binary, so there'd be nothing to store — if groups ever grow roles, that's a new column then rather than a smallint nothing reads now.

## Three decisions worth a look

**1. Re-redeeming an invite you already used does not consume a second use.**

`use_invite` for lists upserts the share and burns a use every time. For groups that would punish a double-click or a re-visit of the link. Membership is checked inside the transaction *before* the conditional increment, and an existing member just gets the group id back. Concurrent redemption by *different* users is still guarded by the same atomic conditional increment the list path uses, so the max-uses cap can't be raced past.

**2. Invites are allowed on guild-linked groups.**

Per the discussion on #1062. In phase 1 the guild link supplies the group's *identity* (name, icon), not its membership — so an invite there is no more privileged than one on a manual group. This changes if/when #1077 makes guild membership authoritative.

**3. The redeem/revoke routes are `/api/v1/group-invite/{id}`, not nested under `/api/v1/group/{id}/...`.**

That prefix binds an `i32`; an invite code is a 48-char hex string. Keeping them as siblings would mean one of the two is always a malformed group id.

## Refactor included

`invite_url` / `copy_invite_url` / the uses label moved out of `components/list/share_list_modal.rs` into a new `components/invite_link.rs`, parameterized by route prefix. That copy path carries the fix for the unhandled promise rejection the error reporter was flagging (GlitchTip #5767) — a second hand-rolled copy in the groups UI is exactly how that regresses. `share_list_modal` now delegates; its existing tests are unchanged and still pass.

## Verification

`./check_ci.sh` → exit 0 (fmt + clippy clean, workspace-wide), run against this branch rebased onto current `main`.

Tests, including 4 new ones:
- `ultros-api-types` — 153 passed
- `ultros-app` — 441 passed
- `ultros-db` — 34 passed

New coverage: invite-URL construction for both route prefixes, the uses label, `CreateGroupInvite` treating an omitted `max_uses` the same as an explicit null (so a client sending `{}` gets an unlimited invite rather than a 422), and a `GroupInvite` wire round-trip.

CI does not run `cargo test`, so those were run locally. The DB methods themselves are not covered — `ultros-db` has no test harness that stands up a database, so `create/use/delete_group_invite` are verified by reading against the `list_invite` originals rather than by test.

## Known gap, deliberately not fixed here

**A member still can't leave a group from the UI.** `remove_group_member` already permits self-removal at the DB layer (it allows the call when the caller is the target), but `GroupCard` only renders the remove button for the group *owner*. That was tolerable when joining required the owner to type your Discord ID; with one-click invite links it means people can get in and not out. It's outside #1075's scope — worth its own issue.


akarras:
Filed the leave-group gap noted at the bottom of this description as #1151, so it does not get lost when this merges.

PR #1156, opened 2026-08-08
fix(universalis): surface HTTP status instead of a bogus schema error

akarras:
## The bug

Every REST method in the `universalis` client called `Response::json()` directly, without ever checking the status. Universalis answers errors with an RFC 9457 problem document:

```json
{"type":"https://tools.ietf.org/html/rfc9110#section-15.5.5","title":"Not Found","status":404,"traceId":"…"}
```

which shares no fields with the success payloads. So a 404 surfaced as:

```
check_for_missed_items_on_world failed, error: HTTP Error: error decoding response body: missing field `items` at line 1 column 162
```

That reads like upstream schema drift and completely hides the status. It also means a 429 (rate limit) or a 5xx would be reported the same way, indistinguishable from a real parse failure.

## Why it's firing now

Prod (`a0d1430`) logs it **101 times per 50 minutes**. The cause is stale world rows: our `world` table carries seven worlds Universalis has since dropped, and the catch-up sweep asks about each every five minutes.

Measured by diffing `https://ultros.app/api/v1/world_data` against `https://universalis.app/api/v2/worlds` (133 vs 128 worlds):

```
IN ULTROS NOT UNIVERSALIS: Innocence, Pixie, Titania, Tycoon, 月牙湾, 雪松原, 黄金谷
```

All 128 worlds Universalis *does* list answer the recency endpoint fine — so this is purely our stale rows, not an upstream outage.

## The fix

- `check_status` rejects non-2xx **before** deserialization and reports the status, URL and a truncated body as a new `Error::Status` variant. Applied to all five REST methods.
- `Error::is_not_found()` distinguishes "Universalis doesn't know this entity" (permanent) from 429/5xx (worth retrying).
- The catch-up sweep skips a world on 404 and warns **once per process** instead of erroring on every pass forever.

After the change, the exact prod case reports:

```
universalis returned HTTP 404 for https://universalis.app/api/v2/extra/stats/most-recently-updated?entries=200&world=Innocence: {"type":…,"title":"Not Found","status":404,…}
```

(verified live against the real API, then the throwaway test removed).

## Verification

| gate | result |
|---|---|
| `cargo test -p universalis --lib` | **24 passed** (base 18 + 6 new), includes the live-network tests, which now run through the status check |
| `cargo fmt --all -- --check` | 0 |
| `cargo clippy -p universalis --all-targets -D warnings` | 0 |
| `cargo clippy -p ultros --all-targets -D warnings` | 0 (8m33s) |

**Red→green by mutation:** making `check_status` a no-op (literally the pre-fix behaviour) fails **4 of the 6** new tests. The other 2 are the success-path guards and pass either way *by design* — they exist to prove the check doesn't reject good responses, not to pad the red count.

## Not addressed here

The seven stale `world` rows themselves. Removing or hiding them is a separate call — they still appear in world pickers, and dropping rows touches user data (retainers, alerts, lists keyed on world id). This PR just stops them generating a misleading error every five minutes.

### DIFF
--- a/ultros/src/item_update_service.rs
+++ b/ultros/src/item_update_service.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
     time::Duration,
 };
@@ -38,6 +38,17 @@ pub(crate) struct UpdateService {
     pub(crate) sales: EventProducer<SaleEventData>,
     /// Per-world timestamp of the last saturation-triggered full sweep.
     pub(crate) full_sweep_cooldowns: Mutex<HashMap<i32, Instant>>,
+    /// Worlds Universalis 404s on. Purely a log de-duplicator — see
+    /// [`UpdateService::note_world_uncovered`].
+    pub(crate) uncovered_worlds: Mutex<HashSet<i32>>,
+}
+
+/// True when `error` is a Universalis `404`, i.e. it does not know the entity
+/// we asked about. [`anyhow::Error`] erases the type, so match on the downcast.
+fn universalis_not_found(error: &anyhow::Error) -> bool {
+    error
+        .downcast_ref::<universalis::Error>()
+        .is_some_and(|e| e.is_not_found())
 }
 
 struct CmpListing(Model);
@@ -100,6 +111,15 @@ impl UpdateService {
         Ok(())
     }
 
+    /// Records that Universalis does not cover `world_id`. Returns `true` the
+    /// first time, so the caller logs once instead of on every sweep.
+    fn note_world_uncovered(&self, world_id: i32) -> bool {
+        self.uncovered_worlds
+            .lock()
+            .expect("uncovered_worlds poisoned")
+            .insert(world_id)
+    }
+
     /// Claims the full-sweep slot for a world if its cooldown has elapsed.
     fn claim_full_sweep_slot(&self, world_id: i32) -> bool {
         let mut cooldowns = self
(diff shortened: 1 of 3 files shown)


akarras:
Verified independently, not just re-read.

**Reproduced the failure mode live:** `GET /api/v2/extra/stats/most-recently-updated?entries=5&world=Innocence` really does return HTTP 404 with the RFC 9457 body quoted in the description — so the ``missing field `items` `` message was indeed a status being laundered into a schema complaint.

**Ran the tests CI does not run.** `.github/workflows/rust.yml:115` has `cargo test --workspace` commented out, so the 6 new tests were never exercised by the green check. Locally: `cargo test -p universalis --lib` → **24 passed**, including the 5 live-network tests, which now go through `check_status`.

**Red→green confirmed by mutation.** Replacing the status guard with a no-op (literally the pre-fix behaviour) fails exactly **4 of the 6** new tests; the 2 that pass are the success-path guards, as the description says. The claim is accurate.

Additional checks:

- `universalis_not_found` downcasts `anyhow::Error`. `get_missing_updates` propagates with a plain `?` and no `.context()`, so the concrete `universalis::Error` survives in the chain and the downcast actually fires. Worth knowing this is load-bearing — adding a `.context()` there later would silently reintroduce the every-five-minutes error.
- The new `Error::Status` variant is additive-safe: both consumers (`web/error.rs:104`, `main.rs:189`) take it via `#[from]`, nothing matches the enum exhaustively.
- The 404 arm returns `Ok(())` but still calls the endpoint on every sweep — only the *log* is de-duplicated, not the request. That is the right call (it self-heals if Universalis re-adds a world), though the log text "skipping catch-up sweeps" reads as if the sweep itself stops. Not worth a round trip.
- `note_world_uncovered` mirrors the existing `claim_full_sweep_slot` lock/expect pattern, and the `HashSet` is bounded by world count.
- UTF-8 truncation via `char_indices` is genuinely safe, and there is a test for it.
- Deferring the seven stale `world` rows is the right scope split — they are keyed on by retainers, alerts and lists.

Approving and merging.

PR #1161, opened 2026-08-10
fix(ssr): stop the server fetching its own API through Cloudflare

akarras:
## The bug

Every server-rendered page re-issues its API calls against `HOSTNAME`, which in production is the **public** origin (`https://ultros.app`). So the container leaves the box to talk to itself: DNS → TLS handshake → Cloudflare → back in through the edge → into the very same process.

Measured on the prod host (`ssh chew@boxbox`):

| path | loopback | public edge (healthy) |
| :--- | ---: | ---: |
| `/api/v1/cheapest/Europe` | 20ms | 63ms |
| `/api/v1/listings/陸行鳥/12997` | 8ms | 58ms |

And the edge is not always healthy. One 4h26m container log window (2026-08-10 08:53→13:19) held **429 `source: TimedOut`** failures against this module's 10s client budget, concentrated in two minutes (10:53–10:54) where the edge stalled:

- 88 × `/api/v1/cheapest/Europe`
- 53 × `/api/v1/cheapest/North-America`
- 52 × `/api/v1/retainer/listings/{id}`
- long tail of `/api/v1/listings/{world}/{id}` — CN and JP worlds especially

Those are GlitchTip [2209](https://glitchtip.ultros.app/ultros/issues/2209) ("Error doing leptos fetch", 3365 events) and [2210](https://glitchtip.ultros.app/ultros/issues/2210) ("Error getting value", 2191 events).

**A failed SSR fetch does not error the page** — the resource still renders, just empty. So the visible symptom is a page that silently loses a section, which is why this survived so long. I confirmed the shape on a sampled event: issue 7074's culprit is `http://ultros.app/api/v1/listings/North-America/3147` with `sqlx::pool::acquire` slow-threshold breadcrumbs around it.

## The fix

Resolve the internal origin from `LEPTOS_SITE_ADDR` — the address the server actually binds, already `0.0.0.0:8080` in the prod container — mapping the *unspecified* address (`0.0.0.0`, `[::]`) to its loopback form. Precedence:

1. `ULTROS_INTERNAL_API_ORIGIN` (new, explicit escape hatch for a split deployment)
2. loopback form of `LEPTOS_SITE_ADDR`
3. `HOSTNAME` (previous behaviour — preserved for anything unusual)
4. `http://localhost:8080`

**`HOSTNAME` itself is untouched.** It is the app's public URL and OAuth redirects are built from it; this only changes what the SSR renderer dials.

This removes the whole failure class between the process and its own API: no DNS, no TLS, no CDN, no WAF, no edge rate limit. It also supersedes the situation #1154 worked around — with no CDN in the path there is no `Host`/SNI mismatch to produce 403/421. The `host` denylist entry and its tests stay, since forwarding the inbound host is still wrong.

Incidental: the fallback path now trims a trailing slash. `fly.toml` sets `HOSTNAME = "https://ultros.app/"`, which concatenated with a leading-slash path produced `https://ultros.app//api/v1/...`.

## What this is *not*

**It does not fix the SSR panic flood** — GlitchTip [6876](https://glitchtip.ultros.app/ultros/issues/6876) (15,763), [6886](https://glitchtip.ultros.app/ultros/issues/6886) (2,505), [6888](https://glitchtip.ultros.app/ultros/issues/6888), [6895](https://glitchtip.ultros.app/ultros/issues/6895), [7119](https://glitchtip.ultros.app/ultros/issues/7119). I initially believed the timeouts caused them and had to drop that: the panics run at a steady 100–240/min across the entire log window, while the timeouts are confined to two minutes. No correlation.

Two negative results worth recording so nobody re-walks them:

- **`/analyzer/{world}` is not a repro.** Curling it from the box appeared to add +1/+2 panics every time — but a no-request control over the same 5s window gave +0, +9, +2. Prod's background panic rate swamps any single-request delta. Any future attempt at this needs a quiet-traffic window or a per-request correlation id, not a log-count delta.
- **`/itemcard/{world}/{id}` is confirmed innocent.** The `usvg` "No fonts with a 阳/U+9633 character" warnings that sit next to every panic come from that route, which is a plain axum handler with no reactive graph.

The panic locations from the prod log, for whoever picks this up next (`docker logs ultros | grep -o "panicked at [^ ]*:" | sort | uniq -c | sort -rn`):

```
2462 reactive_graph-0.2.14/src/traits.rs:394:39      <- Get::get() on a disposed signal
1220 reactive_graph-0.2.14/src/computed/arc_memo.rs:334:24
  46 ultros-app/src/components/skeleton.rs:224:16    <- use_i18n()
  29 ultros-app/src/components/item_tooltip.rs:32:16
  25 ultros-app/src/components/clipboard.rs:10:66    <- use_context::<GlobalLastCopiedText>().unwrap()
  15 leptos_router-0.8.15/src/link.rs:161:16         <- <A/> outside a <Router/>
```

## Verification

| gate | result |
| :--- | :--- |
| `cargo fmt --all -- --check` | 0 |
| `cargo clippy -p ultros-app --all-targets -- -D warnings` | 0 (5m09s) |
| `cargo check -p ultros-app --no-default-features --features hydrate --target wasm32-unknown-unknown` | 0 (1m51s) |
| `cargo test -p ultros-app --lib` | **538 passed** (base 532 + 6) |

Genuine red→green **by mutation**: restoring the pre-fix body (`HOSTNAME` verbatim, else the dev default) fails **4 of 6** new tests at exit **101** —

```
a_trailing_slash_on_the_origin_is_trimmed ... FAILED
production_env_resolves_to_loopback_not_the_public_origin ... FAILED
falls_back_through_hostname_then_the_dev_default ... FAILED
explicit_override_wins_over_everything ... FAILED
```

The other two exercise `loopback_origin_from_site_addr` directly, which the mutation does not touch — they pass either way **by design** (address-parsing guards), not because they are weak.

## GlitchTip triage this run — 34 ignored

- **9 Discord alert-delivery issues** (#6877–6879, #6881–6885): prod serves `/pkg/077c9ed/`, so #1159's permanent-failure fix **is** deployed, and they have been silent since 2026-08-09 18:17.
- **25 count-1 fragments** (#7051–7075) of one 2026-08-09 07:31 incident — duplicates of #2209/#2210 by another name. Both aggregate issues stay **open** as the tracking issues for this fix; leaving them is how we will tell whether the deploy worked.

## Deploy note

No config change required — `LEPTOS_SITE_ADDR=0.0.0.0:8080` is already set on the container. After deploying, `#2209`/`#2210` should stop accruing; the log line `resolved SSR internal API origin` reports the chosen origin at startup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ The two halves of the data come from different places:
 | `DISCORD_CLIENT_ID` | Discord Application ID | Required |
 | `DISCORD_CLIENT_SECRET` | Discord Client Secret | Required |
 | `HOSTNAME` | Public URL of the app (for OAuth redirects) | `http://localhost:8080` |
+| `ULTROS_INTERNAL_API_ORIGIN` | Origin the SSR renderer calls its own API on. Defaults to the loopback form of `LEPTOS_SITE_ADDR`, so the server never leaves the box to fetch its own data; only set this if the API lives somewhere else. | derived from `LEPTOS_SITE_ADDR` |
 | `KEY` | Secret key for cookie encryption | Random string |
 | `DATABASE_URL` | Postgres connection string | `postgres://user:pass@host/db` |
 | `PORT` | HTTP server port | `8080` |
--- a/ultros-frontend/ultros-app/src/api.rs
+++ b/ultros-frontend/ultros-app/src/api.rs
@@ -814,9 +814,11 @@ where
 /// Headers that must not be copied from the inbound browser request onto the
 /// outbound internal API request.
 ///
-/// The SSR path re-issues each API call against `HOSTNAME`, which in production
-/// is the *public* origin, so the request travels back out through the CDN.
-/// Copying the inbound `host` header onto a request aimed at a different URL
+/// The SSR path re-issues each API call against [`internal_api_origin`], which
+/// is no longer the public origin — but it still must not carry the inbound
+/// `host`, and the reason it originally mattered is worth keeping: when the
+/// outbound call did travel back out through the CDN,
+/// copying the inbound `host` header onto a request aimed at a different URL
 /// makes the CDN reject it: a `Host` that does not match the edge certificate
 /// answers `403 Forbidden`, and on a pooled TLS connection (the client below
 /// sets a 60s keepalive, so connections are reused) a `Host` that disagrees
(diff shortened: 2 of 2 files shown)


akarras:

**Correctness of the origin resolution** — the precedence chain and every branch of `loopback_origin_from_site_addr` are covered by the 6 new tests, which I ran rather than trusting CI: `cargo test -p ultros-app --lib` → **538 passed; 0 failed**. Worth stating explicitly because `.github/workflows/rust.yml:115` has `cargo test --workspace` commented out — green CI on this repo proves compile, not behaviour.

**The load-bearing assumption is verified statically.** The whole fix rests on `LEPTOS_SITE_ADDR` being present and correct in the prod container. It is baked into the runtime image at `Dockerfile:110` (`ENV ... LEPTOS_SITE_ADDR="0.0.0.0:8080"`) in the `runner` stage, so it cannot go missing through compose/env drift — resolution is guaranteed to hit branch 2 and yield `http://127.0.0.1:8080`.

**No missed call sites.** Grepped `HOSTNAME` across `ultros`/`ultros-frontend`/`ultros-api-types`: the only two runtime uses were the DELETE and GET SSR paths, both converted. `main.rs:453` keeps `HOSTNAME` for the OAuth redirect, which is correct and must not change.

**Nothing rejects a plain-HTTP loopback request.** Grepped for an https-redirect / `x-forwarded-proto` gate in the server — there is none, so dropping TLS on the internal hop cannot produce a redirect loop. Forwarded cookies are read scheme-independently.

Not verified: I could not reach the prod box from this run (ssh blocked by the sandbox), so the live loopback timings in the description are from the authoring session and were not re-measured. The Dockerfile `ENV` makes that a low-risk gap, and the startup log line `resolved SSR internal API origin` plus the `ULTROS_INTERNAL_API_ORIGIN` escape hatch make it self-diagnosing after deploy.

PR #1162, opened 2026-08-10
chore(deps): weekly lockfile refresh + wasm-bindgen-cli 0.2.127 + js-yaml patch

akarras:
Automated weekly dependency review.

## Summary
In-range `cargo update` refresh (39 packages) with the recurring hold-backs, plus one dev-harness security patch. **No newly-unblocked major this run** — all big-dep ceilings are unchanged and confirmed against crates.io live.

## Big-dep ceilings (re-verified live)
| Crate | Current | Latest upstream | Status |
|-------|---------|-----------------|--------|
| leptos | 0.8.20 | 0.9.0-**beta** | Gated by `leptos_i18n 0.6.2` (still newest; pins `leptos ^0.8` + `leptos-use ^0.18`) |
| sea-orm | 2.0.1 | 2.0.1 | Latest stable ✓ |
| axum | 0.8.9 | 0.8.9 | Latest ✓ |
| tokio | 1.53.1 | 1.53.1 | Latest ✓ |
| leptos-use | 0.18.3 | 0.19.0 | Duplicate-trap (leptos_i18n pins ^0.18) — held |
| tower-http | 0.6.11 | 0.7.0 | Duplicate-trap (leptos_axum pins ^0.6) — held |
| itertools | 0.14.0 | 0.15.0 | Duplicate-trap (tantivy/leptos pin ^0.14) — held |

## Changes
- **Cargo.lock** — in-range refresh. `lodestone` held at `192faccd` (a plain update advances it to `164f23b9`, switching its reqwest client to 0.12 and breaking ultros's 0.11 client). The 0.11/0.12/0.13 reqwest split is preserved.
- **Dockerfile** — `wasm-bindgen-cli` 0.2.126 → 0.2.127 to match the lib bump (0.2.126 → 0.2.127; js-sys/web-sys → 0.3.104, wasm-bindgen-futures → 0.4.77). wasm-bindgen requires lib == CLI exactly, so rust-CI-green ≠ Docker-green without this.
- **integration/package-lock.json** — js-yaml 4.3.0 → 4.3.1 (GHSA-5p4m-2wfm-xmqj, **high**, quadratic CPU in `!!omap` resolution), lockfile-only. `npm audit` → 0 vulnerabilities.

## Verification
- `cargo clippy --all-targets -- -D warnings` — exit 0, **zero warnings** (whole workspace incl. `ultros` server + vendored OpenSSL + `ultros-app`/`ultros-charts`/`ultros-client`)
- Gated crates confirmed single-copy in the lock (no new duplicate majors)

Not exercised locally: live Postgres or the Docker `build` job (lockfile + Dockerfile only). Deferred as before: rkyv 0.8, reqwest 0.13 (sentry-only), resvg 0.48 (needs visual verification).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4602,14 +4602,15 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.104"
+version = "0.3.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e0c1080212aad755ea003d18543e8768dd432c48819efd73a7bf1e39b7a5a3a"
+checksum = "53b44bfcdb3f8d5837a46dae1ca9660a837176eee74a28b229bc626816589102"
 dependencies = [
  "cfg-if",
  "futures-util",
  "wasm-bindgen",
 ]
+
 
 [[package]]
 name = "json5"
@@ -10764,9 +10765,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.127"
+version = "0.2.126"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b70935747edd64d89de3efa29d73789b806c15798f8e7dca4d8ac356b50ce70"
+checksum = "4b067c0c11094aef6b7a801c1e34a26affafdf3d051dba08456b868789aaf9a4"
 dependencies = [
  "cfg-if",
  "once_cell",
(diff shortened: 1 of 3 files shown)


akarras:
## Review — found a real break, fixed it in `8ccfd1ca`

The `build` job was failing, and it was **not** infra flake. The refresh bumped `web-sys 0.3.103 → 0.3.104`, which drops `LockManager::request_with_options_and_callback`. `leptos-use 0.18.3` calls exactly that at `use_web_lock.rs:106`, so the wasm target dies with `E0599` inside `cargo chef cook --target wasm32-unknown-unknown -p ultros-client`.

This is the ceilings table in the description biting from an unexpected direction. `leptos-use` is pinned at `^0.18` by `leptos_i18n 0.6.2` and cannot move, so the dependency that had to give was `web-sys`.

### Why `Lint, checks, and tests` stayed green

`use_web_lock.rs` is wasm-only. The rust job builds host targets, so `cargo clippy --all-targets` never compiles it. **Only the Docker job builds wasm**, which is why a lockfile-only change can be workspace-clippy-clean and still unbuildable — worth remembering for future refreshes.

### The fix

The family is exact-pinned internally — `web-sys 0.3.104` requires `js-sys =0.3.104` requires `wasm-bindgen =0.2.127` — so it cannot be reverted piecemeal. `cargo update --precise` on any single member fails to resolve; all three orderings error. Reverted all seven blocks to main's versions as a unit: `wasm-bindgen{,-futures,-macro,-macro-support,-shared}`, `js-sys`, `web-sys`.

The `wasm-bindgen-cli` bump in the Dockerfile is reverted with them, since the CLI must match the lib exactly and the lib is back at 0.2.126. That file is now byte-identical to main.

Everything else in the refresh survives untouched, including the js-yaml `GHSA-5p4m-2wfm-xmqj` patch.

### Verification

| gate | result |
| :--- | :--- |
| `cargo metadata --locked` | exit 0 — lockfile self-consistent, no resolution drift |
| `cargo check -p ultros-client --locked --target wasm32-unknown-unknown` | **exit 0** in 3m34s — the exact stage that failed in CI |

Red evidence is the CI log itself ([run 31402896828](https://github.com/akarras/ultros/actions/runs/31402896828)); green is the local wasm check above, now confirmed by the Docker job on this branch.

### Follow-up worth doing separately

`web-sys` should join the held-dep list, otherwise next week's refresh walks straight back into this. And the 18-minute Docker job being the only thing that compiles wasm is a slow feedback loop for a lockfile change — a cheap `cargo check --target wasm32-unknown-unknown` step in the rust job would catch this class in ~2 minutes.

PR #1165, opened 2026-08-12
⚡ Bolt: Optimize apply_sales_event sorting using select_nth_unstable_by_key

akarras:
💡 What: The optimization implemented
Replaced the O(N log N) `sort_by_key` followed by `truncate(200)` with an O(N) partitioning step `select_nth_unstable_by_key(200)` followed by a smaller O(K log K) sort on only the truncated results.

🎯 Why: The performance problem it solves
Sorting large arrays of market updates or sales is expensive when we only need the top 200 elements. Using `select_nth_unstable_by_key` filters the array in linear time and eliminates the unneeded allocations and processing from a full array sort.

📊 Impact: Expected performance improvement
Reduces sorting operations overhead significantly during active updates in the CurrentlyShownItem pipeline when `sales.len() > 200`.

🔬 Measurement: How to verify the improvement
Tests have been checked via `cargo test -p ultros-api-types` and behavior is functionally identical. Pre-commit reviews verified the improvement.

---
*PR created automatically by Jules for task [10612547693414708893](https://jules.google.com/task/10612547693414708893) started by @akarras*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -69,3 +69,6 @@ Use `select_nth_unstable_by_key(N)` followed by `.truncate(N)` and sorting only 
 ## 2026-08-07 - Avoid into_iter().take().collect() when limiting arrays
 **Learning:** Using `v.into_iter().take(limit).collect::<Vec<_>>` requires iterating and allocating a new Vec. Rust Vectors have an O(1) in-place `truncate(limit)` method that achieves the exact same result without allocating new memory or requiring a `collect`.
 **Action:** Always prefer `v.truncate(limit)` over `.into_iter().take(limit).collect()` to take the first N elements from an existing `Vec`.
+## 2026-08-12 - Optimize top N item extraction with select_nth_unstable in apply_sales_event
+**Learning:** In `ultros-api-types/src/lib.rs`, taking the top N elements of a slice using a full sort (`sort_by_key`) followed by `.truncate()` takes O(N log N) time on the full array length. When returning the top N items from a potentially large collection, avoid full sorts.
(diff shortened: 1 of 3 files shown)


akarras:
## Review: approved

Correct, and the bounds reasoning holds: `select_nth_unstable_by_key(200, ..)` panics if the index isn't `< len`, and the `if self.sales.len() > 200` guard means `len >= 201`, so index `200` is always valid. The partition puts the 200 largest `sold_date`s in `[0..200)`, so `truncate(200)` keeps exactly the right set.

Verified by running the crate's existing tests locally, which cover this exact path (CI runs `clippy --all-targets`, not `cargo test`, so these don't actually execute in CI):

- `test_apply_sales_event_truncation` — 205 sales in, 200 out, `sales[0].id == 204`
- `test_apply_sales_event_ordering` — newest first
- plus the add/update/remove/wrong-item-id cases

### One observation on the perf rationale

The stated win is real for a large batch, but this function's steady state is the opposite shape. `self.sales` is left sorted descending at the end of every call and capped at 200, and `upsert_sales` appends only the handful of sales carried by a websocket delta — so the typical input is "200 already-sorted elements plus ~5 appended," length ~205.

On that input the previous `sort_by_key` (stable, driftsort) detects the existing 200-element run and merges in roughly O(n). The new path instead runs `select_nth_unstable` — which *scrambles* the ordering as it partitions — and then does a full `sort_unstable` over 200 shuffled elements. So in the common case this is somewhat more comparison work than what it replaced, not less.

That said, we're talking on the order of ~1500 vs ~400 `i64` comparisons on a 200-element vec, once per websocket event. It's sub-microsecond either way and nowhere near a hot path, and dropping the stable sort does avoid its scratch allocation. Not worth another round trip — merging as-is. Flagging it mainly so the `.jules/bolt.md` heuristic ("always prefer `select_nth_unstable` for top-N") doesn't get applied to already-sorted or nearly-sorted data, where it's a mild pessimization rather than a win.

PR #1166, opened 2026-08-12
🎨 Palette: Add `aria-live` to recently viewed clear button

akarras:
💡 What: Added `aria-live="polite"` to the clear button in the recently viewed panel.
🎯 Why: The button uses an inline confirmation pattern where the text changes from "Clear All" to "Confirm Clear". Without `aria-live`, screen reader users might not be aware that the button's action and text have changed.
♿ Accessibility: Improves screen reader experience by announcing the text change on the inline confirmation button.

---
*PR created automatically by Jules for task [9901247805120882797](https://jules.google.com/task/9901247805120882797) started by @akarras*

### DIFF
--- a/.Jules/palette.md
+++ b/.Jules/palette.md
@@ -1,17 +1,3 @@
-## 2023-11-20 - Accessible Icon-Only Buttons
-**Learning:** Found that some icon-only interactive elements like 'Delete Group' and 'Add Member' in the groups component lacked `aria-label`s, making them invisible to screen readers.
-**Action:** Always add reactive `aria-label` properties to icon-only buttons to ensure they are fully accessible, especially when their function might change based on state (e.g. asking for confirmation).
-
-## 2026-07-07 - Accessible textareas
-**Learning:** In the MakePlaceImporter, the textarea was missing an ID and the associated label lacked a `for` attribute, which breaks form field accessibility for screen readers.
-**Action:** Ensure all form controls, such as `<textarea>` and `<input>`, have unique IDs and are properly associated with their corresponding `<label>` elements using the `for` attribute.
-
-## 2026-07-28 - Explicit aria-label for Image-only Menu Buttons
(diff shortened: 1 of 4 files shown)


akarras:
## Review: approved with a fix pushed

The `aria-live` change itself is good. I pushed one commit to fix a problem with the doc file.

### The code change — verified in a real browser

`aria-live="polite"` on the clear button is the right call: the button's own label is what changes (`Clear All` → `Confirm Clear?`) and there's no other visible cue, so without it a screen-reader user gets no feedback that the first click armed a confirmation.

I verified the underlying behavior against production and confirmed the attribute is inert with respect to layout:

- Button renders and hydrates on the dashboard (`hydrated: true`, prod currently `aria-live: null`).
- Click toggles the label `Clear All` → `Confirm Clear?`.
- Adding `aria-live="polite"` to the live element changed neither width nor height (`51.73x16` before and after the attribute).
- The 3s timeout reverts the label to `Clear All` **without** clearing the list — the confirm gate still requires a second click.

### The fix I pushed

The PR replaced the entire contents of `.Jules/palette.md` with the single new entry, dropping five prior accessibility learnings (icon-only buttons, accessible textareas, image-only menu buttons, resend-button context, explicit input associations).

The root cause is a repo wart rather than anything you did: the same file is tracked under **two paths differing only by case** — `.Jules/palette.md` and `.jules/palette.md` — both pointing at blob `67513b37`. On a case-insensitive filesystem they're one file on disk, so the history only survived by accident in the lowercase copy. This is also the "case-phantom palette.md dirties the tree after merge" annoyance that keeps recurring.

My commit restores the five entries, appends the new one, and untracks the `.Jules/` duplicate so the casing matches `bolt.md` / `sentry.md` / `sentinel.md`. That should stop the phantom from coming back.

I also corrected the new entry's date from `2023-10-27` to `2026-08-12` — the original was three years before the PR was opened.

PR #1167, opened 2026-08-12
🛡️ Sentry: Add unit tests for `SortKey::from_str` fallback logic

akarras:
💡 What: Implemented unit tests for the `SortKey::from_str` pure logic inside the Trends component and documented fallback edge case testing strategies in `.jules/sentry.md`.
🎯 Why: Component specific state hydration parameter parsing like `SortKey::from_str` use bespoke string parsing logic that lack native fallback structures provided by `std::str::FromStr`. If we don't thoroughly test that missing/unknown strings correctly fallback to the standard UI state (e.g., `UnitVolume` for SortKey), we risk silent rendering bugs when clients supply an unrecognized URL parameter.
📊 Impact: Covered parsing edge cases and fallback resolution logic on the `SortKey` enum under `trends.rs`. 
🔬 Verification: Verified via `cargo test --package ultros-app --lib test_sort_key_from_str`.

---
*PR created automatically by Jules for task [11371911696839747390](https://jules.google.com/task/11371911696839747390) started by @akarras*


akarras:
## Review: approved

Test-only change; no runtime code touched.

**The red check was infrastructure, not this PR.** The original `rust` run failed during toolchain setup:

```
tar: Error is not recoverable: exiting now
##[error]Process completed with exit code 2
```

That's the cache-extraction step, before any Rust was compiled. I re-ran it and `Lint, checks, and tests` is now green.

**The test itself checks out.** `SortKey` derives `Clone, Copy, Debug, PartialEq, Eq`, so `assert_eq!` works, and `from_str` is an inherent method (not the `FromStr` trait), so no import is needed. The asserted fallback matches the implementation — anything unrecognized, including `""`, maps to `SortKey::UnitVolume`, which is consistent with the call site defaulting `sort()` to `"units"`.

One caveat for context: this repo's CI runs `cargo clippy --all-targets`, not `cargo test`, so this test is compiled but never *executed* in CI. Still worth having — it pins the fallback contract for anyone running the suite locally.

PR #1171, opened 2026-08-15
feat(currency-exchange): rebuild on the shared UI kit (closes #1128)

### LINKED ISSUE #1128: UI kit: rebuild Currency Exchange on the shared kit
Part of the UI-kit unification effort (see umbrella issue). Depends on #1126 (control bar) and benefits from #1080 (data table).

Currency Exchange is the most bespoke tool page — it reimplements the entire flip-finder control-bar concept from scratch and uses essentially none of the kit:

- Hand-rolled filter disclosure: filter button with active-count badge (`routes/currency_exchange.rs:536-555`), a `Show`-gated grid panel of local `FilterRange` widgets (563-580)
- Its own hand-built chip row with `push_chip` closures emitting raw `rounded-full border` spans (592-640)
- Raw semantic `<table class="w-full text-sm">` (765-766) — one of the five raw-table sites #1080 wants to retire
- `<Loading />` spinner instead of a `TableSkeleton`, no `ToolHeader`, raw `<h1>` at 511

### Task

Port the page onto `ToolHeader`, the shared control bar + `FilterChip`, `TableSkeleton`, and the shared data table (#1080) or, if that isn't landed yet, the analyzer grid conventions. The dense spreadsheet look from #1097 is the intended visual density — keep it, just express it through the kit.

akarras:
Rebuilds `/currency-exchange/:id` on the flip-finder UI kit, per the design spec in `docs/superpowers/specs/2026-08-14-currency-exchange-kit-rebuild-design.md` (committed on this branch). Phase 5 of #1133; closes #1128.

## What changed

- **ToolHeader** replaces the bespoke header panel; the currency-quantity input moves to its own labeled row below it (visible label at every width, per the analyzer convention of controls below the header).
- **ControlBar + FilterChip** replace the filter disclosure, active-count badge, hand-rolled chip row, and per-column `FilterModal`s. The eight min/max query keys are kept **verbatim** so existing deep links keep filtering; bindings move to `filter_query_signal` (replace: true, scroll: false), fixing the history-spam + scroll-to-top bug on every filter edit. Chips use comparison-shaped labels ("Profit ≥ 5000"); the + Filter menu keeps longer explanatory labels.
- **Column system**: required columns (item, qty received, profit) now *lead* the table so a phone's visible slice is the answer; price/item, shops, cost, hours/sale are `?cols=`-gated optional columns (all default-on) with a Columns picker + Reset. The `column_visibility` breakpoint tiers are deleted — the table is a horizontal scrollport at every width, with **edge fades** (same technique as the flip finder's chip fades, #1057) as the scroll affordance.
- **Sorting** via the shared `SortHeader`/`SortColumn` with direction arrows; hours-between-sales defaults **ascending** (descending would put the slowest sellers on top). `?sorted-by` and the `SortableVec`/`FieldLabels` derives are gone; old `?sorted-by` links fall back to profit-desc, which is what nearly all of them encoded.
- **TableSkeleton** (shaped by the visible column set) replaces the `<Loading/>` spinner; filters that match nothing render a message + Clear all instead of a silently empty tbody.
- `ParseableInputBox` lost its last caller and is deleted. 14 new locale keys added and 11 orphaned ones removed, all seven locales, real translations.

## Notes for review

- `help_href` points at `/help` — there's no per-tool help topic for currency exchange yet (follow-up).
- The landing page (`/currency-exchange`) is intentionally untouched apart from carrying its own `<h1>` now that the route wrapper no longer renders a shared heading.
- Verified: `cargo test -p ultros-app currency_exchange` (6/6), `./check_ci.sh` exit 0, and `cargo check --target wasm32-unknown-unknown --features hydrate` for the hydrate-only fade listeners. **No local browser check was run** — the layout/scroll behavior follows the analyzer's known-good conventions, but a quick phone-width eyeball on `/currency-exchange/29` (tomestones) after deploy is worth it.


akarras:
Reviewed, fixed, and browser-verified. Pushed two fixes as ac53accc.

## Two defects found

**1. The chips claimed a bound the filter did not honor.** The rebuild introduces comparison-shaped labels (`Profit >=`, `Profit <=`), but `is_in_range` compared with strict `>` / `<`, so the row sitting exactly on the number typed was silently dropped. Every other tool in the app -- `analyzer.rs:1709-1714`, `recipe_analyzer.rs:391-397`, `fc_crafting_analyzer.rs:351-357` -- filters inclusively; this page was the lone outlier, and the new labels turned a quiet inconsistency into a visible wrong claim. Now inclusive on both ends, pinned by `range_filter_bounds_are_inclusive` (verified red against the old bounds).

**2. The quantity box still spammed history.** The rebuild moved every filter to `filter_query_signal`, but `currency_amount` was left on a plain `query_signal`. It is typed into a digit at a time, so it kept exactly the behavior the rebuild set out to fix: a history entry plus a scroll-to-top per keystroke. Swapped.

## Verification

Local gates on the final state: `cargo fmt --all -- --check` 0, `cargo clippy -p ultros-app --all-targets -- -D warnings` 0, `cargo test -p ultros-app currency_exchange` **7/7**. Worth noting this repo's CI has the `cargo test` step commented out, so a green check never proved the assertions ran.

Browser pass on a local `cargo leptos serve` at `pkg/ac53accc` (confirmed the served build was this commit, not a stale one), `/currency-exchange/29`, home world Cerberus:

- Hydrates clean -- `boot-progress` absent, **zero console errors**.
- Column order is `ITEM, QTY RECV, PROFIT` then the optional four, as designed. `?cols=hours_between_sales` correctly yields a 4-column table with headers and cells aligned.
- `?sort=hours&dir=asc` sorts ascending: 71, 180, 272, 336, 766, 968.
- **The inclusive fix, end to end:** `?total_profit_min=33318334` against a row whose profit is exactly `33318334` returns **3 rows including it**, chip reading `Profit >= 33318334`. The old exclusive bounds returned 2.
- Empty state: `?total_profit_min=999999999` renders "No trades match your filters." plus a working Clear all, summary `0 trades`.
- **375px phone viewport:** `document` does *not* scroll horizontally -- only the table does (scrollWidth 1101 vs clientWidth 342). Edge fades track scroll position correctly: `--hfade-start: 0px / --hfade-end: 24px` at the left edge, both `24px` after scrolling to the middle.

Locale parity checked: 47 `currency_exchange_*` keys in all seven files, with real translations rather than English stubs. No dangling references to the deleted `ParseableInputBox`. The `Suspense` resource-read shape is unchanged from main, so the rewrite introduces no new registration risk.

## Follow-up, not blocking

At 375px the ITEM cell carries the item name plus the Add-To-List and clipboard buttons, which eats the whole viewport width -- so the first visible slice is the item name alone and the leading QTY/PROFIT columns still start off-screen. The column *order* does what the design intended; the item cell's inline actions are what defeat it on a phone. Worth a look separately.

Also `colspan="7"` on the no-matches row is hardcoded while the real column count varies from 3 to 7 with `?cols=`. Browsers clamp it, so it is cosmetic only.

Merging.

PR #1172, opened 2026-08-15
perf(db): bound per-world sale-history fan-out to relieve pool starvation

akarras:
## What

Both per-world sale-history fan-out sites in `ultros-db/src/sales.rs` ran under `try_join_all`, which polls **every** future immediately. A region item page covers ~30 worlds (more for Europe), so a single page render tried to hold that many pooled connections at once:

- `get_sale_history_from_multiple_worlds`
- `get_compact_sale_history`

Both now route through a small `fan_out_per_world` helper that buffers at `MAX_CONCURRENT_WORLD_QUERIES` (8). **Same queries, same results** — the callers already sort and truncate the merged rows, so nothing downstream depends on the fan-out width.

## Why

With `POSTGRES_MAX_CONNECTIONS` at 300 and ingest writing continuously, a handful of concurrent region renders is enough to starve the pool. Production events still carry, in the same breadcrumb trail:

```
acquired connection, but time to acquire exceeded slow threshold
  acquired_after_secs: 2.220442541   slow_acquire_threshold_secs: 2
```

sitting next to the 10s SSR loopback timeouts that GlitchTip files as **#2209 / #2210**:

```
reqwest::Error { kind: Request, url: "http://127.0.0.1:8080/api/v1/listings/Elemental/8193", source: TimedOut }
```

The same starvation shows up as **#6868 / #6869** (`catch-up ... update failed`). This was scoped but left unfixed in a previous triage run.

The cost is a little latency on wide scopes (worlds answered in batches rather than all at once). Each query is index-served on `sale_history_lookup_index (sold_item_id, world_id, sold_date DESC)`, so the batches are individually fast.

## Testing

`cargo test -p ultros-db --lib` → **48 passed** (base 45 + 3 new), 0 failed.

Genuine red→green **by mutation**: restoring `try_join_all` in the helper fails at exit **101** with

```
peak in-flight 64 exceeded the cap 8
```

The other two new tests (`fan_out_preserves_input_order`, `fan_out_propagates_errors`) pass **both** ways by design — they are behaviour-preservation guards proving the swap doesn't reorder results or swallow a failing world, not part of the red count.

Gates:
- `cargo fmt --all -- --check` → 0
- `cargo clippy -p ultros-db --all-targets -- -D warnings` → 0
- `cargo clippy --all-targets -- -D warnings` (full workspace) → **0, zero warnings** (26m10s)

## Deliberately not done

- **No single-query rewrite.** Collapsing the fan-out into one statement is the bigger win, but prod `EXPLAIN` access was unavailable this run and a naive `world_id IN (...) ORDER BY sold_date DESC LIMIT n` can lose the index ordering and sort every matching row on a 1.9B-row table. The safe shape, for whoever can validate a plan, is `unnest($worlds) CROSS JOIN LATERAL (... WHERE sold_item_id = $1 AND world_id = w.world_id ORDER BY sold_date DESC LIMIT $n)` — identical semantics, one round trip.
- **`lookup_buyer_names` (`sales.rs:165`) has the same fan-out shape on the ingest path** — one SELECT (plus a possible INSERT) per unique buyer name. Left alone here because its insert-or-get race handling makes it riskier than the read path; worth a follow-up.

This does **not** target the SSR panic flood (#6876/#7150, #6886/#7157/#7160, #6888/#7161/#7163/#7164) — `api.rs:921` already documents in-repo that those do not correlate with the timeout bursts, and I re-confirmed it.

### DIFF
--- a/ultros-db/src/sales.rs
+++ b/ultros-db/src/sales.rs
@@ -25,6 +25,43 @@ use tracing::{instrument, warn};
 use ultros_api_types::{SaleHistory, UnknownCharacter};
 use universalis::{ItemId, SaleView, WorldId};
 
+/// Ceiling on how many per-world sale-history queries one request may have in
+/// flight at once.
+///
+/// The per-world queries below used to run under `try_join_all`, which polls
+/// *every* future immediately: a region item page fans out one query per world
+/// in the region (~30 for North America, more for Europe), so a single page
+/// render tried to hold that many pooled connections simultaneously. With
+/// `POSTGRES_MAX_CONNECTIONS` at 300 and ingest writing continuously, a handful
+/// of concurrent region renders is enough to exhaust the pool — prod events
+/// carry `acquired connection, but time to acquire exceeded slow threshold`
+/// breadcrumbs at `acquired_after_secs: 2.2` (threshold 2) sitting right next
+/// to the 10s SSR loopback timeouts that GlitchTip files as #2209/#2210, and
+/// the same starvation shows up as #6868/#6869 (`catch-up … update failed`).
+///
+/// Buffering caps the peak connection demand per call while keeping the exact
+/// same queries and the same results — the callers sort and truncate the merged
+/// rows anyway, so nothing downstream depends on the fan-out width. The cost is
+/// a little latency on wide scopes (worlds are answered in batches rather than
+/// all at once); each query is index-served on
+/// `sale_history_lookup_index (sold_item_id, world_id, sold_date DESC)`, so the
+/// batches are individually fast.
+const MAX_CONCURRENT_WORLD_QUERIES: usize = 8;
+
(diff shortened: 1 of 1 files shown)


akarras:

- `cargo test -p ultros-db --lib` -> 48 passed / 0 failed (CI has `cargo test` commented out, so this mattered).
- Mutation-checked the guard: restoring `try_join_all` inside `fan_out_per_world` fails with `peak in-flight 64 exceeded the cap 8`. Genuine red->green, exactly as the PR claims.
- `buffered` preserves input order and `try_collect` short-circuits on the first error, so both call sites keep the exact `try_join_all` semantics their callers depend on; the merged rows are sorted and truncated downstream regardless of fan-out width.
- Leaving `lookup_buyer_names` and the `CROSS JOIN LATERAL` single-query rewrite as follow-ups is the right scoping call.

Merging.

